PR #15, opened 2017-04-09
Autocomplete input focus fix

### LINKED ISSUE #10: Autocomplete Multiple feature missing input focus
If we use autocomplete by enabling multiple feature(i.e, multiple=true) then input should get the focus like other cases.This can be replicated on showcase as well.

sudheerj:
#10


mertsincan:
Thanks @sudheerj, I made some changes on your PR.

PR #27, opened 2017-04-26
Fixed #26

### LINKED ISSUE #26: activeIndex property on TabView doesn't work
Unable to create a TabView where the activeIndex is not the first, because the component uses the value from its state and ignores the value it receives in props.

We should use the componentWillReceiveProps lifecycle method to copy the value from component props to state.

koverg70:


mertsincan:
Thanks for the PR. I added only your tabview fix into next version. Thanks again.

PR #73, opened 2017-08-28
Port reorder column from ngDataTable.

alwasa0b:
I ported the current reorder column functionality from the primeng project.

### DIFF
--- a/src/components/datatable/DataTable.js
+++ b/src/components/datatable/DataTable.js
@@ -66,7 +66,8 @@ export class DataTable extends Component {
         onRowUnselect: null,
         onRowExpand: null,
         onRowCollapse: null,
-        onContextMenuSelect: null
+        onContextMenuSelect: null,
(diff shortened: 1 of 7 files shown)


mertsincan:
Thank a lot for PR, @alwasa0b. I'll review and get back to you.

alwasa0b:
This seems to have broken the Column Toggler demo. That is because I am keeping a variable to hold the reordered columns within the DataTable.js. I am not sure how you want to go about that. I can think of two ways to go about this. First option, would be a callback function to update the column array with the new order when onDrop is done. Second option is to move the Multi-Select toggle to the DataTable.js. I like the second option because it allows the caller the freedom to create columns dynamically or statically. Let me know what you think..

cagataycivici:
Please monitor at #81 

We can't merge it but will consult to it during our implementation which is very soon.

PR #78, opened 2017-09-03
Added local gulp script and node-sass to build App.scss

francesco-strazzullo:
I Added a local `gulp` script in the `package.json`. So in order to work with PrimeReact it's not mandatory to install gulp globally.

I've also added a `node-sass` command in order to compile the `App.scss` file in order to correctly start the application.


cagataycivici:
Will merge manually, thank you.

PR #124, opened 2017-10-05
Fix for column header issue when resizable is enabled

jimmyjamieson:
This should fix the alignment issue on datatable head. Give it a test and let me know what you think, or if you have any suggestions.


jimmyjamieson:
Found another issue caused by moi. Closed until fix

Issue #123, opened 2017-10-05
Dropdown Menu in DataTable with resizableColumns


jimmyjamieson:
When resizeable columns in enabled in DataTable the layout for the Dropdown breaks

https://prnt.sc/gtm5wx

cagataycivici:
Fixed at https://github.com/primefaces/primereact/issues/126

cagataycivici:
Better to add className="ui-column-filter" to the dropdown or any other custom filter element.

PR #125, opened 2017-10-06
Fix/data table resizable th fix

jimmyjamieson:
https://github.com/primefaces/primereact/issues/123


cagataycivici:
This breaks compatibility with other Prime* libraries theming so we'll fix it differently.

jimmyjamieson:
Yeah. I noticed some of the layout is a bit off. Flex wouldn't go a miss for aligning things, it's a bit better than what's currently there. If it breaks compatibility though, I'm not sure on the path you want to take.

Is there a minimum browser requirement for primereact?

Issue #126, opened 2017-10-06
Dropdown menu sits behind grid


jimmyjamieson:
The dropdown menu in DataTable sites behind the grid when resizableColumns or scrollable is enabled.

Caused by the overflow hidden;

https://screenshots.firefox.com/49XyjS9g09rpWvUd/localhost

cagataycivici:
Thank you.

PR #127, opened 2017-10-06
Fix for dropdown sitting behind header when 

jimmyjamieson:
Fix for https://github.com/primefaces/primereact/issues/126


jimmyjamieson:
This only fixes the header btw. There's still an issue when scrollable is active.

Issue #133, opened 2017-10-10
Click on input filter causes sort


jimmyjamieson:
When you click in an input in the DataTable header, it automatically re-orders the data by that column.

https://screenshots.firefox.com/WfF9eAOP21cpz7z9/localhost

It should only re-order when you click outside the input.

jimmyjamieson:
I tried setting event.stopPropagation() on the click event in HeaderCell.js but no luck.

jimmyjamieson:
Ok, using event.stopPropagation(); inside the input component onClick works

jimmyjamieson:
Also I should probably open a new issue for this. But from my side resizing columns breaks too. I had to fix it locally with custom drag areas. Then reordering I had to implement outside the grid itself.

PR #135, opened 2017-10-11
Fix/input propagation

jimmyjamieson:
https://github.com/primefaces/primereact/issues/133

I see you've already assigned it. But here's the fix i'm using for my local to do what you wish.


cagataycivici:
We solved this specifically on DataTable as we need propagation in other cases like closing an overlay when outside (e.g. a random input) is clicked.

PR #142, opened 2017-10-16
fix/#133

lucapinelli:
I still have this bug if I use a custom filter. Adding a wrapper to stop the propagation fix this issue also in that case. I tested it doing this change on the version tagged 1.0.1, I'm not sure if this works on the last version in master. 
I hope that this can be of some help.


cagataycivici:
I'll choose another way that we did in PrimeNG

Issue #147, opened 2017-10-20
MultiSelect: onClick: event.stopPropagation is not a function


lucapinelli:
Selecting an option of a MultiSelect causes the exception:
`"Uncaught TypeError: event.stopPropagation is not a function"`

Above the plunkr to replicate the issue:
https://plnkr.co/edit/BlGiYZs0RxB4uLbsYNjt?p=preview

cagataycivici:
Grazie!

PR #148, opened 2017-10-20
fix/#147-multiselect-bug

lucapinelli:
Fix for https://github.com/primefaces/primereact/issues/147


cagataycivici:
I'll choose another way that we did in PrimeNG

PR #150, opened 2017-10-20
fix for event.stopPropagation is not a function issue #147

alwasa0b:


alwasa0b:
seem to have already been resolved by another PR

PR #161, opened 2017-10-26
DataTable performance improvement

michalgebauer:
DataTable.processData() does sorting and filtering of the table data. Former code called this method inefficiently 3-4 times during rendering, which had significant impact on performance in case table contained lots of data and used sorting & filtering & pagination.


cagataycivici:
Thank you.

PR #194, opened 2017-11-29
Spinner buttons submit form

leny:
Missing `type` for button in Spinner triggers the submit of the containing form.

Issue was corrected in *primeng*, but not in *primereact*.

cf. primefaces/primeng#2079


cagataycivici:
Got it, thank you!

PR #252, opened 2017-12-13
ProgressSpinner css broken

leny:
Hi, 

This quickfix resolve a bug ([which was already fixed/not appeared in primeng](https://github.com/primefaces/primeng/blob/master/src/app/components/progressspinner/progressspinner.css#L1-L13)), which cause breaking CSSO (and potentially other css-analysis tool, since the target piece of css is not valid) with this specific CSS.

Hope you will merge this soon (had some code depends on it 😉 ).

Thanks 👍


cagataycivici:
Awesome, thanks.

PR #297, opened 2018-01-29
Including the tree key attribute customization.

leospinola:
When the tree data source is changed, automatically attributed keys on the nodes still the same. React was unable to re-render the component.

This kind of approach should be extended to other components.


cagataycivici:
We'll refactor Tree in next version so I'll check this soon.

cagataycivici:
Tree is being reimplemented, I'll make sure this does not happen in new tree.

cagataycivici:
This is obselete now with the new Tree. Please review it next week when 2.0.0-beta.4 comes out.

PR #303, opened 2018-02-01
Clear method of messages is broken

mrganser:
Clear method was setting messages to null and breaking the render. I saw on the last commit that the initial value of messages was changed to an empty list, but clear wasn't change.


cagataycivici:
Fixed grow as well, thanks.

Issue #306, opened 2018-02-06
File Upload Dialog opens only once if set to Auto


tsinghammer:
### There is no guarantee in receiving an immediate response in GitHub Issue Tracker, If you'd like to secure our response, you may consider *PrimeNG PRO Support* where support is provided within 4 business hours

**I'm submitting a ...**  (check one with "x")
```
[x ] bug report
[ ] feature request
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=57
```

**Plunkr Case (Bug Reports)**
Please fork the plunkr below and create a case demonstrating your bug report. Issues without a plunkr have much less possibility to be reviewed.

https://www.primefaces.org/primereact/#/fileupload

**Current behavior**
Currently, if auto is set to true for the file upload opens the choose window only once. Clicking it a second time has no effect.

**Expected behavior**
Die choose file dialog opens and another file can be selected.

**Minimal reproduction of the problem with instructions**
This issue can be reproduced on the PrimeReact samples page. Maybe this behavior is intetnional. But if possible, it would be easier to deal with the component if you could open the choose dialog multiple times.

* **Browser:**  Chrome 63 | Firefox 58 
 
* **Language:** [all]

PR #307, opened 2018-02-06
Resetting the file input style to be able to select another file.

tsinghammer:
###Defect Fixes
In basic mode, the file upload component does not allow to select another file if auto is set to true.
This should reset the display style once the upload is complete, so that the file input can be clacked another time.


cagataycivici:
This is fixed, thank you.

https://github.com/primefaces/primereact/issues/306

PR #352, opened 2018-03-18
Use processed data for exporting into CSV file

### LINKED ISSUE #349: DataTable export doesn't respect filters and sorting
```
[x] bug report
[ ] feature request
[ ] support request => Please do not submit support request here, instead see http://forum.primefaces.org/viewforum.php?f=57
```

**Current behavior**
Export into CSV exports the all dataset without applying filters and selected sorting

**Expected behavior**
Exported dataset should be equal to that is visible in the table row now.

berdyshev:
closes #349

### DIFF
--- a/src/components/datatable/DataTable.js
+++ b/src/components/datatable/DataTable.js
@@ -657,7 +657,7 @@ export class DataTable extends Component {
     }
 
     exportCSV() {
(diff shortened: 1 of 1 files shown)


mertsincan:
Thanks for the PR!

Issue #179, opened 2017-11-09
Unsortable option for DataTable columns


ebonventi:
After sorting a 'DataTable', there are no options to remove the applied sorting.

cagataycivici:
Better to introduce an option like unsortable since not all users may expect it to work like this.

cagataycivici:
Better to introduce an option like unsortable since not all users may expect it to work like this.

PR #360, opened 2018-03-22
Option to reset sort per column (allowUnsort prop on DataTable)

bassarisse:
Following what was discussed here: https://github.com/primefaces/primereact/issues/179

I added a functionality to remove the sort of a column. This first click sorts by ascending, the second by descending, and the third removes the sort configuration. However, this behavior is enabled only if the prop `allowUnsort` is included, so current usage won't break.

Also, as a side-effect, the modifications resolved a bug in multiple sort mode, where a click on a column header right after clicking another one didn't flip the sort order, but instead fixed it to ascending.

Is this suitable?

### DIFF
--- a/src/components/datatable/DataTable.js
+++ b/src/components/datatable/DataTable.js
@@ -36,6 +36,7 @@ export class DataTable extends Component {
         sortOrder: null,
         multiSortMeta: null,
         sortMode: 'single',
+        allowUnsort: false,
         emptyMessage: "No records found",
         selectionMode: null,
         selection: null,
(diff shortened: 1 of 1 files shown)


salup:
+1

Merge please.

cagataycivici:
Thank you, no other Prime table have this due to lack of interest in this particular feature, if more users demand it, we'll reconsider, thank you.

bassarisse:
It's really unfortunate we'll have to rely on using a fork for that =(

Djuki:
+1

Please reconsider adding this feature. @cagataycivici @mertsincan

PR #361, opened 2018-03-22
DataTable scrollHeight enhancement

balloy:
The code change includes 2 items:
1. Call this.setScrollHeight() in componentDidMount()
This is to provide a chance to handle events like parent container changes height (e.g. within a split pane)

2. support scrollHeight format like '100% fixed'
The added " fixed" option allow DataTable take the space even there's no enough rows.

Thanks.


cagataycivici:
Handled differently, calculations must be the same with other Prime projects, if issue persists please create issue tickets.

Issue #382, opened 2018-04-12
Org chart does not update when value property changes


james7272:
**I'm submitting a ...**  (check one with "x")
```
[ X] bug report
[ ] feature request
[ ] support request => Please do not submit support request here, instead see https://forum.primefaces.org/viewforum.php?f=57
```

**Plunkr Case (Bug Reports)**
[Link to plunker](https://plnkr.co/edit/VCVZsO6J0q6i71o0Ssti?p=preview)

**Current behavior**
When the Org Chart's 'value' property is updated the chart does not re-display with the new data.

**Expected behavior**
The chart should refresh

**Minimal reproduction of the problem with instructions**
See the plunker above.

* **React version:**
16.2.0

* **PrimeReact version:**
1.5.1

* **Browser:**  Chrome 65.0.3325 | Firefox 59.0.2
 
* **Language:** TypeScript 2.8.1 | Whatever plunker uses

mertsincan:
Thanks a lot for the PR! I think we can solve it without setState. Fixed now.

PR #383, opened 2018-04-17
#382 Fix update Org chart on change value property.

sswebcoder:
Small changes for support updating Org Chart component in case on change value property..Issue #382


mertsincan:
Thanks a lot for the PR! I think we can solve it without setState. Fixed now.

sswebcoder:
For fix it without setState need convert state component to stateless component.
If you know another way, please tell me.

Issue #389, opened 2018-04-23
onColReorder return columns: undefined 


Rarity:
```
[x] bug report
```

**Plunkr Case (Bug Reports)**
Please fork the plunkr below and create a case demonstrating your bug report. Issues without a plunkr have much less possibility to be reviewed.

**Current behavior**
event in onColReorder method return columns: undefined

**Expected behavior**
event in onColReorder method should return column list

**Minimal reproduction of the problem with instructions**
Look console after reorder column
https://plnkr.co/edit/hTW5tm8brpHb9EpzchRh?p=preview

* **React version:**
16.3.1

* **PrimeReact version:**
1.5.1

* **Browser:** All
 
* **Language:** All

mertsincan:
Thanks a lot for the PR!

PR #390, opened 2018-04-23
fix onColumnDrop method

Rarity:
fix https://github.com/primefaces/primereact/issues/389


mertsincan:
Thanks a lot for the PR!

Issue #433, opened 2018-06-04
Can't set csv export filename in DataTable.js using Internet Explorer, it returns undefined.csv


RyanKys:
[x ] bug report
[ ] feature request
[ ] support request => Please do not submit support request here, instead see 

Can reproduce it using IE in demo site:
https://forum.primefaces.org/viewforum.php?f=57

https://plnkr.co/edit/C6qjkdLtbmy92Hogdnjb?p=preview

**Current behavior**
IE can't define CSV export filename

**Expected behavior**
Should able to define filename similar to Chrome

* **PrimeReact version:**
1.5.2

* **Browser:** [Affect IE only]

RyanKys:
As I mentioned in the primefaces forum, the issue is due to a bug in DataTable.js:

if(windows.navigator.msSaveOrOpenBlob){
  navigator.msSaveOrOpenBlog(blob,this.exportFilename+ '.csv');
}

it should be "this.**props**.exportFilename"

mertsincan:
Thanks a lot for the  PR!

PR #434, opened 2018-06-04
Update DataTable.js to fix IE csv filename export

RyanKys:
fix navigator.msSaveOrOpenBlob(blob, this.props.exportFilename + '.csv'); so that IE can define CSV export filename correctly

###Defect Fixes
https://github.com/primefaces/primereact/issues/433


mertsincan:
Thanks a lot for the PR!

PR #521, opened 2018-07-29
AutoComplete should accept array of objects

mspclaims:
Suggested changes to AutoComplete: 
- accept array of objects instead of array of strings as data feed
- allow user to choose what object's property to display in the field when the item is selected (default is 'label')

It's a very common use case.

New properties added:
```
        objectsInList: PropTypes.bool,
        displayField: PropTypes.string,
```

Example of data:
```
this.icd9codes=[
      {'value':'102','label':'YAWS'},
      {'value':'1020','label':'INITIAL LESIONS, YAWS'},
      {'value':'1021','label':'MULTIPLE PAPILLOMATA AND WET CRAB YAWS, YAWS'}];
```
Instantiating of the control:
```
<AutoComplete id="ICD9Codes"  value={this.state.code} onChange={(e) => this.setState({code: e.value})}
                suggestions={this.state.suggestedICD9} completeMethod={this.suggestICD9Data}  
                objectsInList = {true} displayField='value' multiple={true} 
                dropdown={true} dropdownMode='current' placeholder=''/>   
```

### DIFF
--- a/src/components/autocomplete/AutoComplete.d.ts
+++ b/src/components/autocomplete/AutoComplete.d.ts
@@ -24,6 +24,8 @@ interface AutoCompleteProps {
     size?: number;
     appendTo?: any;
     tabindex?: number;
+    objectsInList?: boolean;
+    displayField?: string;
     completeMethod?(e: {originalEvent: Event, query: string}): void;
     itemTemplate?(data: any): JSX.Element | undefined;
     selectedItemTemplate?(data: any): JSX.Element | undefined;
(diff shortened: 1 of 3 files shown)


aloker:
Isn't that already supported? See https://www.primefaces.org/primereact/#/autocomplete, section "Objects". There's already the prop "field" to determine the display property.

mspclaims:
Indeed.. Though it works a bit different (and a bit weird). Please check example: https://next.plnkr.co/edit/HTIYgj7Oct9GWO1m .  I.e. start typing 'in' - search is performed on 'label' field but dropdown displays 'value' field. My version would display values of 'label' field in dropdown and display 'value' only when item is selected.
Anyway, I'll close my pull request for now. 

Thanks for your awesome set of controls!

PR #525, opened 2018-08-02
Improve typescript definitions

aloker:
The typescript definitions of this lib appear to leave something to be desired. I took a first step to improve them by making them at least compatible with the showcase. That is, I ported the showcase to Typescript in my own repo and fixed the issues that I found in the definitions. 

This PR fixes several issues:
- Syntax errors (e.g. in Editor.d.ts)
- Missing members (e.g. in Growl.d.ts)
- Typos (e.g. in Inplace.d.ts)
- Minor issues discovered by tslint (mostly converting Array<T> to T[] for simple T and ordering, ie. methods after properties)

Note that this are only the basic changes to get at least the showcase working in Typescript. There's *a lot* that can be improved, e.g.:
- All template methods should return React.ReactNode, not JSX.Element
- Events should probably be React.SyntheticEvent with or without a template argument
- Many components would benefit from generics (think `<AutoComplete<Car>..../>`)
- Check which props are actually optional
- In general, most of those `any`s can be taken care of.

I'm willing to polish the definitions, but obviously I need to know if you'll accept and merge such a PR.

### DIFF
--- a/src/components/button/Button.d.ts
+++ b/src/components/button/Button.d.ts
@@ -1,6 +1,7 @@
 import React = require("react");
+import {Omit} from "../util";
 
-interface ButtonProps extends React.HTMLProps<HTMLButtonElement> {
+interface ButtonProps extends Omit<React.HTMLProps<HTMLButtonElement>, 'ref'> {
     label?: string;
     icon?: string;
     iconPos?: string;
--- a/src/components/accordion/Accordion.d.ts
+++ b/src/components/accordion/Accordion.d.ts
@@ -22,4 +22,6 @@ interface AccordionProps {
     onTabChange?(e: {originalEvent: Event, index: number}): void;
 }
 
+// tslint:disable-next-line:max-classes-per-file
 export class Accordion extends React.Component<AccordionProps,any> {}
+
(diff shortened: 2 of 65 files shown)


src/components/button/Button.d.ts
```diff
@@ -1,11 +1,12 @@
 import React = require("react");
+import {Omit} from "../util";
 
-interface ButtonProps extends React.HTMLProps<HTMLButtonElement> {
+interface ButtonProps extends Omit<React.HTMLProps<HTMLButtonElement>, 'ref'> {
```
aloker: This one is tricky. `React.HTMLProps<HTMLButtonElement>` already has a `ref`-prop which exposes the HTMLButtonElement. However `Button` (as every component) has its own `ref`. Those definitions now clash, so I removed it from the inherited props type. Also, the ref is now for the Button instance, not the DOM element. So if you need to expose the native element, there are two solutions: either have an extra prop that is passed to the DOM element (e.g. nativeRef) or consider using forward refs (https://reactjs.org/docs/forwarding-refs.html#forwarding-refs-to-dom-components).

This change was necessary for the DataScroller demo.

cagataycivici:
Merged, please keep the improvements coming

Issue #537, opened 2018-08-17
PrimeReact input events are not compatible with React forms libraries


rodriguezc:
**I'm submitting a ...** 
```
[ ] bug report
[ x] feature request
[ ] support request
```

**Plunkr Case (Bug Reports)**
See Prime React Show Case in the project PR.

**Current behavior**
Forms libraries like Formik provides a global onChange and onBlur handler in order to manage the form state. 
Some Prime React components don't provides a "standard" event object (with e.target.*)

**Expected behavior**
onChange events must provide a standard event

**Minimal reproduction of the problem with instructions**

**Please tell us about your environment:**
N/A

* **React version:**
Current Prime React version
* **PrimeReact version:**
Current Prime React version

cagataycivici:
PR merged, thank you.

PR #538, opened 2018-08-17
Support for React Form libraries

rodriguezc:
###Defect Fixes
PrimeReact input events are not compatible with React forms libraries. (#537 )
This PR normalize Prime React input events

### DIFF
--- a/src/components/autocomplete/AutoComplete.js
+++ b/src/components/autocomplete/AutoComplete.js
@@ -198,10 +198,18 @@ export class AutoComplete extends Component {
     }
     
     updateModel(event, value) {
+        const resolvedFieldData = this.props.field ? ObjectUtils.resolveFieldData(value, this.props.field) : undefined;
         if(this.props.onChange) {
             this.props.onChange({
                 originalEvent: event,
-                value: value
+                value: value,
+                stopPropagation : () =>{},
+                preventDefault : () =>{},
+                target: {
+                    name: this.props.name,
+                    id :  this.props.id,
+                    value: resolvedFieldData ? resolvedFieldData: value
+                }
             });
         }
     }
--- a/src/components/calendar/Calendar.js
+++ b/src/components/calendar/Calendar.js
@@ -667,7 +667,14 @@ export class Calendar extends Component {
         if (this.props.onChange) {
             this.props.onChange({
                 originalEvent: event,
-                value: value
+                value: value,
+                stopPropagation : () =>{},
+                preventDefault : () =>{},
+                target: {
+                    name: this.props.name,
+                    id :  this.props.id,
+                    value
+                }
             });
         }
     }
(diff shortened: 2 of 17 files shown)


cagataycivici:
Is it possible to resolve the conflicts?

cagataycivici:
Also please drop formik demo, we don't need that in the showcase. Then we'll merge.

cagataycivici:
Also is there a standard on these form libraries on how to integrate with 3rd party UI suites, this PR seems to be tuned only for formik, do we need to change code again for another library, why do we need to change code on our side?

rodriguezc:
Managing forms is really verbose in react. 

It quickly becomes necessary to integrate a React form library.
Best form libraries for React are):
-Formik
-Final Form
-React-redux-form

These libraries implements a default onChange and onBlur handler.
All of them check value in "e.target.value",

React-redux-form also check if event has stopPropagation and preventDefault before getting the value in e.target.value. Otherwise the value is the event object itself. In case of primereact it will be {e: value} 

It's possible to override their implementation in order to be compatible with Prime React. Everyone who use Prime React and a Form library needs to make this override.

Example of default Final Form onChange override (onChange is setted with {...input} but then overrided):
https://codesandbox.io/s/40mr0v2r87 

It think that it could be much better if Prime React is compatible with these default handlers.

Formik default implementation of onChange can be found there:
https://github.com/jaredpalmer/formik/blob/master/src/Formik.tsx#L287

Redux form default implementation of onChange can be found there:
https://github.com/davidkpiano/react-redux-form/blob/afa9a9680c426fd745d35e4edd360e36d3439557/src/utils/get-value.js#L31

Final form default implementation of onChange can be found there:
https://github.com/final-form/react-final-form/blob/master/src/Field.js#L117

I removed the Formik demo in the show case and make my PR compatible with these 3 libraries.

I also created a branch to test the compatibility with Formik, Redux and Final Form for each of these libraries:
https://github.com/rodriguezc/primereact/tree/formlibs

cagataycivici:
Thank you, will review later today as soon as beta.4 dev starts.

cagataycivici:
Sorry, could you please update it to support the new Spinner.

rodriguezc:
That's done and ok for me.
But, there is a problem with actual AutoComplete implementation: If it's value is updated from props, the visible input value is not updated. I found it when trying to reset the form. 
In this case, React Form libraries update components props with initial form values.
https://github.com/rodriguezc/primereact/tree/formlibs2

cagataycivici:
Thank you, we'll get into it this week.

cagataycivici:
Merged, thank you very much, AutoComplete case is fixed by the way.

https://github.com/primefaces/primereact/issues/576

raDiesle:
Merge?

rahulsale:
Is this available into the PrimeReact Library and is there any documentation of the same ?

moviapp:
@rahulsale Had any success integrating Primefaces with Formik? Found any documentation?
Having a hard time here...

shayegh:
Hi,
may you give some examples of how this works?

Issue #552, opened 2018-08-29
DataTable's onRowUnselect causes exception


Javalbert:
**I'm submitting a ...**  (check one with "x")
```
[x] bug report
[ ] feature request
[ ] support request => Please do not submit support request here, instead see https://forum.primefaces.org/viewforum.php?f=57
```

**Plunkr Case (Bug Reports)**

https://plnkr.co/edit/RDEla4rh1qTi4LbBNyb2

**Current behavior**

With onRowUnselect set on DataTable, ctrl-clicking the row throws a JavaScript error and the onRowUnselect function is not called.

**Expected behavior**

onRowUnselect function should be called.

**Minimal reproduction of the problem with instructions**

Go to the Plunkr and ctrl-click the row.

* **React version:**

16.4.2

* **PrimeReact version:**

2.0.0-beta.2

* **Browser:** [ Chrome 68 ]

* **Language:** [ ES6/7 ]

cagataycivici:
Will be reviewed for beta.4, thank you.

PR #553, opened 2018-08-29
Fix `onRowUnselect` call

Javalbert:
The fix for https://github.com/primefaces/primereact/issues/552


cagataycivici:
This is fixed on our side as well.

PR #590, opened 2018-09-25
Add support of 'body' Column prop in Treetable

arotnov:
Utilizes column prop 'body' in UITreeRow as well as in Datatable (except first column)


cagataycivici:
TreeTable is being reimplemented so unable to accept as current one will be removed.

Issue #636, opened 2018-11-09
Add currentPageReportTemplate property to Paginator 


superhp:
**I'm submitting an ...** 
```
[ ] bug report
[ ] feature request
[ ] support request 
[x] enhancement request 
```
for Paginator component https://www.primefaces.org/primereact/#/paginator 

**Current behavior**
It is not possible to provide/change CurrentPageReport template which is currently hardcoded as '({currentPage} of {totalPages})'. This enhancement is needed for two cases: 1) when a different presentation way of current page report is needed, 2) when a site is multilingual and a word 'of' has to be translated. 

**Expected behavior**
It should be possible to provide/change template by providing a new currentPageReportTemplate property to Paginator component like this: 
<Paginator ... currentPageReportTemplate='({currentPage} of {totalPages})'></Paginator>

Merve7:
Thank you @superhp for your feedback and PR.

PR #637, opened 2018-11-09
Enhancement #636 Add currentPageReportTemplate property to Paginator

superhp:
**I'm submitting an ...** 
```
[x] enhancement request 
```
for Paginator component https://www.primefaces.org/primereact/#/paginator 

**Current behavior**
It is not possible to provide/change CurrentPageReport template which is currently hardcoded as '({currentPage} of {totalPages})'. This enhancement is needed for two cases: 1) when a different presentation way of current page report is needed, 2) when a site is multilingual and a word 'of' has to be translated. 

**Expected behavior**
It should be possible to provide/change template by providing a new currentPageReportTemplate property to Paginator component like this: 
<Paginator ... currentPageReportTemplate='({currentPage} of {totalPages})'></Paginator>

### DIFF
--- a/src/components/paginator/Paginator.d.ts
+++ b/src/components/paginator/Paginator.d.ts
@@ -16,6 +16,7 @@ interface PaginatorProps {
     style?: object;
     className?: string;
     template?: string;
+	currentPageReportTemplate?: string; 
     leftContent?: JSX.Element | undefined;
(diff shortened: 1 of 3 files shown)


cagataycivici:
Merged manually, thank you.

PR #648, opened 2018-11-21
Smoother scroll with many columns and large datasets.

eranshapira:
Enable hardware acceleration for datatables in order to increase fps.

Using this technique boosted UI performance by a large margin, making the datatable much more attractive to use.

###Defect Fixes

###Feature Requests


cagataycivici:
Table borders disappered in body so had to revert

PR #676, opened 2018-12-13
MultiSelect - Custom label

josmanual:
Added custom label to MultiSelect component.

MultiSelectDemo is also updated with documentation and demo.


cagataycivici:
We've handled differently in 3.0.0-RC1 with fixedPlaceholder prop.

PR #680, opened 2018-12-13
Ability to pass a non-string based label through

phpsa:
###Defect Fixes
When using the Select button with a custom Label type, proptypes throws errors around label needing to be a string and keys being equal.
this allows us to now pass through a renderable to the label and an optional key / better generated key


cagataycivici:
let key = option.key ? option.key : `selb_${index}` is not utilized anywhere?

PR #716, opened 2018-12-21
Added customIcons render prop to Sidebar, to add new toolbar's icons

keul:
Added a small new render props to Sidebar component, to be able to customize the toolbar with additional icons.

schermata 2018-12-21 alle 11 13 41


cagataycivici:
We should offer a icon template or something similar instead of customIcons. Will do for 3.0.1.

PR #748, opened 2019-01-21
Add caption attribute to lightbox

ba2018ds:
If you add a caption attribute to certain lightbox sub-components it is easy to add a image-caption with the css pseudoelement.

For example:
```
[caption]::after {
        content: attr(caption);
        position: absolute;
        left: 0;
        right: 0;
        bottom: 0;
        color: white;
        background: #00000099;
        white-space: nowrap;
        overflow: hidden;
        text-overflow: ellipsis;
 }
```


cagataycivici:
We'll rather add it using templating.

PR #796, opened 2019-03-01
Editor component doc say readonly (not readOnly)

josmanual:


cagataycivici:
It is readOnly;

https://quilljs.com/docs/configuration/

carloshn90:
The error continue in the code, if you go to the editor folder in the file  Editor.d.ts the property name continue as readonly instead of readOny

PR #797, opened 2019-03-04
Update defaultProps in Button.js

JoshK2:
I see that the default label is actually 'p-btn', so I added in the propTypes.


cagataycivici:
It is kinda placeholder and not visible.

PR #798, opened 2019-03-04
Fix: export to csv adding separator header to file

BernatWozzo:
###Defect Fixes
When exporting a CSV file, no header is added to the exported file with the separator.
Now, the chosen separator is added to the file and programs like Excel detect it automatically instead of having to import the data and chose the separator.


cagataycivici:
The PR causes compilation problems so had to revert.

PR #800, opened 2019-03-05
Exclude column from global filter

jeremyrea:
Brings feature parity with PrimeNG (https://github.com/primefaces/primeng/commit/ffb4252d008c9eb1c7f3180cee91839050e6b44d)


cagataycivici:
Good one, thanks.

PR #818, opened 2019-03-21
button extension to work with binary data

flashboss:
This feature allows you to extend the Button component to add an icon in binary format as a property


cagataycivici:
We'll consider if there is more demand, thank you.

PR #824, opened 2019-03-27
Fixed renderItem uses wrong [first] on lazy DataView.

NXJeff:
###Defect Fixes
When using same lazy implementation with DataTable and applied on DataView, the rendering of data does not seems to work correctly when switching between pages. It was caused by the renderItems method in the DataView.js, let say rows = 10, it always took the data[first] which the variable first is followed with the offset of the page, which made the variable first as 10 on the second page, hence, data[10] is undefined, because each lazy load of the data has a maximum number of 10. So by fixing it, I put the props lazy so that it won't affect the implementation without lazyload.

### DIFF
--- a/src/components/dataview/DataView.js
+++ b/src/components/dataview/DataView.js
@@ -101,7 +101,8 @@ export class DataView extends Component {
         style: null,
         className: null,
         itemTemplate: null,
-        onPage: null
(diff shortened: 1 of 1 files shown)


cagataycivici:
Merged manually with added docs, thank you.

PR #863, opened 2019-04-26
Update Editor.js

dmitrsavk:
Add editor modules props


cagataycivici:
The PR misses documentation so unable to merge, we'll add it on our side, thanks.

PR #887, opened 2019-05-15
Fixed #843 by restoring tabindex when tabing out of editor

### LINKED ISSUE #843: Unlogical editor navigation with shift+tab in DataTable
```
[x] bug report
[ ] feature request
```
Bug:
Editor doesn't move in logcal way when using SHIFT + TAB when editing a cell in a data table.

Reproduction:
https://www.primefaces.org/primereact/#/datatable/edit
Click a cell 
press `tab` N times to move editor forward N times. 
Press `Shift+tab` M times, editor is opened at - N -M places from current editor. (Ie one back from initial opening of the editor) 
Press 'tab' again, editor moves to N+1  from initial editor location.  (Ie skipping n+M cells from current edior location)

Expected results: 
Editor always moves +1 or -1 from current editor position. 

Tested with:
FireFox 
Chromium

React version:
Version currently on is being run on primefaces.org and my local env: 16.0.0
PrimeReact version:
Version currently on is being run on primefaces.org and my local env: 2.0.0-rc.1

CylonOven:
Fixed issue #843 
When tabbing out of editor, we don't want to remove the tabindex attr, as that will make it so it is impossible to open the editor with tab ever again. 

Also set the timeout var to this.tabindexTimeout and clear it on the start of the method. 
Reason being in my env sometimes the method is run twice, as when entering the editor the data changes, causing two updates, and due to the timeout the tab-index of 0 is restored, this making it impossible to shift-tab out of the editor as it will always open the current editor.


CylonOven:
Seems that as I did a PR with my master github automatically adds all commits to the PR, that's dumb. Going to close and reopen on another branch of my fork

PR #902, opened 2019-05-25
Fixes #901 - fix calendar parseTime function when showSeconds is false

### LINKED ISSUE #901: Calendar time parsing broken for showSeconds == false
```
[x ] bug report
[ ] feature request
```

Please fork the codesandbox below and create a case demonstrating your bug report. Issues without a codesandbox have much less possibility to be reviewed.

https://codesandbox.io/s/qjx332qq4

**Current behavior**
The `parseTime` function throws an exception when `props.showSeconds` is false. This causes the user input not to be parsed to a Date.

**Expected behavior**
The `parseTime` works when `props.showSeconds` is false.

Just create a Calendar with showTime="true" but showSeconds="false" and try to manually type a date time string in the input field. The bound date value will a string and not a Date because parsing fails.

* **PrimeReact version:**
3.1.3

Mobe91:


cagataycivici:
Thank you.

Issue #908, opened 2019-06-03
ProgressSpinner not visible in IE11


NoFr1ends:
**I'm submitting a ...** 
```
[x] bug report
[ ] feature request
[ ] support request => Please do not submit support request here, instead see https://forum.primefaces.org/viewforum.php?f=57
```

**Current behavior**
The ProgressSpinner is in IE11 invisible

**Expected behavior**
The ProgressSpinner should be visible.

**Minimal reproduction of the problem with instructions**
Just create a react app, add the IE11 polyfills and add a ProgressSpinner.

It can be easily fixed by adding the following css:
```css
.p-progress-spinner-circle {
    /* fix for IE browsers */
    stroke: #d62d20;
    stroke-dasharray: 89,200;
}
```

* **React version:**
16.8.6

* **PrimeReact version:**
3.1.3

* **Browser:** [
IE11
 
* **Language:**
all

PR #909, opened 2019-06-03
Fix invisible ProgressSpinner in IE11

NoFr1ends:
Fixes the invisible progress spinner explained in #908


mertsincan:
Good PR, Thanks a lot ;)

PR #911, opened 2019-06-04
Fix Dialog to allow dialog inside other dialog.

leonardo-machado:
When you need a modal dialog within another dialog, the overlay created in document.body does not allow access to the content.

###Defect Fixes
Calendar compile error.

###Feature Requests
The overlay was created in the parentNode of the dialog container.


mertsincan:
Hi,
Could you please create a Github issue for this? We'll decide it.

Thanks a lot for the PR!

Best Regards,

cagataycivici:
It is not suggested to nest Dialogs.

PR #912, opened 2019-06-05
Allow switching between snow & Bubble toolbar theme

phpsa:
Update to the Editor component to allow passing the prop for theme to cater for snow / bubble,


mertsincan:
Good PR, thanks a lot ;)

PR #919, opened 2019-06-14
Adding support for required, pattern and placeholder props on spinner.

myyellowshoe:
We thought it was interesting that there were no pass-throughs for props on the spinner component. Given that the spinner is slightly more complicated than just basic text input, it seemed that there may have been some reasoning for this. We think long term a props pass-through would be a better idea, but that requires more refactoring that necessary right now. 

We needed this for something we're working on so we just decided to submit a PR to help out others that may need it as well.

This PR adds support for the following props on the Spinner component:
- Required (boolean)
- Placeholder (string)
- Pattern (string)

Let me know what issues/questions you may have.
Thanks for the great lib!

### DIFF
--- a/src/components/spinner/Spinner.d.ts
+++ b/src/components/spinner/Spinner.d.ts
@@ -9,6 +9,9 @@ interface SpinnerProps {
     min?: number;
     max?: number;
     disabled?: boolean;
+    required?: boolean;
+    pattern?: string;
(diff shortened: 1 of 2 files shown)


mertsincan:
Good PR, Thanks a lot ;)

myyellowshoe:
You bet! Thanks! 👍

PR #931, opened 2019-06-19
Add method for replacing messages of Message component

### LINKED ISSUE #930: Add "replace" method in Messages component
```
[ ] bug report
[x] feature request
```

I'd like to propose a new method for the `<Messages />` component to replace messages already being showed.

Imagine you are currently showing 3 messages and you then want to show 2 different messages instead. It is not possible to first do `clear()` and then immediately do `show()` because of `setState` being asynchronous. You can see this here:
https://codesandbox.io/s/primereact-test-1t723

Here it is with a method that works as the new messages are shown once the state change has already happened:
https://codesandbox.io/s/primereact-test-9uh86

I'll create a PR for it and hope you will approve.

nikkop:
By first clearing messages with setState and then
provide a callback to set new messages we ensure
that the state has been updated before we try
to add new messages that are to replace existing
ones.

Fixes #930

### DIFF
--- a/src/components/messages/Messages.js
+++ b/src/components/messages/Messages.js
@@ -59,6 +59,15 @@ export class Messages extends Component {
         })
     }
 
+    replace(messages) {
+        this.setState(
+          {
+            messages: [],
+          },
+          () => this.show(messages)
+        );
+    }
+
     onClose(message) {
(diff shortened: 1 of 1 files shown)


mertsincan:
Good PR ;)

PR #936, opened 2019-06-20
added null check for panel on dropdown

BuckAMayzing:
###Defect Fixes
This is #178 - it looks like it wasn't fixed with the first PR. It looks like "panel" was the null variable, not panel.element.

@mertsincan - mind to take a look?


mertsincan:
Good PR, thanks a lot ;)

Issue #839, opened 2019-04-07
Allow aria-label and aria-labelledby attributes on all input components.


efc:
```
[ ] bug report
[X] feature request
[ ] support request
```

**Current behavior**

I cannot associate `aria-label` or `aria-labelledby` attributes with `Dropdown` or other PrimeReact input components. (This was noted in issue #729, but that issue was closed without addressing this request.)

**Expected behavior**

I would expect that I could define a `Dropdown` and any other PrimeReact input components like this:

```
<Dropdown value='test' editable={true} options={[]} ariaLabel='Test' />
```

or

```
<Dropdown value='test' editable={true} options={[]} ariaLabelledby='testid' />
```

* **React version:**
16.8.6

* **PrimeReact version:**
3.1.2

* **Browser:**
all
 
* **Language:**
all

mertsincan:
Components;
- [x] InputText
- [x] InputTextarea
- [x] DropDown https://github.com/primefaces/primereact/issues/855
- [ ] Calendar
- [ ] MultiSelect

arnauddem:
InputSwitch should also allow aria attributes

ankitprahladsoni:
Will you add this support for `InputNumber` as well? Right now it only has `aria-labelledby` attribute.

jeffgaynorspok:
`aria-*` for accessibility and `data-*` for testing should be passed through on the root of all components.

melloware:
@jeffgaynorspok this feature was added in this ticket: https://github.com/primefaces/primereact/issues/2620

PR #939, opened 2019-06-21
InputSwitch aria-label support

arnauddem:
aria-label need to be on <input> for screen readers

this PR is linked to enhancement:  #https://github.com/primefaces/primereact/issues/839


cagataycivici:
We have ariaLabelledBy in latest codebase

jagobabarron:
ariaLabelledBy in InputSwitch is added to the html checkbox element, but it should be in the div with role="checkbox"

PR #951, opened 2019-07-03
Made spinner formatInput related types optional.

ilnicki:
###Defect Fixes
Updated TypeScript typings for Spinner. **formatInput**, **decimalSeparator** and **thousandSeparator** made optional as it comes from Spinner's sources. All of them are described in 
defaultProps and backed by fallbacks during execution, so there is no need to require them.


mertsincan:
Thanks a lot for the PR!

Best Regards,

PR #960, opened 2019-07-14
Fix DataTable render issue when receiving arrays of array of children

### LINKED ISSUE #959: DataTable fails to render if columns are mapped and has a static column
```
[x] bug report
```

The sandbox example showcasing the error can be found in the following link:
https://codesandbox.io/embed/primereact-test-j7j5j

In often cases, tables are created as Components in a bigger application. Those tables are usually have headers generated, so a common use case would be to map some array to Column components like this:

```javascript
 <DataTable value={this.rows} ...}>
 {
    this.columns.map(x =>  <Column field={x.field} header={x.header} sortable />)
 }
 </DataTable>
```

There can be requirements for such table to include either check-able rows, or a column that every table must have, like operations (show and delete).

```html
 <Column selectionMode="multiple" header="Select all">
```

For the simplicity, in the codesandbox I inlcuded just a new simple column.

**Expected behavior**
The expected outcome would be, that the table renders normally, including all the mapped columns, and the added extra static column. The order of these should not matter. 

**Current behavior**
Currently the component receives the children objects as array of arrays, instead of a single array, containing all the children.
Code recieves:
```javascript
[ [ {Column1},  {Column2} ],  {Static Column} ]
``` 
But expects the children to be flattened like this:
```javascript
[ {Column1},  {Column2},  {Static Column} ]
``` 

This causes an exception in `src\components\datatable\DataTable.js:622` 
```javascript
for(let i = 0; i < this.props.children.length; i++) {
                        if(this.props.children[i].props.footer) {
                                                    ^
```
that the children object is not flattened, so an array object does not have the `props` property. 

The dome is available on : https://codesandbox.io/embed/primereact-test-j7j5j

1. Create a DataTable component
2. Add a mapped array of Column components
3. Renders OK
4. Add a new static column after the mapped
5. Render fails

* **PrimeReact version:**
Occurs in all version of PrimeReact - tested on commit; e24afe442cf833d53591baf57cb0afe24cebf556

**Work Around for those that cannot upgrade**

There is a possible workaround if you only use 1 array to map, so you need to all the static component to it:
```javascript
{
   this.columns.map(x =>  <Column field={x.field} header={x.header} sortable />)
      .concat([<Column field="b" header="b" sortable />]
}
```

I will submit a pull request with the fix.

kovyfive:
**Bug**
Fixed bug when a DataTable component receives array of arrays of children and fails to read footer information

**Defect Fixes**
Fixes #959 

**Details**
Currently the component receives the children objects as array of arrays, instead of a single array, containing all the children.
Code recieves:

`[ [ {Column1},  {Column2} ],  {Static Column} ]`
But expects the children to be flattened like this:

`[ {Column1},  {Column2},  {Static Column} ]`

This causes an exception in `src\components\datatable\DataTable.js:622`
```javascript
for(let i = 0; i < this.props.children.length; i++) {
                        if(this.props.children[i].props.footer) {
                                                    ^
```
that the children object is not flattened, so an array object does not have the props property.

**Fix**
The pull request will introduce the flattenDeep method, and will use it to flatten the children, so any mapping logic can be applied to the children objects.

### DIFF
--- a/src/components/datatable/DataTable.js
+++ b/src/components/datatable/DataTable.js
@@ -619,8 +619,10 @@ export class DataTable extends Component {
             }
             else {
                 if(this.props.children instanceof Array) {
-                    for(let i = 0; i < this.props.children.length; i++) {
-                        if(this.props.children[i].props.footer) {
+                    var flattened =  ObjectUtils.flattenDeep(this.props.children);
+
+                    for(let i = 0; i < flattened.length; i++) {
+                        if(flattened[i].props.footer) {
                             return true;
                         }
                     }
--- a/src/components/utils/ObjectUtils.js
+++ b/src/components/utils/ObjectUtils.js
@@ -132,6 +132,13 @@ export default class ObjectUtils {
         return index;
     }
 
+    static flattenDeep(list) {
+        if(!list){
+            return [];
+        }
+        return list.reduce((acc, val) => Array.isArray(val) ? acc.concat(this.flattenDeep(val)) : acc.concat(val), []);
+    }
+
     static filterConstraints = {
 
         startsWith(value, filter) {


mertsincan:
Thanks a lot for the PR! But, I added null pointer check to the above live and this issue is resolved.

Thanks again.
Best Regards,

Issue #968, opened 2019-07-23
Calendar ButtonBar Clear button does not use clearButtonStyleClass prop


pwojcieszuk:

pwojcieszuk:
https://github.com/primefaces/primereact/pull/969

PR #969, opened 2019-07-23
Fix Calendar ButtonBar clear button class name prop

pwojcieszuk:
https://github.com/primefaces/primereact/issues/968


mertsincan:
Good catch! Thanks a lot for the PR!

PR #982, opened 2019-08-06
dependencies problem solved

antsegcan:
###Defect Fixes
When you are installing and trying to run the server, you get an error from @types dependencies, developers who don't have this dependencies globally installed have this kind of problem. This fix will allow to avoid install globally this ones.


cagataycivici:
Seems to be working with latest create-react-app with no change.

PR #991, opened 2019-08-15
Include parent position in contextmenu calculation

### LINKED ISSUE #990: ContextMenu issue with position calculation
**I'm submitting a ...**
```
[X] bug report
[ ] feature request
[ ] support request
```

https://codesandbox.io/s/primereact-test-gpde3?fontsize=14

**Current behavior**
When having a parent with `position: relative;` somewhere higher up in the DOM tree the context menu will get a wrong position calculated.

**Solution**
The issue is due to the fact that the calculation is based on the `pageX/pageY` of the event which are values relative to the viewport. The calculation should be based on the context menu parent position to the viewport instead.

**Expected behavior**
I expect the context menu to always get a correct alignment regardless of DOM tree layout.

See the Codesandbox above.

* **PrimeReact version:**
3.1.8

NellyHaglund:
Fixes #990

### DIFF
--- a/src/components/contextmenu/ContextMenu.js
+++ b/src/components/contextmenu/ContextMenu.js
@@ -277,8 +277,9 @@ export class ContextMenu extends Component {
 
     position(event) {
         if(event) {
-            let left = event.pageX + 1;
-            let top = event.pageY + 1;
+            const rect = this.container.parentNode.getBoundingClientRect();
+            let left = event.pageX + 1 - rect.left;
+            let top = event.pageY + 1 - rect.top;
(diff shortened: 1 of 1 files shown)


partynikko:
Nice fix! 👏

mertsincan:
Thanks a lot for the PR! But, it breaks the current behavior. You can try this fix with our Tooltip demo. To fix the issue, you can use appendTo attribute. Please try;
```
<ContextMenu model={menu} ref={ref} appendTo={document.body} />
```

Best Regards,

PR #1027, opened 2019-10-01
Flexible panels feature

NathanPB:
I added a feature to the [Panels](https://www.primefaces.org/primereact/#/panel) that makes the body of one or more panel grow or shrink to fit the space of a ``display: flex`` parent.

**Quick Demo:**
Demo

I think I should underscore that it does not affects existing panels, the flex behavior is activated with a ``flex`` boolean property in the panel attributes.

I'll be using that feature in a project, so I hope it helps the entire community.

I'm first-time in this project, so please tell me if I have to do something else, like crate a feature request or something. I read the whole repository and didn't find anything about.

### DIFF
--- a/src/components/panel/Panel.css
+++ b/src/components/panel/Panel.css
@@ -31,3 +31,23 @@
 	padding: .25em .5em;
 	text-align:left;
 }
+
+.p-flex {
+  display: flex;
+  flex-direction: column;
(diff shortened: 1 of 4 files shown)


cagataycivici:
Thank you, I think accordion is a better fit for it using a property like fullHeight.

PR #1033, opened 2019-10-03
Add option for conditionally render

### LINKED ISSUE #1031: Datatable Columns: Conditional select
```
[ ] bug report
[x] feature request
```

**Current behavior**
If you have selectionMode on a column this will always be visible for all columns.

**Expected behavior**
I have encountered use cases where I would want this conditionally per row. Default would be to render it, but having the possibility to conditionally render.

### LINKED ISSUE #1032: DataTable Columns: Conditional reorder
```
[ ] bug report
[x] feature request
```

**Current behavior**
If you have reorder on a column this will always be visible for all columns.

**Expected behavior**
I have encountered use cases where I would want this conditionally per row. Default would be to render it, but having the possibility to conditionally render.

NellyHaglund:
* Optional render for row reorderer
* Optional rander for row selection

Fixes #1031
Fixes #1032

### DIFF
--- a/src/components/column/Column.js
+++ b/src/components/column/Column.js
@@ -41,7 +41,8 @@ export class Column extends Component {
         excludeGlobalFilter: false,
         rowReorder: false,
         rowReorderIcon: 'pi pi-bars',
-        rowEditor: false
+        rowEditor: false,
+        display: null
     }
 
     static propTypes = {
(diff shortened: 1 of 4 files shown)


cagataycivici:
Will do by renaming the prop, thank you.

Issue #1054, opened 2019-10-18
Press Enter key in AutoComplete as a DataTable cell editor should not close the editor when no suggestion is highlighted


sylven32:
**I'm submitting a ...**  (check one with "x")
```
[x] bug report
[ ] feature request
[ ] support request => Please do not submit support request here, instead see https://forum.primefaces.org/viewforum.php?f=57
```

**Current behavior**
Press Enter when no suggestion is highlighted, the editor is closed and the current editing value is lost. This doesn't occur when the AutoComplete is used out of a DataTable editor.

**Expected behavior**
The editor should not close so that the end-user understands he has to select a suggestion.

* **React version:**
16.8.4

* **PrimeReact version:**
3.2.0

* **Browser:** [all]
 
* **Language:** [all]

cagataycivici:
But regular input fields close on enter key, wouldn't be inconsistent? Shouldn't we check for required or something instead?

mertsincan:
Hi,

For now, close. 

If the problem persists, please reopen this ticket.

Best Regards,

PR #1055, opened 2019-10-18
Fix Enter behavior when the AutoComplete is a DataTable cell editor

sylven32:
In case we're in a datable editor, do not close the editor if no suggestion is highlighted.
This will reproduce the same behavior that applies when used out of a DataTable.

###Defect Fixes
https://github.com/primefaces/primereact/issues/1054


cagataycivici:
But regular input fields close on enter key, wouldn't be inconsistent? Shouldn't we check for required or something instead?

Issue #1086, opened 2019-11-19
Month navigation buttons aren't positioned correctly when `headerTemplate` is used.


tomaskikutis:
### There is no guarantee in receiving an immediate response in GitHub Issue Tracker, If you'd like to secure our response, you may consider *PrimeReact PRO Support* where support is provided within 4 business hours

**I'm submitting a ...**  (check one with "x")
```
[x] bug report
[ ] feature request
[ ] support request => Please do not submit support request here, instead see https://forum.primefaces.org/viewforum.php?f=57
```

**Codesandbox Case (Bug Reports)**
Please fork the codesandbox below and create a case demonstrating your bug report. Issues without a codesandbox have much less possibility to be reviewed.

https://codesandbox.io/s/primereact-test-g4qs6

**Current behavior**
Month navigation buttons are rendered on top of `headerTemplate`.

**Expected behavior**
Month navigation buttons should be aligned to a month name label and not overflow on top of `headerTemplate`.
**Minimal reproduction of the problem with instructions**

**Please tell us about your environment:**

* **React version:**
16.8.6

* **PrimeReact version:**
3.3.2

* **Browser:** Chromium. Will probably reproduce in all due to the DOM strructure.
 
* **Language:** [all | TypeScript X.X | ES6/7 | ES5]
Typescript 3.5.1
![screenshot_16](https://user-images.githubusercontent.com/2076953/69158361-98c49f80-0ae6-11ea-95f9-5743966be9cf.png)

mertsincan:
Thanks a lot for the PR! I'll check it on all Premium Layouts and get back to you.

cagataycivici:
Not sure, @mertsincan wdyt?

cagataycivici:
We're refactoring our css for upcoming 5.0 so we'll handle it during the refactor, thank you.

PR #1087, opened 2019-11-19
Fix positioning of month navigation buttons when `headerTemplate` is used.

tomaskikutis:
Fixes https://github.com/primefaces/primereact/issues/1086


cagataycivici:
Not sure, @mertsincan wdyt?

cagataycivici:
We're refactoring our css for upcoming 5.0 so we'll handle it during the refactor, thank you.

Issue #1095, opened 2019-11-25
Growl messages from "bottom*" is not displaying as expected


PaulInglis:
### There is no guarantee in receiving an immediate response in GitHub Issue Tracker, If you'd like to secure our response, you may consider *PrimeReact PRO Support* where support is provided within 4 business hours

**I'm submitting a ...**  (check one with "x")
```
[X] bug report
[ ] feature request
[ ] support request => Please do not submit support request here, instead see https://forum.primefaces.org/viewforum.php?f=57
```

**Codesandbox Case (Bug Reports)**
Please fork the codesandbox below and create a case demonstrating your bug report. Issues without a codesandbox have much less possibility to be reviewed.

https://codesandbox.io/s/primereact-growltest-29idu

**Current behavior**

Setting the `position` prop of the `Growl` component displays the messages in the top area still.

**Expected behavior**

The messages should be displayed in the lower area of the screen/display area

**Minimal reproduction of the problem with instructions**

The codesandbox link above should give a good enough reproduction.  If I remove `top: 70px;` from `body .p-growl` in the styles then it works as desired.

**Please tell us about your environment:**

Windows 10, Visual Code, CRA

* **React version:**

16.9.0

* **PrimeReact version:**

* **Browser:** [all | Chrome XX | Firefox XX | IE XX | Safari XX | Mobile Chrome XX | Android X.X Web Browser | iOS XX Safari | iOS XX UIWebView | iOS XX WKWebView ]

All 

* **Language:** [all | TypeScript X.X | ES6/7 | ES5]
Typescript

PR #1096, opened 2019-11-25
Fix for #1095 - Growl messages from "bottom*" is not displaying as expected

PaulInglis:
Unsetting the `top` attribute for those relating to `bottomleft` or `bottomright` positioning


PaulInglis:
Fixed from the issue

PR #1143, opened 2020-01-10
fix InputTextarea rows cols issue

### LINKED ISSUE #1127: InputTextarea does not render props like cols and rows
**[X] bug report**
[ ] feature request
[ ] support request 

**Current behavior**
When we use a InputTextarea with props that the component also has as "default"
(e.g. autoResize,  onInput,  cols, rows,  tooltip and tooltipOptions) those sent props are never used.

Example:
`<InputTextarea rows={15} cols={30} value={this.state.value} onChange={(e) => this.setState({value: e.target.value})} />`

This happens because the line
`let textareaProps = ObjectUtils.findDiffKeys(this.props, InputTextarea.defaultProps);`
filters the props that have defaults.
See https://github.com/primefaces/primereact/blob/master/src/components/inputtextarea/InputTextarea.js

**Expected behavior**
Those properties should be considered.
e.g. If you say rows={15} the component should render a inputarea with 15 rows.

* **React version:**
16.12.0

* **PrimeReact version:**
3.3.3

gulgunp:
###Defect Fixes

Fixes #1127

### DIFF
--- a/src/components/inputtextarea/InputTextarea.js
+++ b/src/components/inputtextarea/InputTextarea.js
@@ -154,8 +154,17 @@ export class InputTextarea extends Component {
         let textareaProps = ObjectUtils.findDiffKeys(this.props, InputTextarea.defaultProps);
 
         return (
-            <textarea {...textareaProps} className={className} ref={input => this.element = input} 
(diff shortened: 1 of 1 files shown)


mertsincan:
Thanks a lot for the PR! But, it would be better if I remove rows and cols properties.

Issue #1094, opened 2019-11-25
Deadlock situation in range slider


Sanbala:
I'm submitting a ... (check one with "x")

[x] bug report
[ ] feature request
[ ] support request => Please do not submit support request here, instead see https://forum.primefaces.org/viewforum.php?f=57

Current behavior
When both the left & right thumb are moved to the right most part the slider, after mouse out thumb is not movable. It becomes a deadlock situation

Expected behavior
Thumb should be movable.

Minimal reproduction of the problem with instructions
https://codesandbox.io/s/primereact-zwof8
The issue is also reproducible in the demo examples given in the official documentation

PrimeReact version:
Latest version

Browser: [all ]

Language: [all | TypeScript | ES6]

Sanbala:
the provided fix didnt fix the issue completely. Now there is a deadlock in left side when both the handles are in min position.

PR #1147, opened 2020-01-13
#1094 - Deadlock situation in range slider

oytuntutun:
Handles seperated. when min handle reaches 50, min handle overlaps the max handle.


oytuntutun:
@mertsincan @cagataycivici closing this PR due to [this commit](https://github.com/primefaces/primereact/commit/31190bd87222acca4916e2fe24c9e3edf211a4a8?diff=unified)

PR #1153, opened 2020-01-16
fix Slider deadlock situation issue

### LINKED ISSUE #1094: Deadlock situation in range slider

gulgunp:
###Defect Fixes

Fixes #1094

### DIFF
--- a/src/components/slider/Slider.js
+++ b/src/components/slider/Slider.js
@@ -239,12 +239,19 @@ export class Slider extends Component {
     renderRangeSlider() {
         let values = this.props.value||[0,0];
         let horizontal = (this.props.orientation === 'horizontal');
+        let rangeEndHandle;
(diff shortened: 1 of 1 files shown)


mertsincan:
Thanks a lot for the PR! I think we can add z-index to the start handle.

Issue #1159, opened 2020-01-17
PanelMenu: expandedKeys property 


Rainson12:
**I'm submitting a ...**  (check one with "x")
```
[ ] bug report
[X] feature request
[ ] support request => Please do not submit support request here, instead see https://forum.primefaces.org/viewforum.php?f=57
```

**Current behavior**
Currently there is no way to initally set the expanded menu. This functionality is available for primeNg

**Expected behavior**
additional property on model items called "expanded"

Rainson12:
This is being fixed by https://github.com/primefaces/primereact/pull/1160

mertsincan:
The 'expandedKeys' property can be added like in primevue;
https://primefaces.org/primevue/showcase/#/panelmenu

I'll work on it and get back to you.
Best Regards,

melloware:
cc @arkadiusztw any interest in this one too?

FrootLoopsR:
Hi,
is this still under development?
will help a lot !

habubey:
Issue tracker is used for defects only as part of our commitment to quality and continuous improvement in all areas. Enhancements are collected as valuable community feedback and managed internally so moving this enhancement ticket to our internal project management backlog.

PR #1166, opened 2020-01-23
added expanded property to panelmenu items

Rainson12:
this fixes https://github.com/primefaces/primereact/issues/1159


cagataycivici:
We rather not manipulate the model and keep it at PanelMenu using something like expandedMenuitems or so.

Issue #1175, opened 2020-01-31
Implement className in MultiSelect and Dropdown options


Hafas:
**I'm submitting a ...**  (check one with "x")
```
[ ] bug report
[x] feature request
[ ] support request => Please do not submit support request here, instead see https://forum.primefaces.org/viewforum.php?f=57
```
**Current behavior**
Every item in a `MultiSelect` has the same class name `p-multiselect-item`. No way to change that

**Feature request**
I'd like to be able to set custom class names to each item.

**Suggestion**
Similar to the prop `optionLabel` I'd like to introduce a new prop `optionClassName`. Just like `optionLabel` it is either a `function` or a `string` that points to the key of the option's `Object`.

Example:

```js
render () {
  const options = [
    { label: "New York", value: "NY", className: "green" },
    [ label: "Rome", value: "RM", className: "red" }
  ]
  return (
    <MultiSelect options={options} optionClassName="className" />
  );
}
```

Result:
```html
<ul class="p-multiselect-items p-multiselect-list p-component" role="listbox" aria-multiselectable="true">
  <li class="green p-multiselect-item" tabindex="0" role="option" aria-selected="false"><!-- ... --></li>
  <li class="red p-multiselect-item" tabindex="0" role="option" aria-selected="false"><!-- ... --></li>
</ul>
```

Implementation:
```js
// Multiselect.js

// define new class method
getOptionClassName(option) {
    return ObjectUtils.resolveFieldData(option, this.props.optionClassName);
}

// in the render method:
items = items.map((option, index) => {
    let optionLabel = this.getOptionLabel(option);
    let optionClassName = this.getOptionClassName(option);

    return (
        <MultiSelectItem key={optionLabel + '_' + index} label={optionLabel} option={option} template={this.props.itemTemplate} className={optionClassName}
        selected={this.isSelected(option)} onClick={this.onOptionClick} onKeyDown={this.onOptionKeyDown} tabIndex={this.props.tabIndex} />
    );
});

```
```js
// MultiSelectItem.js
// in the render method:
const className = classNames(this.props.className, 'p-multiselect-item', {'p-highlight': this.props.selected});
```

I can provide a proper pull request if you accept this suggestion.

master117:
I'm not sure if I understand this. MultiSelect takes an array of SelectItems. SelectItem contains an optional className field:

```
interface SelectItem {
    label?: string;
    value: any;
    className?: string;
    icon?: string;
    title?: string;
    disabled?: boolean;
[...]

PR #1177, opened 2020-02-01
Added className to MultiSelectItem

master117:
Added className to MultiSelectItem, so classNames as specified in the SelectItem api work on MultiSelect.

###Defect Fixes
I believe this Fixes:
https://github.com/primefaces/primereact/issues/1175

###Feature Requests

### DIFF
--- a/src/components/multiselect/MultiSelect.js
+++ b/src/components/multiselect/MultiSelect.js
@@ -387,6 +387,10 @@ export class MultiSelect extends Component {
         return this.props.optionLabel ? ObjectUtils.resolveFieldData(option, this.props.optionLabel) : option.label;
     }
 
+    getOptionClassName(option) {
+        return this.props.optionLabel ? ObjectUtils.resolveFieldData(option, this.props.className) : option.className;
+    }
+
     isEmpty() {
         return !this.props.value || this.props.value.length === 0;
     }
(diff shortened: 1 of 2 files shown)


src/components/multiselect/MultiSelect.js
```diff
@@ -387,6 +387,10 @@ export class MultiSelect extends Component {
         return this.props.optionLabel ? ObjectUtils.resolveFieldData(option, this.props.optionLabel) : option.label;
     }
 
+    getOptionClassName(option) {
+        return this.props.optionLabel ? ObjectUtils.resolveFieldData(option, this.props.className) : option.className;
```
Hafas: `this.props.className`  should be replaced with a new prop like `this.props.optionClassName` and default it to `"className"` if not set.

I'm also not sure if it is wise to make `optionLabel` necessary to make the new prop `optionClassName` have an effect.

master117: You are right, optionsLabel requirement should be dropped, do you want to add another review?

Hafas: > do you want to add another review?

Sorry. I don't understand the question.

But as I said `this.props.className` in `MultiSelect.js` is inappropriate, because it is used for the actual class name for the main component. A new prop is required.

If we add a new prop then we also need to update `MultiSelect.d.ts`.

cagataycivici:
Will be fixed via https://github.com/primefaces/primereact/pull/1211

PR #1195, opened 2020-02-14
Outlined buttons

NathanPB:
I'm not a designer but I personally love outlined buttons. I was in need of this feature while building a personal project and I noticed PrimeReact does not have such feature, so I implemented it. Everything is already documented and the showcase is done!

Screenshot_20200214_003341

The colors in the "Secondary" button aren't working very well, and I honestly don't know what to do because its theme based. I'm asking for help if someone know how to improve this.

There is a little trick to set the text color of the button, because the themes doesn't provide any kind of global CSS variable to tell the rest of the framework its color scheme (if it does please tell me and I will improve, but I couldn't manage to find such behavior). The tricky consists in getting the computed styles of the element after its rendered for the first time, and then providing the computed ``border-color`` property to the component's state, creating the need of re-render the component to change the text color.

To make outline buttons its required to provide a ``p-button-outlined`` class into the component's ``className``. It also works well with the other styles.

Screenshot_20200214_004324

Well, hope it helps the project. I understand if the PR couldn't get accepted because of the company bureaucracy, but if it doesn't get accepted due bad practices of my code please tell me what, so I can improve it and possibly help other new contributors.

Thank you

### DIFF
--- a/src/components/button/Button.css
+++ b/src/components/button/Button.css
@@ -116,5 +116,4 @@ button.p-button::-moz-focus-inner {
     -moz-box-sizing: border-box;
     -webkit-box-sizing: border-box;
     background-color: transparent !important;
-    color: black !important;
 }
--- a/src/components/button/Button.css
+++ b/src/components/button/Button.css
@@ -109,3 +109,12 @@ button.p-button::-moz-focus-inner {
         width: 100%;
     }
 }
+
+/** Outlined **/
+.p-button.p-button-outlined {
+    box-sizing: border-box;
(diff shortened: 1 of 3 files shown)


cagataycivici:
This is handled by the PrimeReact designer per theme, each theme defines;

```sass
$focusOutlineColor:lighten($primaryColor, 35%);
$focusOutline: 0 none;
$focusOutlineOffset: 0;
$focusShadow: 0 0 0 0.2em $focusOutlineColor;
```
Since this can easily be changed by the designer, unable to accept the PR, thank you.

PR #1197, opened 2020-02-14
Fixed show/hide triggers being triggered even though already shown/hidden

### LINKED ISSUE #1189: ContextMenu hide/onHide triggers even though already hidden
```
[x] bug report
[ ] feature request
```

https://codesandbox.io/s/primereact-contextmenu-onhide-bug-ee7c9

**Current behavior**
`hide` and `onHide` are triggered even though the ContextMenu is not visible/already been hidden.

**Expected behavior**
`hide` and `onHide` should only be triggered if the ContextMenu is visible.

Check the CodeSandbox. Bringing up the ContextMenu and then clicking anywhere after it is hidden will continue to call `hide` and `onHide`.

* **React version:**
16.8

* **PrimeReact version:**
3.4.0

NathanPB:
### Defect Fixes
Fixes #1189 

Using a boolean state instead of directly modifying the DOM to perform checks if the component is already hidden/shown

### DIFF
--- a/src/components/contextmenu/ContextMenu.js
+++ b/src/components/contextmenu/ContextMenu.js
@@ -217,7 +217,8 @@ export class ContextMenu extends Component {
     constructor(props) {
         super();
         this.state = {
-            resetMenu: false
+            resetMenu: false,
+            hidden: true
         }
         this.onMenuClick = this.onMenuClick.bind(this);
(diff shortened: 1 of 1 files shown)


cagataycivici:
documentClickListener should be removed which would be simpler as it is there as a leak.

Issue #1193, opened 2020-02-13
Datatable: Reorder creates duplicate columns


master117:
**I'm submitting a ...**  (check one with "x")
```
[x] bug report
[ ] feature request
[ ] support request => Please do not submit support request here, instead see https://forum.primefaces.org/viewforum.php?f=57
```

**Codesandbox Case (Bug Reports)**
https://codesandbox.io/s/primereact-test-e6t3n

**Current behavior**
When Reordering a column more columns appear. This may be related to the columns field, but I keep getting this error even when all columns have fields.

**Expected behavior**
The usual reorder behavior

**Minimal reproduction of the problem with instructions**
https://codesandbox.io/s/primereact-test-e6t3n
Drag one of the columns to reorder, more columns appear

* **PrimeReact version:**
3.4.0
 
* **Language:** 
TypeScript

NathanPB:
I noticed special not directly mentioned condition to reproduce this: The ``Datatable`` needs one or more ``Column``s with repeated ``field`` property. The duplicated property is the one with repeated field. If there is more than two distinct repeated fields, so (as long as I could reproduce) the first duplicated ``field`` in the ``Column`` order gets duplicated.

master117:
I have seen this also happen when fields are unique but a template is used on the cells. Aka the fields value does not match the column content.

mertsincan:
Please try columnKey instead of key property. 

> I have seen this also happen when fields are unique but a template is used on the cells. Aka the fields value does not match the column content.

After I change key with columnKey property, I couldn't replicate this issue. Could you please try it again?

Best Regards,

master117:
After using columnKey the issue vanishes. Maybe columnKeys should contains keys if otherwise empty. Thanks.

cagataycivici:
@NathanPB added a PR to check the column keys for 4.1.0 as well.

PR #1198, opened 2020-02-18
Added warnings to columnKeys in reorderable DataTables

NathanPB:
I was investigating #1193 and found out that there is a mess in the column identifiers, because it can be both ``columnKey`` (that can be omitted) or ``field`` (that can be duplicated), and the duplication **and** omission of the cited properties at the same ``Column`` cause the duplication of the column array on reorder actions, at the following line: https://github.com/primefaces/primereact/blob/ce2d51a6469d5e240e4c199d924f67b71e96ac37/src/components/datatable/DataTable.js#L1274

I just added a warning when a ``DataTable`` is constructed with ``reorderableColumns`` as ``true`` **and** some ``Column`` has a omitted ``dataKey``.

I don't know about PrimeReact's plans with this behavior, but I suggest removing the functionality of ``field`` property to act as a identifier, it's just too problematic.

https://github.com/primefaces/primereact/blob/ce2d51a6469d5e240e4c199d924f67b71e96ac37/src/components/datatable/DataTable.js#L948

### DIFF
--- a/src/components/datatable/DataTable.js
+++ b/src/components/datatable/DataTable.js
@@ -230,6 +230,10 @@ export class DataTable extends Component {
         this.onColumnDrop = this.onColumnDrop.bind(this);
         this.onVirtualScroll = this.onVirtualScroll.bind(this);
         this.frozenSelectionMode = null;
+
+        if (this.props.reorderableColumns && !this.props.children.every(it => it.columnKey)) {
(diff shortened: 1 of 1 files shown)


cagataycivici:
I think the info is already in the docs but it seems like an easy mistake to do so a log will help, thanks.

Issue #1206, opened 2020-02-26
RowGrouping, ExpandableRows Selects All Rows in Group


jmeberst:
### There is no guarantee in receiving an immediate response in GitHub Issue Tracker, If you'd like to secure our response, you may consider *PrimeReact PRO Support* where support is provided within 4 business hours

**I'm submitting a ...**  (check one with "x")
```
[X ] bug report
[ ] feature request
[ ] support request => Please do not submit support request here, instead see https://forum.primefaces.org/viewforum.php?f=57
```

**Codesandbox Case (Bug Reports)**
Please fork the codesandbox below and create a case demonstrating your bug report. Issues without a codesandbox have much less possibility to be reviewed.

https://codesandbox.io/s/primereact-demo-26o2w

**Current behavior**

When selection (single or multiple) and rowgrouping, expandableRows are enabled, selecting a single row under a group will select all rows under the group. 

**Expected behavior**

This bug is useful, because it would be desirable to have a feature for selecting all rows in the group, but I also want the ability to select only a single row in the group.

**Minimal reproduction of the problem with instructions**

Expand any row with multiple entries, then select the checkbox of any item. All rows in the group will be selected. Only the checked one should be selected

**Please tell us about your environment:**

Code Sandbox template

* **React version:**

16.12

* **PrimeReact version:**

3.4.0

* **Browser:** [all | Chrome XX | Firefox XX | IE XX | Safari XX | Mobile Chrome XX | Android X.X Web Browser | iOS XX Safari | iOS XX UIWebView | iOS XX WKWebView ]

Chrome, Firefox, Edge
 
* **Language:** [all | TypeScript X.X | ES6/7 | ES5]
ES6

jmeberst:
I added a pull request to fix this issue.
https://github.com/primefaces/primereact/pull/1209

Hopefully that is a good start.

PR #1209, opened 2020-02-27
Expandable row selection fix

jmeberst:
###Defect Fixes
https://github.com/primefaces/primereact/issues/1206

I tried to take a light touch on fixing this issue.

There are two problems

1.  When the selection attempts to do a comparison it uses only the field set in data key which will be equal for all grouped items.
2. If using checkboxes with multiple selections there won't be a metakey, so this update is needed to keep the multiselections spread correctly.

###Feature Requests


cagataycivici:
Fixed with a different approach

PR #1211, opened 2020-02-29
Add className to MultiSelectItem and DropdownItem

Hafas:
Solves #1175 for `MultiSelect` and `Dropdown` components. Each item may now have its own `className`

An other pull request (#1177) already exists but it is incomplete.


cagataycivici:
I'll just use the className of option instead.

PR #1287, opened 2020-03-25
Fix #1286: Improved ObjectUtils.equals method

### LINKED ISSUE #1286: Not nullable `dataKey` in DropDown break `selectedOption` property
```
[x] bug report
[ ] feature request
```
**Current behavior**
Next compiled code of Dropdown component doesn't work properly
```
key: "findOptionIndex",
    value: function findOptionIndex(value) {
      var index = -1;

      if (this.props.options) {
        for (var i = 0; i < this.props.options.length; i++) {
          var optionValue = this.getOptionValue(this.props.options[i]);

          if (value === null && optionValue == null || _ObjectUtils.default.equals(value, optionValue, this.props.dataKey)) {
            index = i;
            break;
          }
        }
      }

      return index;
    }
```
if `value` and `optionValue` are primitive strings and `dataKey` is not null which is a not rare case I think. In this scenarion `selectedOption` in render will be not correct.

**Expected behavior**
Selected option according to value prop.

The PR is ready actually. I just wanted to create issue to follow your procedures.

m1n0s:
Fixing #1286

### DIFF
--- a/src/components/utils/ObjectUtils.js
+++ b/src/components/utils/ObjectUtils.js
@@ -1,7 +1,7 @@
 export default class ObjectUtils {
     
     static equals(obj1, obj2, field) {
-        if(field)
+        if(field && obj1 && typeof obj1 === 'object' && obj2 && typeof obj2 === 'object')
             return (this.resolveFieldData(obj1, field) === this.resolveFieldData(obj2, field));
         else
             return this.deepEquals(obj1, obj2);


m1n0s:
Hi @cagataycivici! Sorry for direct mention but I see you merging master in this repo and we really need this fix to be done ASAP. Can you assist please?

mertsincan:
Hi @m1n0s,
Thanks a lot for the PR ;)

PR #1322, opened 2020-04-14
Disable onChange event if toggle button disabled

### LINKED ISSUE #1321: Disable prop on ToggleButton not working
```
[x] bug report
[ ] feature request
```

**Current behavior**
`disabled` prop on ToggleButton component not disabling onclick on the button, but only applies disabled style.

**Expected behavior**
As well as applying disabled styling on the button, it should disable any actions connected to the button.

NellyHaglund:
Fixes #1321


cagataycivici:
It also receives focus so need to apply a bit more.

PR #1345, opened 2020-04-26
Fix: when I'm using react-hook-form, I can't benefit from label position toggling since I don't pass value

CYB3RL1F3:
###Defect Fixes
When no value passed, check whether value is present in the input or when an input event is dispatched. 
Opens possibility to use primereact input text with react-hook-form

### DIFF
--- a/accordion.d.ts
+++ b/accordion.d.ts
@@ -1,0 +1,1 @@
+export * from './components/accordion/Accordion';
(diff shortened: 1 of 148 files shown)


CYB3RL1F3:
Finally since no merge or review possible from the owner, I'll publish a separate fork that works.

Issue #1346, opened 2020-04-28
Editor Component readOnly property name wrong


carloshn90:
There is an error in the editor component, the readOnly property is wrong in the file Editor.d.ts, should be readOnly instead of readonly

```
[x] bug report
[ ] feature request
[ ] support request => Please do not submit support request here, instead see https://forum.primefaces.org/viewforum.php?f=57
```

PR #1347, opened 2020-04-28
Editor Component readOnly property name wrong

carloshn90:
Ref: #1346


mertsincan:
Thanks a lot for the PR!

PR #1371, opened 2020-05-16
Fix isSelectableTime()

alexshawnee:
Whole array passed instead of single Date item


mertsincan:
Thanks a lot for the contribution!

Best Regards,

PR #1376, opened 2020-05-20
Fix access to undefined first property in TableBody.js:143

schernoudov:
###Defect Fixes
Fix access to undefined first property in TableBody.js:143


mertsincan:
Sorry for the delayed response. I think this has been fixed with the latest version.

Best Regards,

PR #1400, opened 2020-06-09
deprecated installation option --save npm removed and included  yarn add

thiagoalvesfoz:
`npm install` no longer has a --save option, and as of version 5 of npm this option is no longer needed as it is performed by default!

more details [NPM v5.0.0](https://blog.npmjs.org/post/161081169345/v500 "NPM v5.0.0")

This PR removes --save from the suggested npm install command
and add the yarn equivalent command


mertsincan:

Best Regards,

PR #1414, opened 2020-06-17
Fix file selection not working after validation fails

nattukan0713:
```
[x] bug report
[ ] feature request
```

**Current behavior**
If the mode property is "basic", file selection will not work after validation fails.

**Expected behavior**
Disable file selection only if the validation is successful.

It is possible to reproduce the problem using an example for "Basic" in showcase.(Upload of a file larger than 1,000,000 bytes reproduces this problem.)

https://www.primefaces.org/primereact/showcase/#/fileupload

* **React version:**
16.13.1

* **PrimeReact version:**
4.2.2

Browser: Chrome  83.0.4103.106 but I think it will happen in any other browser
 
JS

### DIFF
--- a/src/components/fileupload/FileUpload.js
+++ b/src/components/fileupload/FileUpload.js
@@ -182,7 +182,7 @@ export class FileUpload extends Component {
             this.clearInputElement();
         }
 
-        if(this.props.mode === 'basic') {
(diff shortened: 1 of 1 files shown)


nattukan0713:
#1415

mertsincan:
Thanks a lot for the PR!

PR #1429, opened 2020-07-04
Fix not be directly mutated component’s state in FileUpload

nattukan0713:
###Defect Fixes
Component’s state has been directly mutated in FileUpload.

**References**
"state is a reference to the component state at the time the change is being applied. It should not be directly mutated" 
is specified in the official documentation of the react.
[setState](https://reactjs.org/docs/react-component.html#setstate)


mertsincan:

Best Regards,

PR #1460, opened 2020-07-18
Fix incorrect behavior during file upload - #1525

### LINKED ISSUE #1525: FileUpload: Disable buttons while uploading
[x] bug report

### Problem
Parallel upload is not considered, but it can be operated during upload. Therefore, the following problems occur.

- **If you upload while uploading, the files to be uploaded are uploaded in duplicate.**

  **Reproduction procedure**
Click the upload button many times during the upload.

- **The progress bar will be displayed more or less as the progress of multiple uploads is received in one state.**

  **Reproduction procedure (Advanced mode only)**
You can easily check by uploading large files in succession.

- **When selecting a file during upload, the files parameter of onUpload(), onError() may be different from the file that was actually uploaded or failed to upload.**

### Defect Fixes
**The current implementation only supports a single upload, so operations during upload are disabled.**

Supplement

- If the auto property is true, uploading will start after file selection, so disable file selection during uploading.

- The files selected during upload are reset after the upload is complete, so the file selection feature during upload is useless.

- Since the function of the cancel button is to reset the selected state of the file and it is useless during uploading, the cancel button during uploading is disabled.

- You may confuse that you can cancel the upload if the Cancel button is enabled during the upload.

nattukan0713:
Fix #1525
### Problem
Parallel upload is not considered, but it can be operated during upload. Therefore, the following problems occur.

- **If you upload while uploading, the files to be uploaded are uploaded in duplicate.**

  **Reproduction procedure**
Click the upload button many times during the upload.

- **The progress bar will be displayed more or less as the progress of multiple uploads is received in one state.**

  **Reproduction procedure (Advanced mode only)**
You can easily check by uploading large files in succession.

- **When selecting a file during upload, the files parameter of onUpload(), onError() may be different from the file that was actually uploaded or failed to upload.**

### Defect Fixes
**The current implementation only supports a single upload, so operations during upload are disabled.**

Supplement

- If the auto property is true, uploading will start after file selection, so disable file selection during uploading.

- The files selected during upload are reset after the upload is complete, so the file selection feature during upload is useless.

- Since the function of the cancel button is to reset the selected state of the file and it is useless during uploading, the cancel button during uploading is disabled.

- You may confuse that you can cancel the upload if the Cancel button is enabled during the upload.


mertsincan:
Thanks a lot for the new ticket and PR!

Best Regards,

mertsincan:
For now, reverted. I have to do more tests on this PR.

PR #1471, opened 2020-07-29
Add new messages to Growl in the same render cycle

DanielCBezerra:
Function show in Growl is using old state values when adding new messages.

This will ignore messages previously added in the same render cycle.


mertsincan:
Thanks a lot! Fixed for Toast component.

Best Regards,

Issue #1112, opened 2019-12-03
Customize Multiselect Header


sharmillamuthu:
**Current behavior**
In the multiselect header part there is no text / props to display text is provided. As of now, only filter props which adds filter in Multiselect  header section is provided.

![image](https://user-images.githubusercontent.com/53299877/70024456-25c82800-15c0-11ea-9027-9dd123eb367e.png)

![image](https://user-images.githubusercontent.com/53299877/70042835-99c8f700-15e5-11ea-8f85-e300ba7763d0.png)

**Expected behavior**
Provide a way to add custom text (or) adding template to the Multiselect header part

![screenshot](https://user-images.githubusercontent.com/53299877/70024751-04b40700-15c1-11ea-8576-c5d5fad4a59b.png)

mertsincan:
Hi,

Added panelHeaderTemplate property to MultiSelect. Please see;
https://github.com/primefaces/primereact/issues/1841

You can customize all header and footer;) 

Best Regards,

PR #1476, opened 2020-07-31
Refactor #1112 - Add prop to support customisable Multiselect header

praveenkumar-kalidass:
Refactor #1112 

Added `optionHeader` prop to take customisable header text.
Default text will be null, same as in current scenario.
If filter is enabled, header text will be overridden.


mertsincan:
Hi,
Thanks a lot for your contribution. Fixed in https://github.com/primefaces/primereact/issues/1841

Best Regards,

PR #1480, opened 2020-08-06
[Card] Allow React.ReactNode in title and subtitle

NoFr1ends:
Often it would be nice to embed icons etc inside a card title or subtitle.
Currently these properties are marked as string but would also work fine if they are a ReactNode.

Using ``//@ts-ignore`` works but also generates a warning on the console which is not that great.


mertsincan:
Hi,

Sorry for the delayed response! This feature is supported in the latest version. Could you please try it?

For now, closed. If the problem persists, please reopen this PR.
Best Regards,

PR #1537, opened 2020-08-29
Fixed #1535

### LINKED ISSUE #1535: Wrong scroll of big Dropdowns inside Dialogs
```
[ x ] bug report
[ ] feature request
```

https://codesandbox.io/s/primereact-test-forked-x0mqq?file=/src/index.js

**Current behavior**
When a Dropdown with a long list of options is shown inside a Dropdown the scroll of both the components acts weird, the overlay is limited by the Dialog area and a bunch of scrollbars appears

image

**Expected behavior**
The overflow of the Dropdown component should not be limited by the Dialog

Just open the attached codesandbox and click in the Dropdown

* **React version:** 16.13.1

* **PrimeReact version:** 5.0.0-rc.1

* **Browser:** all
 
* **Language:** all

NathanPB:
### Defect Fixes
Fixed #1535 

This is a very common CSS limitation, unfortunately I couldn't find a more elegant solution of doing this other by wrapping the Dropdown's overlay in a chain of ``div``, ``div`` again (with position absolute), and another ``div`` (with position fixed).

The overlay is only wrapped when the component is inside a Dialog, and this is known by iterating down the DOM tree searching for a parent element with the class ``p-dialog-content``. If found, the overlay will be wrapped into that 3 div chain I mentioned.

Not an elegant solution, but now I'm able to see the entire dropdown content when inside a dialog, and the behavior when its not is just the same as before

https://github.com/w3c/csswg-drafts/issues/4092

### DIFF
--- a/src/components/dropdown/DropdownPanel.js
+++ b/src/components/dropdown/DropdownPanel.js
@@ -23,10 +23,38 @@ export class DropdownPanel extends Component {
         onClick: PropTypes.func
     };
 
+    isInsideDropdown() {
+        let component = ReactDOM.findDOMNode(this);
+        while (component) {
+            if (component.className?.includes('p-dialog-content')) {
+                return true;
+            }
+            component = component.parentNode;
+        }
+        return false;
+    }
+
     renderElement() {
         let className = classNames('p-dropdown-panel p-component', this.props.panelClassName);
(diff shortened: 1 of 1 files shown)


mertsincan:
Thanks a lot! Please see my comment in #1535 

Best Regards,

PR #1572, opened 2020-09-10
Fix #1453

### LINKED ISSUE #1453: Pagination (currentPageReportTemplate) is not showing correct page no for first
### Pagination is not showing correct page no.

**When there are no records available for table then its showing no records in table body but the pagination is not correct**

Showing 1 to 0 of 0 entries

```
currentPageReportTemplate : Showing {first} to {last} of {totalRecords} entries
```

{last} and {totalRecords} are updating properly but {first} is set to 1 default so this not correct in case of 0 records

omrfrks:


mertsincan:
Thanks a lot for the PR!

Best Regards,

Issue #1404, opened 2020-06-10
findDOMNode is deprecated in StrictMode


yanai101:
this is the error message that I have  in the console: (using ts )
```
Warning: findDOMNode is deprecated in StrictMode. findDOMNode was passed an instance of Transition which is inside StrictMode. Instead, add a ref directly to the element you want to reference. Learn more about using refs safely here: https://fb.me/react-strict-mode-find-node
    in div (created by GrowlMessage)
    in GrowlMessage (created by Growl)
    in Transition (created by CSSTransition)
    in CSSTransition (created by Growl)
...
```

this is a code sandbox:
https://codesandbox.io/s/vibrant-haze-pxqw6

the error is trigger when the Message/ Growl component  change is inside state ...

Thanks

ckay2003:
Seems to happen in multiple components. I get this message when using a calendar as a custom editor in datatable

mertsincan:
Hi,

It is a warning, not an error. Also, this issue is related to transition API. Please see; https://github.com/reactjs/react-transition-group/issues/429

Also, https://github.com/facebook/react/issues/9217#issuecomment-292159323

mertsincan:
Hi,

Fixed in #1790

Best Regards,

PR #1608, opened 2020-10-02
Replace findDOMNode as it is deprecated in Strict Mode #1404

DeeHechEss:
Related to #1404 

I have looked through the code and replaced the use of findDOMNode with a normal ref or a forwarded ref from a DOM element where possible. This is just a suggestion, I'm happy to refactor any areas or for this to just be a starting point for a later refactor.

I was not able to fix DataScroller as the loader component as a ref is passed in as a prop and that may not always point to a DOM element, in case of the demo it points to another component. My suggestion is to pass the component to act as the button and render it within the DataScroller and directly apply the event listeners on it.

I also found the findDOMNode in the Galleria is not used but I left it that way as it seems that component is still in development.

I tested all the changes thoroughly but I've only recently started using primereact so I may not have caught everything.

react-transition-group has also added a fix to not use findDOMNode by passing it the ref of the child component using the nodeRef prop. I briefly tested this and it seems to require the DOM element ref not the component, this could be fixed by forwarding those component ref's as well. But it is probably easier to use a different animation method or library.

Hope this helps. Keep up the great work!


src/components/calendar/CalendarPanel.js
```diff
@@ -18,7 +18,7 @@ export class CalendarPanel extends Component {
 
     renderElement() {
         return (
-            <div ref={(el) => this.element = el} className={this.props.className} style={this.props.style}>
+            <div ref={(el) => {this.element = el; if(this.props.forwardRef) this.props.forwardRef(el)}} className={this.props.className} style={this.props.style}>
```
DeeHechEss, line 21: this.element is not used anywhere as far as I can tell, maybe it can be removed.

kashifshamaz21:
Can we have this PR merged? I think this is an important one considering it fixes the StrictMode warning which is happening in many components.

mertsincan:
Hi @DeeHechEss,

Best Regards,

kashifshamaz21:
Thanks a lot @DeeHechEss for providing the fix for this and thank you @mertsincan for merging the PR into master.

PR #1767, opened 2021-01-21
Made the Knob component controllable from external sources

### LINKED ISSUE #1766: Controlled Knob does not react to value changes on props
```
[ X ] bug report
[ ] feature request
```

https://codesandbox.io/s/primereact-test-forked-vpkxq

**Current behavior**
The Knob component will not update its value when the data passed in the ``value`` prop is changed. It always keep the initial ``value`` until an ``onChange`` event happens.

**Expected behavior**
The Knob should update whenever the ``value`` prop change.

1. Create a component with a button to increment a stated counter
2. Pass the value of the counter to the knob
3. See the problem happening

* **React version:** 16.8.0

* **PrimeReact version:** 6.0.0

* **Browser:** All
 
* **Language:** All

NathanPB:
### Defect Fixes
#1766 

This PR will remove the internal value of the Knob component that stores its value. Instead, it's expected to always receive the value from the props. This way, the component can be controlled from external sources and the problem described in the issue above does not happen.

This should not affect already existing Knobs if those were made according to the documentation (with value/setValue states)

### DIFF
--- a/src/components/knob/Knob.js
+++ b/src/components/knob/Knob.js
@@ -52,7 +52,6 @@ export class Knob extends Component {
 
         this.state = {}
 
-        this.value = this.props.value;
         this.radius = 40;
         this.midX =  50;
         this.midY = 50;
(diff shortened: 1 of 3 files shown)


NathanPB:
Force-pushes because I messed up the git history by mistake

mertsincan:

PR #1824, opened 2021-02-19
Add template to MenuItem interface

### LINKED ISSUE #1823: MenuItem interface missing template property
```
[x] bug report
[ ] feature request
```
**Current behavior**
There is no `template` property in [`MenuItem.d.ts`]
(https://github.com/primefaces/primereact/blob/master/src/components/menuitem/MenuItem.d.ts)

For example, if I use the [Template example from the site](https://www.primefaces.org/primereact/showcase/#/menumodel), it gives the error (in VSCode, for instance):
```
Type '{ label: string; template: (item: any, options: any) => Element; }' is not assignable to type 'MenuItem'.
  Object literal may only specify known properties, and 'template' does not exist in type 'MenuItem'.
```

* **PrimeReact version:**
6.0.1
 
* **Language:**
Typescript

bihanviranga:
Interface was missing this definition, making it difficult to use with Typescript.

###Defect Fixes
#1823

### DIFF
--- a/src/components/menuitem/MenuItem.d.ts
+++ b/src/components/menuitem/MenuItem.d.ts
@@ -9,4 +9,5 @@ export interface MenuItem {
     style?: any;
     className?: string;
     command?(e: {originalEvent: Event, item: MenuItem}): void;
+    template: JSX.Element;
 }


mertsincan:

Best Regards,

Issue #1848, opened 2021-03-05
Tooltip component has not defined showDelay, hideDelay prop types (Typescript).


Hoodrobinrs:
**I'm submitting a ...**
```
[x] bug report
[ ] feature request
[ ] support request => Please do not submit support request here, instead see https://forum.primefaces.org/viewforum.php?f=57
```

**Codesandbox Case**
https://codesandbox.io/s/pensive-ride-s0gg8?file=/src/demo/TooltipDemo.tsx

**Current behavior**
Currently the props showDelay and hideDelay are not defined in TooltipProps interface, they work when compiled to js, but in the typescript type definition they don't exist.

**Expected behavior**
The props showDelay and hideDelay should be defined.

**Minimal reproduction of the problem with instructions**
As shown in the codesandbox case:
https://codesandbox.io/s/pensive-ride-s0gg8?file=/src/demo/TooltipDemo.tsx
On line 18 the code throws an error, but the delay still works when we hover the button.

**Please tell us about your environment:**

Linux, VSCode, Yarn

* **React version:**
17.0.1

* **PrimeReact version:**
6.0.2

* **Browser:** [all | Chrome XX | Firefox XX | IE XX | Safari XX | Mobile Chrome XX | Android X.X Web Browser | iOS XX Safari | iOS XX UIWebView | iOS XX WKWebView ]
Brave, Chrome, Firefox
 
* **Language:** [all | TypeScript X.X | ES6/7 | ES5]
Typescript 4.1.3

PR #1849, opened 2021-03-05
Added delay prop types to Tooltip component

Hoodrobinrs:
###Defect Fixes
 [#1848 ](https://github.com/primefaces/primereact/issues/1848)


Hoodrobinrs:
Seems like i was late to the party, as it is already implemented in master. Sorry for the trouble.

PR #1915, opened 2021-03-29
Improving typing of ToolTip position

master117:
I have added all valid values of tooltips position argument as typing.


mertsincan:

PR #1952, opened 2021-04-12
Update Calendar.js

paulorievrs:
If an language uses the dayNamesMin with single characters it causes an error of key list. Fixed using the index insted of the value.

###Defect Fixes

###Feature Requests

### DIFF
--- a/src/components/calendar/Calendar.js
+++ b/src/components/calendar/Calendar.js
@@ -2592,9 +2592,9 @@ export class Calendar extends Component {
     }
 
     renderDayNames(weekDays) {
-        const dayNames = weekDays.map(weekDay =>
+        const dayNames = weekDays.map( (weekDay, index) =>
         (
-            <th key={weekDay} scope="col">
+            <th key={index} scope="col">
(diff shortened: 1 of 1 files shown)


src/components/calendar/Calendar.js
```diff
@@ -2592,9 +2592,9 @@ export class Calendar extends Component {
     }
 
     renderDayNames(weekDays) {
-        const dayNames = weekDays.map(weekDay =>
+        const dayNames = weekDays.map( (weekDay, index) =>
         (
-            <th key={weekDay} scope="col">
+            <th key={index} scope="col">
```
master117, line 2597: Using index for Keys is an Antipattern as it can lead to errors when weekdays change, as the indexes won't change. https://robinpokorny.medium.com/index-as-a-key-is-an-anti-pattern-e0349aece318

```suggestion
            <th key={`${weekDay}-${index}`} scope="col">
```

master117 (changes requested):

mertsincan:

Best Regards,

PR #2049, opened 2021-05-19
CalendarProps  value type  undefined et null

TrabelsiHoussem:
i propose to change the value type of the calendar Props interface  to  avoid this Error 
verload 1 of 2, '(props: CalendarProps | Readonly<CalendarProps>): Calendar', gave the following error.
    Type 'Date | Date[] | null | undefined' is not assignable to type 'Date | Date[] | undefined'.
  Overload 2 of 2, '(props: CalendarProps, context: any): Calendar', gave the following error.
    Type 'Date | Date[] | null | undefined' is not assignable to type 'Date | Date[] | undefined'

###Defect Fixes

###Feature Requests

### DIFF
--- a/src/components/calendar/Calendar.d.ts
+++ b/src/components/calendar/Calendar.d.ts
@@ -40,7 +40,7 @@ export interface CalendarProps {
     id?: string;
     inputRef?: React.Ref<HTMLInputElement>;
     name?: string;
-    value?: Date | Date[];
+    value?: Date | Date[]|undefined | null;;
     viewDate?: Date;
     style?: object;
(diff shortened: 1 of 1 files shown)


master117:
Value is ?, so you can (and should) already use undefined, so you don't need to union with undefined. 

Adding null does not provide a benefit, Furthermore PrimeReact code is not guaranteed to handle null values there, it does in no other place either. But it looks like it would work. 

A space is missing before the first pipe.

You should probably remove null from your data structure. Having both null and undefined as values is rarely a good idea.

PR #2068, opened 2021-05-28
fixed condition used to determine size of splitter container

### LINKED ISSUE #2067: Splitter considers the height as the size of the container irrespective of the layout
```
[x] bug report
[ ] feature request
```
**Current behavior**
The size used to determine the size of a splitter panel is always the height of the container irrespective of the layout.

**Expected behavior**
The size should be width of the container for a horizontal layout and height for vertical layout.

anudeepreddy:
###Defect Fixes
Fixes #2067 

There is a small bug in the Splitter code. The splitter always computes the size of the panel based on the height of the container due to the use of a undefined variable for conditional check. 

This PR adds the proper variable for conditional check and fixes the issue.

### DIFF
--- a/src/components/splitter/Splitter.js
+++ b/src/components/splitter/Splitter.js
@@ -135,7 +135,7 @@ export class Splitter extends Component {
 
     onResizeStart(event, index) {
         this.gutterElement = event.currentTarget;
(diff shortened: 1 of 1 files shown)


mertsincan:
Good catch! Thanks a lot for your contribution! 

Best Regards,

PR #2172, opened 2021-07-07
fix name types to slidemenu

vpsgvitor:
this bug there is a problem in the slidemenu type, the import is breaking the component's usage


mertsincan:
Thanks a lot!!!!!

PR #2219, opened 2021-07-27
Fix File Upload Deleting Emails

CalebDepatie:
###Defect Fixes
[See here for issue submitted to the forum last year.](https://forum.primefaces.org/viewtopic.php?t=64337)
When dragging and dropping an email into the file upload component it sends the email to the deleted folder.
The fix is simple and detailed in the same forum post, I've just implemented it and submitted a PR.


mertsincan:
Thanks a lot for the contribution!

Best Regards,

PR #2335, opened 2021-10-03
Improve custom icon support

amymartyn:
Buttons already accept custom icons, but most other components do not. This changes allows all components that have icon props to use custom icons, similar to Buttons.

The main changes are 

- Adding a generic icon type in typescript to avoid repetitive code
- Adding an `IconUtils` class that has a method `getJSXIcon`. This method is based on `renderIcon` in the `Button` class. For strings, it will create a `span` with the string as the `className`. For `React.ReactNode` objects or functions that return a `React.ReactNode` object, it will wrap the object in a span and return it
- Updating components to use `getJSXIcon` instead of creating `span` or `i` icons directly

I don't know if this will be merged, but I thought it might useful to other people who want to use custom icons


amymartyn:
@mertsincan Any thoughts on this?

mertsincan:
Hi @amymartyn,

Nice PR! Can you update the documentation and codes according to conflicts?

Best Regards,

amymartyn:
@mertsincan OK, conflicts have been fixed

mertsincan:
Thanks a lot for your effort, @amymartyn ;) Good idea, Nice PR ;)

PR #2356, opened 2021-10-19
Fix #2354: Datatable CRUD Demo correction

### LINKED ISSUE #2354: Problem with DataTable Crud demo (Hooks source)
There is a problem with 
https://primefaces.org/primereact/showcase/#/datatable/crud
Hooks source demo

in deleteProduct method

`const deleteProduct = () => {
        ...
        setProduct(_products);
       ...
    }`
should be
`const deleteProduct = () => {
        ...
        setProducts(_products);
       ...
    }` (setProduct**s**)

That is why if you try to delete item using yellow button, it won't work in the current version: https://codesandbox.io/s/i01ys

melloware:
###Defect Fixes
Fixed #2354 the crud demo was calling `setProduct` instead of `setProducts`

### DIFF
--- a/src/showcase/datatable/DataTableCrudDemo.js
+++ b/src/showcase/datatable/DataTableCrudDemo.js
@@ -891,7 +891,7 @@ const DataTableCrudDemo = () => {
 
     const deleteProduct = () => {
         let _products = products.filter(val => val.id !== product.id);
-        setProduct(_products);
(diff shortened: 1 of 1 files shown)


mertsincan:
This may have changed., therefore closed. We're working to update all DataTable demo.

Best Regards,

PR #2379, opened 2021-11-01
Fix #2363: Fileupload do not allow multiple files if multiple=false

### LINKED ISSUE #2363: Bug - FileUpload allows multiple even if "multiple" prop is not set
```
[x] bug report
[ ] feature request
```

Forked for reporting: https://codesandbox.io/s/primereact-test-forked-gb3ym?file=/src/index.js

**Current behavior**

Using the default example for an Advanced FileUpload component (ie: do not provide `multiple` or even provide `multiple={false}` it will still accept multiple files.

Code snippet context:
When using drag and drop, on this line it is checking if there IS exactly one other file, it should disallow if there is more than 0 files selected
https://github.com/primefaces/primereact/blob/7a095558d78eadc2417f4c15d304127d789ead5e/src/components/fileupload/FileUpload.js#L428

When using select there is no check for multiple within this function, it just pushes more files and should be checking if multiple is set (or ideally a max number you can pass in)
https://github.com/primefaces/primereact/blob/7a095558d78eadc2417f4c15d304127d789ead5e/src/components/fileupload/FileUpload.js#L213

Screenshot examples

From codesandbox example
image

From showcase
image

**Expected behavior**

It should only allow the first file to be selected. (You may wish to allow users to cancel and preview so Advanced is still preferred here)

see demo.

1. Remove the `multiple` prop from the Advanced FileUpload in the showcase
2. Note you can select OR drag multiple files still

MacOS 11
VS Code

* **React version:**

Whatever is used on the showcase site

* **PrimeReact version:**

Both 6.5.1 and 6.6.0

Firefox
 
all

melloware:
###Defect Fixes
Fix #2363: Fileupload do not allow multiple files if multiple=false

### DIFF
--- a/src/components/fileupload/FileUpload.js
+++ b/src/components/fileupload/FileUpload.js
@@ -425,7 +425,7 @@ export class FileUpload extends Component {
             event.preventDefault();
 
             let files = event.dataTransfer ? event.dataTransfer.files : event.target.files;
-            let allowDrop = this.props.multiple || (files && files.length === 1);
+            let allowDrop = this.props.multiple || (files && files.length === 0);
 
             if (allowDrop) {
                 this.onFileSelect(event);


mertsincan:
Thanks a lot, Mello ;)

PR #2394, opened 2021-11-05
fix: don't always add p-splitter-panel-nested class

stambata:
p-splitter-panel-nested class is always added no matter whether the panel element contains children with 'p-splitter' or not.
[DomHandler.find](https://github.com/primefaces/primereact/blob/e9137c63eaa44a3e1037c5775b0645a4a5e2a2b6/src/components/utils/DomHandler.js#L194) always returns an array.


mertsincan:
Good catch! Thanks a lot for your contribution!

Best Regards,

Issue #2406, opened 2021-11-08
Tooltip positioned incorrectly when target is an SVG


raymondcao-everlaw:
**I'm submitting a ...**  (check one with "x")
```
[x ] bug report
[ ] feature request
[ ] support request => Please do not submit support request here, instead see https://forum.primefaces.org/viewforum.php?f=57
```

**Codesandbox Case (Bug Reports)**

The first SVG circle in the sandbox below shows the bug. The second shows expected behavior (this is achieved by wrapping the SVG in a div as a workaround).

https://codesandbox.io/s/primereact-test-forked-97oko?file=/src/index.js

**Current behavior**
If the target of a global tooltip is an SVG, the tooltip will not be positioned next to the target when triggered (it will often appear offscreen).

**Expected behavior**
If the target of a global tooltip is an SVG, the tooltip will should be positioned next to the target.

**Minimal reproduction of the problem with instructions**

See sandbox link.

```jsx
<>
    <Tooltip target=".svgtarget" autoHide={false} focusable={false}>
        Test Tooltip
    </Tooltip>
    <svg width="100" height="100" className="svgtarget">
        <circle cx="50" cy="50" r="40" stroke="green" strokeWidth="4" fill="yellow" />
    </svg>
</>
```

**Please tell us about your environment:**

* **React version:** 17.0.1

* **PrimeReact version:** 7.0.0

* **Browser:** Chrome 95.0.4638.69

 
* **Language:** Typescript, Javascript

PR #2407, opened 2021-11-08
Fix Tooltip positioning issue with SVG targets

raymondcao-everlaw:
### Defect Fixes
Addresses https://github.com/primefaces/primereact/issues/2406.

The issue was that if the target of a global tooltip was an SVG, the tooltip would not be positioned next to the target when triggered (it will often appear offscreen). The source of the problem was that `DomHandler.getOuterWidth` and `DomHandler.getOuterHeight` rely on `Element.offsetWidth` to get the target element's width. This works for HTMLElement instances but not SVGElement instances.

My PR uses `Element.getBoundingClientRect` to get the target element's width when `Element.offsetWidth` is undefined.

### DIFF
--- a/src/components/utils/DomHandler.js
+++ b/src/components/utils/DomHandler.js
@@ -34,7 +34,7 @@ export default class DomHandler {
 
     static getOuterWidth(el, margin) {
         if (el) {
-            let width = el.offsetWidth;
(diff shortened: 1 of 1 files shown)


melloware (approved):

mertsincan:

Best Regards,

PR #2442, opened 2021-11-17
Fix #2423: Add CSP NONCE to inline style elements if found

### LINKED ISSUE #2423: CSP: Violating Inline Style Policy
[x ] bug report

**Current behavior**
Using Create React App and enabling CSP protection of the app some components in PrimeReact violate the inline style policy. This is due to use of `document.head.appendChild` in things like Datatable to make it responsive.

**Policy:**
```javascript
const cspConfigPolicy = {
    'default-src': "'self'",
    'connect-src': "'self' localhost:* 127.0.0.1",
    'base-uri': "'self'",
    'object-src': "'none'",
    'script-src': ["'self'  https:"],
    'style-src': ["'self'  https:"]
};
```

**Error:**
image

The offending code for example can be found here when creating Responsive Styles: https://github.com/primefaces/primereact/blob/77b1ce46af7cc8ace370f47f7b0df332e8e34178/src/components/datatable/DataTable.js#L953-L995

**Expected behavior**
PrimeReact can run with `'style-src': self` CSP Policy.

Follow these instructions to add CSP to a Create React App: https://medium.com/@nrshahri/csp-cra-324dd83fe5ff

Then run the showcase and hit the Datatable page you will see the issue

* **React version:** 17.0.2
* **PrimeReact version:** 7.0.0
* **Browser:** [all ]
* **Language:** [all]

melloware:
###Defect Fixes
Fix #2423: Add CSP NONCE to inline style elements if found

This is a very important fix for security.   This checks if the downstream application using PrimeReact has set a CSP NONCE environment variable `process.env.REACT_APP_CSS_NONCE`.  This can be hardcoded in their `.env` file or if they are using a dynamic NONCE generator like CRA CSP WebPack Plugin: https://github.com/slackhq/csp-html-webpack-plugin.

If you do dynamically generate your NONCE you can easily set it in your application like this for PrimeReact to pick it up and use it in all its inline styling efforts.

```typescript
// attempt to read an existing NONCE from a  CSS link on the page
let nonce;
let cssLink = document.querySelector('link[nonce]') as HTMLScriptElement;
if (cssLink) {
    nonce = cssLink?.nonce!;
}

// now set the environment variable for PrimeReact to read
if (nonce) {
    process.env.REACT_APP_CSS_NONCE= nonce;
}
```

### DIFF
--- a/src/components/carousel/Carousel.js
+++ b/src/components/carousel/Carousel.js
@@ -371,8 +371,7 @@ export class Carousel extends Component {
 
     createStyle() {
         if (!this.carouselStyle) {
-            this.carouselStyle = document.createElement('style');
-            document.body.appendChild(this.carouselStyle);
+            this.carouselStyle = DomHandler.createInlineStyle();
         }
 
         let innerHTML = `
--- a/src/components/datatable/DataTable.js
+++ b/src/components/datatable/DataTable.js
@@ -951,14 +951,12 @@ export class DataTable extends Component {
     }
 
     createStyleElement() {
-        this.styleElement = document.createElement('style');
-        document.head.appendChild(this.styleElement);
+        this.styleElement = DomHandler.createInlineStyle();
     }
 
     createResponsiveStyle() {
         if (!this.responsiveStyleElement) {
-            this.responsiveStyleElement = document.createElement('style');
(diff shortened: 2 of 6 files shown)


mertsincan:
Thanks a lot!

PR #2468, opened 2021-11-25
fix: avoid DOMException

kalinkrustev:
The below exception is thrown during react fast refresh:
DOMException: Failed to execute 'removeChild' on 'Node': The node to be removed is not a child of this node.


melloware:
Please create an issue so we can link this PR to the issue for tracking.

melloware:
Superseded by: https://github.com/primefaces/primereact/pull/2471

PR #2472, opened 2021-11-25
Fix #2465: InputNumber default to min on allowEmpty=false

### LINKED ISSUE #2465: InputNumber Critical Issue

```
[X] bug report
[ ] feature request
```

https://codesandbox.io/s/primereact-test-forked-i946f

**Current behavior**
When uses InputNumber property "min"

**Expected behavior**
if Input a letter this letter will never appear, and when the component delete this letter respect the min property

create a <InputNumber with min property and put a letter on the input, then the input will delete de letter but will delete everything and the value become 0, ignoring the min property

Windows 10
node 16.13.0
react 17.0.2
primeReact 7.0.1

* **React version:**
react 17.0.2

* **PrimeReact version:**
primeReact 7.0.1

* **Browser:** 
* [ Chrome XX  ]
 
* **Language:** [ES6/7]

melloware:
###Defect Fixes
Fix #2465: InputNumber default to min on allowEmpty=false

Currently its always defaulting to 0 if allowEmpty=false and it should respect the minimum like `min={2}` and reset it to `2`.  If however min < 0 like `min={-100}` we should leave it at 0 and not make the default value `-100`

### DIFF
--- a/src/components/inputnumber/InputNumber.js
+++ b/src/components/inputnumber/InputNumber.js
@@ -783,11 +783,19 @@ export class InputNumber extends Component {
 
         if (valueStr != null) {
             newValue = this.parseValue(valueStr);
-            newValue = !newValue && !this.props.allowEmpty ? 0 : newValue;
+            newValue = this.evaluateEmpty(newValue);
             this.updateInput(newValue, insertedValueStr, operation, valueStr);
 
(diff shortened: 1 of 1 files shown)


IHPNULL (approved):
Great, this will make it work fine

PR #2488, opened 2021-12-02
Fix #2467: Typescript use enums for options

### LINKED ISSUE #2467: Typescript DataTables with filters don't import or compile correctly.
```
[x] bug report
[ ] feature request
```

https://codesandbox.io/s/ykcro
This is the same codesandbox linked from the TS documentation on DataTable's filter page.

**Current behavior**
FilterMatchMode imports as an interface, not an object. The objects/types used on the DataTable component down the file are also incorrect. This causes compilation to break unless errors are ignored.

**Expected behavior**
FilterMatchMode should import the actual object instead of the interface. DataTables should be using the correct interfaces when passing props. Compilation should work without ignoring all errors.

Import FilterMatchMode and try to use it by calling FilterMatchMode.CONTAINS.

Azure VM, VSCode Community 2019, Nextjs 12.0.1, typescript 4.4.4, eslint 7.32.0

* **React version:**
17.0.2

* **PrimeReact version:**
Tried in both 7.0.0-rc2 and 7.0.1.

Chrome
 
Typescript 4.4.4

melloware:
###Defect Fixes
Fix #2467: Typescript use enums for options

Using Typescript `enum` is the more correct way to represent these options

### DIFF
--- a/src/components/api/Api.d.ts
+++ b/src/components/api/Api.d.ts
@@ -283,44 +283,38 @@ export interface PrimeIconsOptions {
 export declare const PrimeIcons: PrimeIconsOptions;
 
 // Severity
-export interface MessageSeverityOptions {
-    readonly SUCCESS: string;
-    readonly INFO: string;
-    readonly WARN: string;
-    readonly ERROR: string;
+export declare enum MessageSeverity {
+    SUCCESS = 'success',
+    INFO = 'info',
+    WARN = 'warn',
+    ERROR = 'error'
 }
-
-export declare const MessageSeverity: MessageSeverityOptions;
 
 // Filter
-export interface FilterMatchModeOptions {
-    readonly STARTS_WITH: string;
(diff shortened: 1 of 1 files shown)


jafin:
Would you consider preferring union types instead of enums? 
https://fettblog.eu/tidy-typescript-avoid-enums/

melloware:
@jafin this will probably be Primetek's preferred approach to use union types as that is how all of the rest of the code base is.  However it was curious they chose to go this Interface route with static read-only when from reading their examples it LOOKS like they wanted an enum for `FilterMatchMode.IN` etc.

Thoughts?

jafin:
@melloware if existing code follows a convention, I assume that would be the approach.

mertsincan:
Thank a lot, @jafin and @melloware! For now, the 'enum' definition seems appropriate for this issue. I'll talk to our team for improvements.

PR #2500, opened 2021-12-07
Fix #2496: MultiSelect null check

### LINKED ISSUE #2496: MultiSelect - breaking when options is empty array
```
[x] bug report
[ ] feature request
```

**Current behavior**
I have a MultiSelect component being generated depending on a switch case. The options values are being generated programmatically in useEffect. 
However, the Multiselect is breaking for the generated options. It works fine if I provide a hardcoded value. The same generated values are working when given to a Dropdown component.

Error message: TypeError: Cannot read properties of null (reading 'length')
    at MultiSelect.isAllSelected

**Expected behavior**
Expecting Multiselect to *not* break for programmatically generated options set.

Help...

* **React version:**
17.0.2

* **PrimeReact version:**
6.5.1

melloware:
###Defect Fixes
Fix #2496: MultiSelect null check

### DIFF
--- a/src/components/multiselect/MultiSelect.js
+++ b/src/components/multiselect/MultiSelect.js
@@ -657,7 +657,7 @@ export class MultiSelect extends Component {
         }
         else {
             let visibleOptions = this.getVisibleOptions();
-            if (visibleOptions.length === 0) {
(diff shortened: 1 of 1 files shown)


mertsincan:
Thanks a lot @melloware ;)

PR #2525, opened 2021-12-15
Fix #2423/#2536: CSP allow setting inline nonce with PrimeReact.inlineCssNonce

### LINKED ISSUE #2423: CSP: Violating Inline Style Policy

melloware:
###Defect Fixes
Fix #2423: CSP allow setting inline nonce with PrimeReact.inlineCssNonce

Setting an `env` at build time is not always convenient and may not be possible. This allows us to set the NONCE at runtime.

### DIFF
--- a/components/lib/api/Api.d.ts
+++ b/components/lib/api/Api.d.ts
@@ -20,6 +20,7 @@ interface FilterMatchModeOptions {
 interface APIOptions {
     ripple?: boolean;
     inputStyle?: InputStyleType;
+    inlineCssNonce?: string;
     locale?: string;
     appendTo?: AppendToType;
     cssTransition?: boolean;
--- a/components/lib/api/PrimeReact.js
+++ b/components/lib/api/PrimeReact.js
@@ -14,6 +14,8 @@ export default class PrimeReact {
 
     static autoZIndex = true;
 
+    static inlineCssNonce = null;
+
     static zIndex = {
         modal: 1100,
         overlay: 1000,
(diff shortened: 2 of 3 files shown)


melloware:
Also fixes : #2536

mertsincan:
Thanks a lot, @melloware! Maybe, we can use only PrimeReact.nonce feature.

melloware:
I like having the build time option too because I automatically add it with my CSP WebPack plugin: https://github.com/melloware/csp-webpack-plugin#about

I like the flexibility and thanks for accepting!

PR #2535, opened 2021-12-17
fix: TabMenu headers bottom border does not align

ziggystar:
In TabMenu, when content begins to line-break, headers have different height and the bottom border does not align anymore. I've set `height: 100%` for the offending element. This works for me; maybe there's a better fix.

image


melloware:
Do you mind opening an issue so PrimeTek can track this PR against and issue.  That is their typical policy.

ziggystar:
fixes #2540

mertsincan:

Best Regards,

PR #2567, opened 2022-01-09
Fix #2566: Sanitize innerHTML for Trusted Types

### LINKED ISSUE #2566: CSP: Support Trusted Types
Trusted Types Doc: https://developer.mozilla.org/en-US/docs/Web/API/Trusted_Types_API

Support Trusted Types in PrimeReact.  With this CSP profile...

```xml
<meta http-equiv="Content-Security-Policy" content="require-trusted-types-for 'script'" />
```

This prevents the use of `innerHTML` unless its been sanitized.   For example in Primereact you will get this error on a icon only button...

image

... because it generates this span...

```xml
<span className="p-button-label p-c" dangerouslySetInnerHTML={{ __html: "&nbsp;" }}></span>
```

You then have to use a TrustedType to set any innerHtml like...

```javascript
if (window.trustedTypes && trustedTypes.createPolicy) { // Feature testing
  trustedTypes.createPolicy('primereact', {
    createHTML: string => DOMPurify.sanitize(string, {RETURN_TRUSTED_TYPE: true})
  });
}
```

React offers a by-design way to assign data to an element: the `dangerouslySetInnerHTML` property. Data assigned to this property is forwarded to the DOM element's `innerHTML` property. This React property is marked as dangerous because there is no sanitization applied. Developers are responsible for handling this themselves, and a failure to do so creates DOM-based XSS vulnerabilities.

Trusted Types effectively prevents insecure usage of `dangerouslySetInnerHTML`. The only way to assign data as HTML with Trusted Types enabled is by ensuring the data is a trusted type. The code example below illustrates how to use DOMPurify to sanitize data and return it as a trusted type.

```xml
<div dangerouslySetInnerHTML={{__html: DOMPurify.sanitize(data, {RETURN_TRUSTED_TYPE: true})}} />
```

DOMPurify sanitizes the data, effectively making it safe to render as HTML. Since the data is now marked as a Trusted Type, the browser will allow this data to end up in the DOM. And just like before, this is applied on the browser level, so it covers your application code and all its dependencies and components.

**NOTE:** DomPurify not only handle sanitizing but it handles browser backward compatibility of Trusted Types.

melloware:
###Defect Fixes
Fix #2566: Sanitize innerHTML for Trusted Types


package.json
```diff
@@ -9,23 +9,24 @@
...
+    "final-form": "^4.20.2",
+    "formik": "^2.2.6",
+    "isomorphic-dompurify": "^0.17.0",
+    "jspdf": "^2.4.0",
+    "jspdf-autotable": "^3.2.5",
     "next": "12.0.4",
+    "primeflex": "2.0.0",
+    "primeicons": "^5.0.0",
+    "quill": "1.3.7",
     "react": "17.0.2",
```
melloware, line 23: When I did NPM INSTALL of isomorphic-dompurify it sorted the deps alphabetically.

components/lib/utils/DomHandler.js
```diff
@@ -1,3 +1,5 @@
+import DOMPurify from 'isomorphic-dompurify';
```
melloware, line 1: I used isomorphic-dompurify because it works serverside or client side so it works for Next SSR as well as client side.

melloware:
Also not sure if I need to add to `rollup.config.js`

```javascript
// dependencies
const GLOBAL_DEPENDENCIES = {
    'react': 'React',
    'react-dom': 'ReactDOM',
    'react-transition-group': 'ReactTransitionGroup'
};
```

PR #2610, opened 2022-01-25
Fix #2604: Sort based on browser locale setting

### LINKED ISSUE #2604: DataTable sorting lacks collation support
It seems that DataTable sorting is fixed to just one collation. Different languages use different sorting for special characters. 

E.g. in German A and Ä are considered the same letter while in Swedish and Finnish they are different. So in German sort would produce
A
Ä
B
C
...

and in SE/FI this would be
A
B
C
...
Z
Ä
Å
Ö

See https://unicode.org/reports/tr10/ (Introduction).

melloware:
###Defect Fixes
Fix #2604: Sort based on browser locale setting

Sort by the browser locale settings and fallback to `en` if not detected.


melloware:
@Makistos please review.

Makistos:
How do you determine the locale? I went through Firefox's about:config and couldn't find any setting there. In Chrome I seem to have English as language and I think this is not uncommon among us non-English speakers as well.

That's why an API to set this manually would also be nice. This way it would also be possible to have user select it.

melloware:
Open up F12 Console in your browser and type...

```
navigator.userLanguage
          || (navigator.languages && navigator.languages.length && navigator.languages[0])
          || navigator.language 
          || navigator.browserLanguage 
          || navigator.systemLanguage 
          || 'en'
```

That should return you the current browser language in my case `en` in the United States.

image

melloware:
But a user really already HAS selected it by setting their browser's locale or their OS setting the browser locale no?

melloware:
Hmm datatable has `filterLocale` property it would be weird to have a `filterLocale` and separate `sortLocale` maybe that property should just be `locale`.

Makistos:
Mine says en-US as well. So even though I'm a Finn that's my Firefox locale. My OS language is English but locale is Finnish so dates etc. look correct. Firefox I downloaded from their website so it's the default English version. 

And as I said I think this is quite common. It's just more convenient to use English version of software as not everything has a Finnish version anyways and/or we like the  English terms better when it comes to computer speak.

Makistos:
> Hmm datatable has `filterLocale` property it would be weird to have a `filterLocale` and separate `sortLocale` maybe that property should just be `locale`.

This would make sense to me. I would think one locale at a time is enough. I can't think we someone would want to use different logic for sorting and filtering.

melloware:
Just updated my PR there is `PrimeReact.locale` global variable.

So in your code you could do `PrimeReact.locale = 'fi`;` and it would set it for the users whole session.

Makistos:
Sounds like a good change to me, now it looks like it works as before if you don't set that value but if you do it overrides everything else.  Minimal impact, maximum customizability.

melloware:
agreed!

melloware:
@mertsincan I can also live with just checking `PrimeReact.locale` and removing the navigator language backup checks?

mertsincan:
@melloware, yes the 'getLanguage' can only remain as a method. Users can use it whenever they want and set the value of this method to PrimeReact.locale.

Thanks a lot for this PR!

PR #2611, opened 2022-01-26
Fix #2609: Associate editingMeta with row key

### LINKED ISSUE #2609: DataTable editingMeta is bound to row index, instead of row key
**I'm submitting a ...**  
```
[ x ] bug report
```

https://codesandbox.io/s/sharp-euler-by70t?file=/src/demo/DataTableCrudDemo.js

**Current behavior**
When rows are deleted and editMode='row', previously unsaved edits of rows after the deleted are shown on wrong rows.

**Expected behavior**
The values in editingMeta should be associated with row key, not row index, so that they are shown at the proper rows after deletion, etc. 
There are two additional issues, which might be related:
- unsaved data is shown on the newly edited row
- deletion of the row causes the rows after it to exit edit mode

1. Open the sandbox URL.
2. Click the pencil (column 2) on rows 1 and 3 to start editing.
3. Select row 2 and delete it with the delete button at the top (confirm the deletion).
4. Click the pencil on row 3 and you will see a wrong value. It equals the value from previous row 3, which is now row 2. If the value was edited after step 2, the edited value will be shown, despite not being confirmed.

* **PrimeReact version:**
7.2.0

kalinkrustev:


melloware (approved):

melloware:
Can you change the title of your two tickets and add `"Fix #xxx" to the beginning.  It helps primetek and also in the future it automatically links the PR to that ticket number.  nice work!

mertsincan:

Best Regards,

Issue #2344, opened 2021-10-12
InputNumber doesn't submit a form


leqwasd:
**I'm submitting a ...** 
```
[ x ] bug report
[ ] feature request
[ ] support request => Please do not submit support request here, instead see https://forum.primefaces.org/viewforum.php?f=57
```

**Codesandbox Case (Bug Reports)**

https://codesandbox.io/s/primereact-issue-inputnumber-submit-5rgpz?file=/src/index.tsx

**Current behavior**
InputNumber pressing Enter doesn't submit a form it is in. InputText does submit it. So does native inputs.

**Expected behavior**
Pressing Enter within focused InputNumber component should submit a form.

* **React version:** 16

* **PrimeReact version:** 6.6.0

* **Browser:** [Chrome XX]
 
* **Language:** [all | TypeScript X.X | ES6/7 | ES5]

PR #2637, opened 2022-02-11
fix: prevent InputNumber from press literal keys

### LINKED ISSUE #2631: InputNumber editor in Datatable doesn't accept only numbers as input values anymore

```
[x] bug report
[ ] feature request
```

https://codesandbox.io/s/92k7m

**Current behavior**
Using an InputNumber component as an editor of a table cell doesn't work correctly anymore. Starting with version 7.2. it is no longer possible that input of characters other than numbers are corrected by the input field. Instead the characters for example an 'a' or any other still remains in the field and only is deleted when a number was put in or the editor is closed.

**Expected behavior**
As in the versions before it shouldn't be possible to input any other characters than values. The field should delete wrong input characters while editing a field.

I simply link the edit example of the Datatable from the showcase for version 7.2 to this issue. When one tries to input any characters to the fields quantity or price of a table cell in the first example or any other it can be seen easily.

* **React version:**
17.0.2

* **PrimeReact version:**
7.2.0

all

walt1992:
Fixes #2631
and Keeps requirement from #2344

### DIFF
--- a/components/lib/inputnumber/InputNumber.js
+++ b/components/lib/inputnumber/InputNumber.js
@@ -534,14 +534,15 @@ export class InputNumber extends Component {
         }
 
         let code = event.which || event.keyCode;
+
+        if (code !== 13) event.preventDefault();
+
         let char = String.fromCharCode(code);
         const isDecimalSign = this.isDecimalSign(char);
         const isMinusSign = this.isMinusSign(char);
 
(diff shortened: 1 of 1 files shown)


mertsincan:
Nice PR! Thanks a lot for your contribution!

Best Regards,

PR #2648, opened 2022-02-20
Jalali calendar

seyed-amir-mehrizi:
###Defect Fixes

###Feature Requests


melloware:
@seyed-amir-mehrizi I don't think this PR will be accepted because it introduces `moment.js` and PrimeTek will not want to add this dependency.

Also `moment.js` is end of life and no longer being supported see: https://momentjs.com/docs/#/-project-status/

seyed-amir-mehrizi:
Dear Primefaces/Primereact ,
Thanks for your reply. Because we are using Jalali Calendar
<https://en.wikipedia.org/wiki/Jalali_calendar> in Iran, because of that,
We must use   moment.js  as a library to handle the date in our Apps.
Best regards,
Seyed Amir Mehrizi
Tehran , Iran

mertsincan:
Hi,
I agree with @melloware.
But, I'll think about how we can solve this without any dependencies. Maybe we can export a method for some things to be edited by the user.
@seyed-amir-mehrizi Could you please open an issue about it?

Best Regards,

Issue #2668, opened 2022-03-02
Toast - Clear and then show doesn't clear existing messages


omerts:
**I'm submitting a ...**  
```
[ X] bug report
```

**Codesandbox Case (Bug Reports)**
https://codesandbox.io/embed/primereact-test-forked-kf580b?fontsize=14&hidenavigation=1&theme=dark]

**Current behavior**

Use Toast's clear method and then show, doesn't clear existing messages, as setState is an async operation, so show is using a stale messages state, and thus returns old messages. 

**Expected behavior**
Messages should be cleared up clear regardless of show being used afterwards. 

**Minimal reproduction of the problem with instructions**
See codesandbox demo

* **PrimeReact version:**
6.0.1

* **Language:** all

melloware:
Also I know Messages has a replace function which maybe you can add which will call clear and add the new message in one method call as a bonus?

melloware:
This is actually fixed in 8.0.0-rc1 now that it has been switched to Hooks.
See: https://codesandbox.io/s/primereact-test-forked-5zuel3

melloware:
OK I have added a `replace` method which also accomplishes this in 1 line without calling `clear` then `show`

PR #2669, opened 2022-03-02
Update toast to always have up to date messages state

omerts:
Fixes bug https://github.com/primefaces/primereact/issues/2668

### DIFF
--- a/components/lib/toast/Toast.js
+++ b/components/lib/toast/Toast.js
@@ -83,6 +83,13 @@ export class Toast extends Component {
         this.setMessages([])
     }
 
+    replace(value) {
+        // We are fine with setMessages being called twice behind the scenes
+        // as react batches state updates
+        this.clear()
+        this.show(value)
+    }
+
     onClose(message) {
         let newMessages = this.messages.filter(msg => msg.id !== message.id);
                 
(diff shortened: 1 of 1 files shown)


melloware (approved):
Looks good to me just wasnt' sure if PrimeTek had it in the `state` for a reason.  But your fix looks good to me.

omerts:
@melloware I presume that he had it on state in order to cause renders when this state changes, so instead I used forceUpdate for rendering on changes.

melloware:
Excellent @omerts   Do you see my suggestion about adding a `replace` method as well like Mssages has so you can call it and it will call clear() and then show() for the new message?

```ts
public replace(message: ToastMessageType): void;
```

omerts:
@melloware done :)

melloware:
Can you also update the .d.tams file to list this new replace method.  Thanks!

melloware (changes requested):
See my comment about the TypeScript definition update.

melloware (approved):

melloware:
There are merge conflicts to fix.

melloware:
@omerts do you mind updating your fixes?

melloware:
This is fixed now that hooks are being used: https://codesandbox.io/s/primereact-test-forked-5zuel3

omerts:
@melloware  Sorry for not getting back to this. Happy to know it is fixed, you are awesome!

PR #2697, opened 2022-03-23
Fix #2696: Toast allow immutable messages

### LINKED ISSUE #2696: Toast component throws error on show

```
[x] bug report
[ ] feature request
```

https://codesandbox.io/s/primereact-test-forked-z3nsj9

**Current behavior**
Toast component throws an error when calling show method with immutable object.

image

**Expected behavior**
Expected to see toast message

I run into this problem passing redux (redux-toolkit) state as parameter to show method.
I used object.freeze method to generate immutable message object.
Click button then you can see error message.

* **React version:**
17.0.2

* **PrimeReact version:**
7.2.1

* **Browser:** [all]

 
* **Language:** [all]

melloware:
###Defect Fixes
Fix #2696: Toast allow immutable messages

Must make a copy of the message before trying to append an ID.

### DIFF
--- a/components/lib/toast/Toast.js
+++ b/components/lib/toast/Toast.js
@@ -53,24 +53,24 @@ export class Toast extends Component {
 
     show(value) {
         if (value) {
-            if (value) {
-                let newMessages;
+            // GitHub #2696 copy value
+            const currentMessage = {...value}
+            let newMessages;
 
-                if (Array.isArray(value)) {
-                    for (let i = 0; i < value.length; i++) {
-                        value[i].id = messageIdx++;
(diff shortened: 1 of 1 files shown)


components/lib/toast/Toast.js
```diff
@@ -53,24 +53,24 @@ export class Toast extends Component {
 
     show(value) {
         if (value) {
-            if (value) {
```
melloware, line 56: For some reason it had the above IF twice in a row...

mertsincan:
Thanks a lot @melloware ;)

PR #2698, opened 2022-03-28
Add Dialog and ConfirmDialog component classes typings

sshulin:
Case:
I use primereact components in my components, where I extend components classes with my classes (ex: `MyDialog extends Dialog`) to redefine DOM creating methods and change dom structure to my requirements. I use typescript, so I use these components typings. However I'm getting warnings both in my IDE and during build, that component's class misses most of it's methods.

Solution:
Add components inner methods to theirs classes typings.


melloware (approved):

mertsincan:
Hi @sshulin, 

We refreshed our code structure and turned our Components into functional Hook components which are currently popular in the React ecosystem. So if you can send a new PR after trying it, I'll be happy to merge it.

Sorry for this confusion!
Best Regards,

PR #2722, opened 2022-04-06
GitHub Actions

melloware:
###Feature Requests
Need a `package-lock.json` for GitHub Actions to run automated builds upon PR's. Will also need this for dependabot

here is a great post on why this should be in source control: https://stackoverflow.com/a/44210813/502366


melloware:
Moved all build tools to `devDependencies` and now have `npm run lint` working to check lint rules

components/lib/confirmdialog/ConfirmDialog.js
```diff
@@ -81,6 +82,7 @@ export const ConfirmDialog = React.memo(React.forwardRef((props, ref) => {
         return () => {
             OverlayService.off('confirm-dialog', confirm);
         }
+        // eslint-disable-next-line react-hooks/exhaustive-deps
```
melloware, line 85: @mertsincan to make these pass the lint rules I had to do this but wasn't sure if you wanted to fix them all the "right" way to not have any of these eslint ignores.

mertsincan: @melloware It can stay that way for now. We can improve on this in future versions. Thanks a lot!

PR #2742, opened 2022-04-11
Fix #2741: SelectItemOptionsType definition

### LINKED ISSUE #2741: SelectButton: SelectItem[] TypeScript Def

- [X] feature request

I'm requesting a feature to be able to style the splitbuttons!
 
**Current behavior**
Hey!

At the moment, a split button is wrapper above few buttons, but we can not style the underlying buttons natively, even though buttons have a variety of available styles.
The most obvious missing feature is changing the color, from `primary` to `secondary`/`warning`/any other of the available styles, but other features such as having it be an outlined/raised/rounded buttons are also relevant.

The only way to style buttons at the moment is via CSS, which means duplicating the style we want instead of using the existing class.

image

image

**Expected behavior**
Being able to style the buttons via one of these methods:
 - Passing a `buttonClassName` like #2461, which would add classes to the child buttons
 - Passing a prop with the desired style, which would apply the adequate classes automatically: `<SelectButton buttonStyle={{color: "secondary", style:"raised"}}`
 
image

melloware:
###Defect Fixes
Fix #2741: SelectButton options TypeScript def

Not sure if this is correct or not!

### DIFF
--- a/components/lib/cascadeselect/CascadeSelect.d.ts
+++ b/components/lib/cascadeselect/CascadeSelect.d.ts
@@ -1,5 +1,6 @@
 import * as React from 'react';
 import { CSSTransitionProps } from '../csstransition';
+import { SelectItemOptionsType } from '../selectitem/SelectItem';
 
 type CascadeSelectItemTemplateType = React.ReactNode | ((option: any) => React.ReactNode);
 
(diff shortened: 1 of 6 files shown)


mertsincan:
Thanks a lot, @melloware ;)

PR #2750, opened 2022-04-13
Fixed #2748 - Calendar month and year picker Improvement

### LINKED ISSUE #2748: Calendar - Year and month picker enhancement
https://github.com/primefaces/primeng/commit/4183276e59548b1e25735647ac7d2d3013eb24ec

mcandu:


components/lib/calendar/Calendar.js
```diff
@@ -320,11 +342,47 @@ export const Calendar = React.memo(React.forwardRef((props, ref) => {
...
+
+        for (let i = start; i <= end; i++) {
+            yearOptions.push(i);
+        }
+
+        setYearOptions(_yearOptions)
+    }
+
+    const decrementYear = () => {
+        setCurrentYear(prevState => prevState - 1)
```
melloware, line 369: Just double checking but when you call `setCurrentYear` i don't think the `currentYear` variable is updated immediately is it?  it might be better to use a local variable and then call `setCurrentYear(localVar)` and use the `localVar` everywhere in this method.

Same comment for the methods below...

melloware (changes requested):

components/lib/calendar/Calendar.js
mertsincan: I agree.

pages/calendar/index.js
```diff
@@ -161,23 +157,10 @@ const CalendarDemo = () => {
                             <label htmlFor="touchUI">TouchUI</label>
                             <Calendar id="touchUI" value={date13} onChange={(e) => setDate13(e.value)} touchUI />
                         </div>
-                        <div className="field col-12 md:col-4">
-                            <label htmlFor="mask">Mask</label>
-                            <Calendar id="mask" value={date14} onChange={(e) => setDate14(e.value)} mask="99/99/9999" />
```
mertsincan, line 166: @mcandu Why did you remove these demos? They were added based on user feedback and to show extra features that differ from other libraries.

mcandu: @mertsincan Because PrimeNG and PrimeVue demos are sync. I thought PrimeReact demos should be the same like these.

components/lib/calendar/Calendar.js
```diff
@@ -1224,16 +1283,63 @@ export const Calendar = React.memo(React.forwardRef((props, ref) => {
...
+
+
         event.preventDefault();
     }
 
+
+    const switchToYearView = (event) => {
+        setCurrentView('year');
+
+
```
melloware, line 1305: can probably remove these spaces too

melloware (changes requested):

PR #2757, opened 2022-04-15
Fix #2683: Overlays should close on browser scroll

### LINKED ISSUE #2683: Dropdown: rendering problem when used in a Dialog
https://codesandbox.io/s/primereact-dialog-dropdown-issue-teiqoz

**Current behavior**
The PrimeReact Dropdown component, when used in a PrimeReact Dialog, suffers from a rendering problem where the dropped-down list becomes detached from the input when the dialog's parent is scrolled.

**Expected behavior**
Dropdown should not be affected by scrolling the parent component.

Create a component that uses a Dialog component. In the Dialog include a Dropdown that has filtering turned on. Make sure the Dialog's parent has enough content to show scrollbars. Open the dialog, drop down the list, scroll the parent - observe that the dropped-down list moves with the parent rather than staying in place with the list input.

Issue occurs both locally on Windows 10 dev PC and in Codesandbox.

* **React version:**
17.0.2

* **PrimeReact version:**
7.2.0

* **Browser:** 
Chrome 99
 
* **Language:** 
All

image

### LINKED ISSUE #3656: Dropdown: Opened panel doesn't update it's position when used in the Bootstrap modal
### Describe the bug

When dropdown used in the Bootstrap modal window and it's panel is shown, the scrolling in the modal window causes reposition of the dom elements, but the panel is in fixed position.

### Reproducer

https://codesandbox.io/s/happy-albattani-ngqnfc

### PrimeReact version

8.7.2

### React version

18.x

### Language

ALL

### Build / Runtime

Create React App (CRA)

### Browser(s)

_No response_

1. In the CodeSanbox example, click on "Show Modal" button
2. Click on dropdown to see the option list
3. Move the mouse pointer to not appear upon the panel (to scroll down the page, not the list)
4. Scroll the page and the panel is in fixed position, on the other hand the dropdown element is moving

### Expected behavior

The panel should to be rendered on same position as it's parent dropdown input.

### LINKED ISSUE #4223: Mention: suggestion panel to hide when scrolling away

Improve the suggestion panel behavior to hide when scrolling away from the trigger position. 

Currently the suggestion panel has a "sticky" behavior when used in a textarea that has a `max-height`: 

scroll-bug-primereact

Something similar to what's implemented on Facebook. The suggestion panel will hide when scrolling away from the place where it's been triggered:
 
facebook-scroll

I've considered using the getOverlay() method to embed the suggestion panel inside a popover component from Floating UI that handles that "hiding" behavior: https://floating-ui.com/docs/hide

The problem in that case it that it will break the arrow navigation, etc...

### Additional context

reproduction link: https://stackblitz.com/edit/react-5snlta?file=src%2FApp.tsx

melloware:
###Defect Fixes
Fix #2683 : Overlays should close on browser scroll
Fix #3656
Fix #4223 Mention: suggestion panel to hide when scrolling away

@mertsincan there are two issues here.

1. it was passing `targetRef` instead of `target` to the `useOverlayScrollListener`

2. it is not handling browser scrolling for closing the overlay when the browser is scrolled.  Here is what we do in PF...

```js
if (scrollParent.is('body') || scrollParent.is('html') || scrollParent[0].nodeType === 9) { // nodeType 9 is for document element
    scrollParent = $(window);
}
```

Basically register the `window` as the scroll monitor if the parent is the `document`, `body`, or `html` nodes.  This now closes the overlays properly whenever the browser is scrolled.


mertsincan:
While this structure was being designed, document's scroll was also included in this structure. User feedback and in our tests were that closing the overlays when scrolling the document made them quite useless. So I will look for a different solution without including the document in it.

melloware:
OK for now we should definitely at least fix the `target: target, listener: event => {` its pointed to targetRef which is incorrect.

marcinb1986:
Hello.
I have implementation of Dropdown as below. Sorry but I cannot figure out based on You replies how to fix this bug
 <StyledDropdown
                    name={`luxmeds.${index}.luxmedType.package`}
                    value={luxmedType.package}
                    options={luxmedPackages}
                    onChange={handleChange}
                    placeholder="Select a package"
                  />

melloware:
@marcinb1986 you can't until this patch is applied.  This PR is still open and has not been applied yet.

marcinb1986:
@melloware thanks for reply.
I was thinking to close the Dropdown once screen is scrolled. 
I see that below aria-expanded changes to true/false once window is opened/closed. I was thinking to write some logic which will return false once screen have been scrolled, but when I add to <Dropdown/> implementation aria-expanded as prop but I don't see any reaction in the app. Do You think it's possible to use somehow aria-expanded ? 
image

melloware:
Nope it needs to be fixed with my patch above in the core code.

marcinb1986:
Ok, clear. Last question and I don't bother You anymore. When fix will be uploaded ? 
It' asking becasue I need to estimate if ok for my project to wait for it or I shall search for another library.
regards

melloware:
It still in discussion by PrimeTek.  This patch has been open since April so I can't give you a timetable if it will be implemented.

melloware:
you in your Dialog you can always do `<Dialog blockScroll ...` and blockScroll will disable the browser scrollbar while the dialog is open.

melloware:
Fix #3656

melloware:
Fixed

artonio:
I'm looking forward for this fix!

melloware:
This ticket suggests this is the correct behavior and what Facebook does.  Closes the overlay upon scrolling.

https://github.com/primefaces/primereact/issues/4223

lagroms:
Any news for this feature ?

melloware:
I have updated this PR to make it a configuration.

```js
PrimeReact.hideOverlaysOnViewportChange = true;
```

> Define behavior if the browser window is scrolled while displaying an overlay panel like a Dropdown or Calendar. Dpending on your organization's accessibility needs some prefer panels to be closed on scrolling and some prefer the overlay follow the scroll. The default behavior will be to close the overlay if you are scrolling the outer window as you are no longer interacting with the component.

Caarlos7x:
I was have a problem with on dropdown list... I tried to fix following the first comment from @melloware  but doesent work
so I put 'appendTo' and work as perfect

```javascript
<Dropdown
            ref={dropdownRef}
            value={value}
            onChange={(e) => onValueChange(e.value)}
            options={options}
            optionLabel={optionLabel}
            placeholder={placeholder}
            filter={filter}
            filterBy={filterBy}
            onFilter={onFilter}
            valueTemplate={valueTemplate}
            itemTemplate={itemTemplate}
            className={`dropdown-menu ${className}`}
            appendTo="self"
            {...rest}
        />
```

PR #2771, opened 2022-04-19
Fix #1661: Calendar do not navigate past min and max

### LINKED ISSUE #1661: Calendar: shows future and/or past dates even when minDate and/or maxDate props are set
**I'm submitting a ...** 
```
[x] bug report
[ ] feature request
[ ] support reques
```

**Codesandbox Case**
https://primefaces.org/primereact/showcase/#/calendar

**Current behavior**
The Calendar component shows past and future dates even when minDate and maxDate are set.  
Case in point: the "MinMax" example in the documentation. You can go past minDate and maxDate.

**Expected behavior**
Although disabled, dates set before minDate and after maxDate (whichever is present, if any) should not be shown.

* **React version:**
16.3.1

* **PrimeReact version:**
15.0.1

melloware:
###Defect Fixes
Fix #1661: Calendar do not navigate past min and max

@mertsincan this actually is much easier now that you switched to hooks because I can put this code in one place when the overlay panel renders and update the navigators appropriately without and additional property.

### DIFF
--- a/components/lib/calendar/Calendar.js
+++ b/components/lib/calendar/Calendar.js
@@ -947,6 +947,51 @@ export const Calendar = React.memo(React.forwardRef((props, ref) => {
         }
     }
 
+    const setNavigationState = (newViewDate) => {
+        if (props.view !== 'date' || !overlayRef.current) {
+            return;
+        }
+
+        const navPrev = DomHandler.findSingle(overlayRef.current, '.p-datepicker-prev');
+        const navNext = DomHandler.findSingle(overlayRef.current, '.p-datepicker-next');
+
+        if (props.disabled) {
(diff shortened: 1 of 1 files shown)


mertsincan:
@melloware, Thanks a lot for this PR! You need to add a property to Calendar for this. We should not break the current default behavior. This feature is optional.

melloware:
I picked `showOutliers` as the property as I could not think of a better name for what we are allowing to be displayed.

mertsincan:
Maybe, its name can be 'showMinMaxRange' and the default value is 'false' to be compatible with other Prime* libraries. WDYT?

melloware:
OK updated to `showMinMaxRange` and defaulted to `false`.

mertsincan:
Thanks @melloware ;)

PR #2772, opened 2022-04-19
Fix #2671: Showcase store theme in local storage

### LINKED ISSUE #2671: Showcase: Save theme setting in localStorage
```
[ ] bug report
[X] feature request
```

Hello, this bug is related only with your documentation page: `https://primefaces.org/primereact/`

**Current behavior**
When I go into your page and select for example: `saga-blue`, then its not saved for next browser session. It's always starts from `lara-dark-indigo` in my case.

**Expected behavior**
Please save chosen theme and then on start check if something is saved. Setting theme always based on user device settings is really annoying.

melloware:
###Defect Fixes
Fix #2671: Showcase store theme in local storage

Added new hook `useStorage` which can be a drop in replacement for anywhere you use `useState`.  You can choose `session` or `local` storage but it defaults to `local` storage


mertsincan:
Hi @melloware,

Could you please create a new issue about useStorage? We can add it. I'm waiting for feedback about the showcase issue from @cagataycivici .

mertsincan:
This commit has been manually reverted manually for now as it causes hydration problems.

PR #2780, opened 2022-04-21
Fix font family of npm install command on getting started section

inad9300:
Not everyone has a Mac :wink:


mertsincan:
Thanks a lot, @inad9300 ;)

PR #2809, opened 2022-04-27
Fix #2808: MegaMenu start and end template attributes

### LINKED ISSUE #2808: MegaMenu: Start and End templates like MenuBar

Logo on MegaMenu similar to Menubar
Sometimes for creating menu with a lot of content MegaMenu is a much better solution
However for styling the menu its not possible to add the app logo at the start of the menu.

melloware:
###Feature Requests
Fix #2808: MegaMenu start and end template attributes

This now brings it in line API wise with the `MenuBar`.

image

### DIFF
--- a/components/lib/menubar/Menubar.js
+++ b/components/lib/menubar/Menubar.js
@@ -52,18 +52,6 @@ export const Menubar = React.memo(React.forwardRef((props, ref) => {
         useCustomContent
     }));
 
-    const createCustomContent = () => {
-        if (props.children) {
-            return (
-                <div className="p-menubar-custom">
-                    {props.children}
-                </div>
-            )
-        }
-
-        return null;
-    }
-
     const createStartContent = () => {
         if (props.start) {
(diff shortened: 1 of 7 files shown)


components/lib/menubar/Menubar.js
```diff
@@ -52,18 +52,6 @@ export const Menubar = React.memo(React.forwardRef((props, ref) => {
         useCustomContent
     }));
 
-    const createCustomContent = () => {
```
melloware, line 55: This method was dead unused code in MenuBar.

mertsincan:
Hi @melloware,

Thanks a lot for this feature! Were you able to test it with orientation="vertical"? Also, can you create a separate demo for this instead of replacing the existing demo?
```
<h5>Templating</h5>
...
```

melloware:
OK good points and I think I definitely need to do more work on vertical may need your help with some CSS 😄

melloware:
Fixed and updated examples.

image

PR #2815, opened 2022-04-28
Fix #2814: Carousel less items than allocated slots

### LINKED ISSUE #2814: Carousel: Display issues when loading less items that the allocated slots
### Describe the bug

Originally reported on PrimeFaces: https://github.com/primefaces/primefaces/issues/8265

The bug is in the calculation of the `totalIndicators` value.

### Reproducer

https://github.com/eframsergio/tests/raw/main/primefaces-test-master.zip

### PrimeReact version

8.0.0

### React version

18.x

### Language

ES6

### Build / Runtime

Next.js

### Browser(s)

ALL

1 Create a Carousel with only 1 data item

```js
productService.getProductsSmall().then(data => setProducts(data.slice(0, 1)));
```

2. Set its numVisible={1} like...

```xml
 <Carousel value={products} numVisible={3} numScroll={1}  itemTemplate={productTemplate} />
```

### Expected behavior

image

melloware:
###Defect Fixes
Fix #2814: Carousel less items than allocated slots

### DIFF
--- a/components/lib/carousel/Carousel.js
+++ b/components/lib/carousel/Carousel.js
@@ -42,9 +42,10 @@ export const Carousel = React.memo(React.forwardRef((props, ref) => {
     const isVertical = props.orientation === 'vertical';
     const circular = props.circular || !!props.autoplayInterval;
     const isCircular = circular && props.value.length >= numVisibleState;
-    const isAutoplay = props.autoplayInterval && allowAutoplay.current;
     const currentPage = props.onPageChange ? props.page : pageState;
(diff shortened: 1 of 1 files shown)


melloware:
The bug was that totalIndicator could be a negative number for example if items=1 and numvis=3

```js
Math.ceil((props.value.length - numVisibleState) / numScrollState) + 1
```

```js
Math.ceil((1 - 3) / 1) + 1 === -1. // bad!
```

melloware:
This also fixes NPE with `circular` if props.value === null

PR #2831, opened 2022-05-02
Fix #1919: Toggle button WCAG standards

### LINKED ISSUE #1919: ToggleButton: WCAG standards
In [your docs](https://www.primefaces.org/primereact/showcase/#/accessibility) it says:

> UI Components for everyone with **full support** for Web Content Accessibility Guidelines (WCAG).

As already[ pointed out by another user](https://github.com/primefaces/primereact/issues/1887), this isn’t true. Neither your website nor _all_ components are WCAG 2.0 or 2.1 compliant. Since some companies base their choice for a specific library on the fact whether or not it’s WCAG compliant, it would be important to either rephrase the statements on the page and list all inaccessible components or to provide us with the information which expert checked which version of the library against which version of WCAG.

Thank you

melloware:
###Defect Fixes
Fix #1919: Toggle button WCAG standards

ToggleButton  should use an HTML button per https://inclusive-components.design/toggle-button/

### DIFF
--- a/components/doc/togglebutton/index.js
+++ b/components/doc/togglebutton/index.js
@@ -155,7 +155,7 @@ import { ToggleButton } from 'primereact/togglebutton';
 </CodeHighlight>
 
                     <h5>Properties</h5>
-                    <p>Any valid attribute is passed to the root element implicitly, extended properties are as follows;</p>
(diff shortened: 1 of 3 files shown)


mertsincan:
I'll review this PR as I begin accessibility improvements.

PR #2835, opened 2022-05-03
Fix #2833: MultiSelect only fire onBlur when focus is lost

### LINKED ISSUE #2833: MultiSelect: onBlur event mismatch
### Describe the bug

The component's event "onBlur" is fired on, not only when focus is lost.

### Reproducer

https://codesandbox.io/s/objective-wind-13w45j?file=/src/App.jsx:448-459

### PrimeReact version

7.2.1

### React version

17.x

### Language

ALL

### Build / Runtime

Create React App (CRA)

### Browser(s)

_No response_

1. click on the `MultiSelect` to show the list of options
2. pick any option
3. `onBlur` event will increase counter above 

1. click on the `MultiSelect` to show the list of options and then click on filter input or "select all" checkbox
2. `onBlur` event will increase counter above 

### Expected behavior

The `onBlur` event should to be triggered only when then component as a whole will lost the focus.

So it shouldn't be triggered during options selection nor filtering, but only when user will click (on|to) some(thing|where) else.

melloware:
###Defect Fixes
Fix #2833: MultiSelect only fire onBlur when focus is lost

### DIFF
--- a/components/lib/multiselect/MultiSelect.js
+++ b/components/lib/multiselect/MultiSelect.js
@@ -331,8 +331,10 @@ export const MultiSelect = React.memo(React.forwardRef((props, ref) => {
     }
 
     const onBlur = (event) => {
-        setFocusedState(false);
-        props.onBlur && props.onBlur(event);
+        if (!overlayVisibleState) {
+            setFocusedState(false);
+            props.onBlur && props.onBlur(event);
+        }
     }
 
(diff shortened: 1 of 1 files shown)


mertsincan:
Thanks @melloware ;)

_Note: When I start accessibility improvements, I think about making improvements on such problems. For now, this PR can wait._

PR #2867, opened 2022-05-10
Fix #2866: wrong tooltip when trying to get to page 0

### LINKED ISSUE #2866: Paginator: Wrong check triggers on example
### Describe the bug

Example is checking if 
```
page < 0
```
When the correct would be to check if 
```
page < 1
```
because there is no page 0 on the paginator, making it impossible to jump to that page, which should trigger said tooltip.

You can reproduce the problem on the example page itself:
https://www.primefaces.org/primereact/datatable/paginator/

If you try to jump to page 0, the element shows nothing in the table.

### Reproducer

https://www.primefaces.org/primereact/datatable/paginator/

### PrimeReact version

8.0.0-rc.1

### React version

17.x

### Language

TypeScript

### Build / Runtime

Vite

### Browser(s)

Mozilla Firefox 100

1. Go to https://www.primefaces.org/primereact/datatable/paginator/
2. Scroll down to the example titled "Custom Paginator Template"
3. On "Go to", type 0
4. Press enter while still in that field
5. Said Example will now show nothing inside the table.

### Expected behavior

The tooltip should appear and it should not try to get to page 0.
```Value must be between 1 and ${options.totalPages}.```

HashBR:
You can't go to page 0, so the correct way is to show the tooltip if the page value is actually smaller than 1, and not 0 as it was before.

Issue here: https://github.com/primefaces/primereact/issues/2866

### DIFF
--- a/pages/datatable/paginator.js
+++ b/pages/datatable/paginator.js
@@ -40,7 +40,7 @@ const DataTablePaginatorDemo = () => {
     const onPageInputKeyDown = (event, options) => {
         if (event.key === 'Enter') {
             const page = parseInt(currentPage);
-            if (page < 0 || page > options.totalPages) {
+            if (page < 1 || page > options.totalPages) {
                 setPageInputTooltip(`Value must be between 1 and ${options.totalPages}.`);
             }
             else {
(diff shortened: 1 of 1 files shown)


melloware (approved):

melloware:
Nice analysis and PR

PR #2884, opened 2022-05-12
Fix #2878: Tooltip global respect data-pr-position

### LINKED ISSUE #2878: Tooltip: data-pr-position doesn't change where the tooltip goes
### Describe the bug

```data-pr-position``` position doesn't seem to be working on Global Tooltips.

I've tried
```js
data-pr-position="left"
data-pr-position="top"
data-pr-position="bottom"
```
And none of them seemed to change where my tooltip would appear. Which was always on the default "right" position.

I've found a similar issue that was fixed in the past: https://github.com/primefaces/primereact/issues/1811

### Reproducer

https://codesandbox.io/s/primereact-tooltip-position-2gyr8h?file=/src/index.js:641-657

or

https://2gyr8h.csb.app/

### PrimeReact version

8.0.1

### React version

18.x

### Language

ALL

### Build / Runtime

Create React App (CRA)

### Browser(s)

Mozilla Firefox 100

1. Go to the sandbox provied.
2. Mouse over "Big!" and "Smaller!" redboxes.
3. Tooltip will appear on the right even though on the code it's stating to appear on "top" of it.

### Expected behavior

```data-pr-position``` is supposed to let me pick where the Global Tooltip would show.

melloware:
###Defect Fixes
Fix #2878: Tooltip global respect data-pr-position

### DIFF
--- a/components/lib/tooltip/Tooltip.js
+++ b/components/lib/tooltip/Tooltip.js
@@ -172,7 +172,7 @@ export const Tooltip = React.memo(React.forwardRef((props, ref) => {
                     DomHandler.removeClass(elementRef.current, 'p-tooltip-active');
 
                     setVisibleState(false);
-                    setPositionState(props.position);
+                    setPositionState(getPosition(currentTargetRef.current));
                     currentTargetRef.current = null;
                     containerSize.current = null;
                     allowHide.current = true;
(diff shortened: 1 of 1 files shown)


mertsincan:
Hi @melloware,

Thanks a lot for this PR! I've seen some bugs in Tooltip related to this issue. Generally, we use methods to change the state, but try to use the new state before the component is updated. The problem seems to come from them. Fixed now. https://github.com/primefaces/primereact/commit/3125839e55f4acad6a3a3cfbce8139bb6c6a2eb6

PR #2895, opened 2022-05-15
Image: is not taking the full width

walidcherhane:
**Before :**

image

**After :**

image

### DIFF
--- a/components/lib/image/Image.css
+++ b/components/lib/image/Image.css
@@ -31,6 +31,11 @@
     cursor: pointer;
 }
 
+.p-image > img{
+    width: 100%;
+    height: 100%;
+}
+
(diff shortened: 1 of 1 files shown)


melloware:
@walidcherhane thank you for your contribution!! Do you mind opening an issue ticket so I can link this PR to an issue.  That is our standard practice.

melloware:
But this CSS will override someone saying `width="250" height="250"`

```xml
<Image src="https://www.primefaces.org/wp-content/uploads/2020/05/placeholder.png" height="250"` width="250" />
```

Your CSS will override that and force it to take 100% of the window?

Got to this Demo page and add your custom CSS you will see the issue:
https://primefaces.org/primereact/image/

PR #2911, opened 2022-05-19
Restrict Toast/Message SeverityType to the only four valid values

inad9300:


melloware:
Shouldn't the same be true for `MessageSeverityType` as well?   Also do you have any docs on how this actually does retrict it in TypeScript?

inad9300:
Yes, it seems that it should be the same for `MessageSeverityType`, I just happened to have found the limitation in toasts while working with them.

As for the information in TypeScript, if you are looking for general information on unions of literal types, this should be a start: https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#literal-types. Though that part isn't new to the code, right? I just removed the extra, odd bit that wasn't letting TypeScript report errors. So I believe a more appropriate question would be, why was the purpose of having `(string & {})` in addition to the set of allowed string literals?

melloware:
Yes that was my question how does `(string & {})` prevent other types.  Thats the part that isn't clear to me.

inad9300:
Wait, but I _removed_ `(string & {})`. Not sure if we are on the same page here yet :smile:

melloware:
HA whoops in my mind you were adding it got the lines backwards :).  I totally agree then can you also put the change in for MessageSeverity here too.

melloware (approved):

inad9300:
Done. I found `MessagesSeverityType` (plural) too in the process, and I changed it like the others, but it sounds like it should be combined with `MessageSeverityType` (singular).

melloware:
Thanks. I think they keep them separate for now just seems to be a pattern in all the components.   The only thing I am thinking is they left the open for people to define their own severity like `fatal` and then then could just use all the theming styles with `fatal` to basically make their own severity?  I will leave this for Mert to review.

inad9300:
Good catch! Done.

mertsincan:
Thanks a lot for your contribution, @inad9300 ;)

PR #2922, opened 2022-05-24
Fix #2921: FloatLabel with webkit autofill fix

### LINKED ISSUE #2921: FloatLabel: Incorrect position when using browser autofill
### Describe the bug

Good afternoon. When you click on the label field, it works correctly. But when choosing a value from the drop-down list provided by the browser (previously entered value in the field), the label moves to the top, when choosing a value from previously entered values, the label remains in the wrong place, but when changes are made to the field, the label returns to its place.

### Reproducer

https://codesandbox.io/s/naughty-poincare-5sxfrq?file=/src/demo/InputTextDemo.tsx

### PrimeReact version

8.1.0

### React version

18.x

### Language

TypeScript

### Build / Runtime

Create React App (CRA)

### Browser(s)

Chrome 101.0.4951.67

1. Focus on the InputText field
2. Select a value from previously entered values (the list is provided by the browser)
3. Select the previously entered value

### Expected behavior

When working with the InputText field, its label was placed at the place specified for it, and did not change position when entering data into this field

melloware:
###Defect Fixes
Fix #2921: FloatLabel with webkit autofill fix

Standardized to the same `top` value as the others.

### DIFF
--- a/components/lib/inputtext/InputText.css
+++ b/components/lib/inputtext/InputText.css
@@ -54,17 +54,13 @@
 }
 
 .p-float-label input:focus ~ label,
+.p-float-label input:-webkit-autofill ~ label,
 .p-float-label input.p-filled ~ label,
 .p-float-label textarea:focus ~ label,
 .p-float-label textarea.p-filled ~ label,
 .p-float-label .p-inputwrapper-focus ~ label,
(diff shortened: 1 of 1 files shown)


mertsincan:
Thanks, @melloware! Good catch!

PR #2930, opened 2022-05-25
Fix #2913: Password also color text

### LINKED ISSUE #2913: Password: Add same color to label of password info

Actually the password panel show the password meter and add a custom class to change the background color, but is not the same to the password info

_No response_

The default behavior is
chrome_SD8BpBDfGk

So what I expect is something like this but natively
chrome_hVgJbtbN5v

Take a look of the source code it would can add easily funcionality to change the behavior and add some alternative to change the color to the password info which it can be aligned with the password meter
chrome_bjdO3eBcQc

### Additional context

_No response_

mhndsbgyn:
###Defect Fixes

###Feature Requests

### DIFF
--- a/components/lib/password/Password.css
+++ b/components/lib/password/Password.css
@@ -15,6 +15,7 @@
 
 .p-password-meter {
     height: 10px;
+
 }
 
 .p-password-strength {
@@ -23,6 +24,21 @@
     transition: width 1s ease-in-out;
 }
 
+
+
 .p-fluid .p-password {
     display: flex;
+}
+
+.p-password-strength-weak
+{
+    color: red !important;
+}
+.p-password-strength-medium
+{
+    color: orange !important;
+}
+.p-password-strength-strong
+{
+    color: green !important;
 }
--- a/components/lib/password/Password.css
+++ b/components/lib/password/Password.css
@@ -25,20 +25,6 @@
 }
 
 
-
 .p-fluid .p-password {
     display: flex;
-}
-
-.p-password-strength-weak
-{
-    color: red !important;
-}
-.p-password-strength-medium
-{
-    color: orange !important;
-}
-.p-password-strength-strong
-{
-    color: green !important;
 }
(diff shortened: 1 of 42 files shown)


pages/inputnumber/index.js
```diff
@@ -27,6 +27,13 @@ const InputNumberDemo = () => {
     const [value19, setValue19] = useState(25);
     const [value20, setValue20] = useState(50);
 
+
+    //  InputNumber : Id not available in the OnChange #2908
+    // const onChangeMethod = (myID)=>{
+    //   setValue8(myID.value);
+    // }
+
+
```
melloware, line 36: Remove this change it looks like you were testing something.

components/lib/password/Password.js
```diff
@@ -262,8 +262,9 @@ export const Password = React.memo(React.forwardRef((props, ref) => {
                 <div className="p-password-meter">
                     <div className={`p-password-strength ${strength}`} style={{ width }}></div>
                 </div>
-                <div className="p-password-info">
-                    {infoTextState}
+
+                <div className={`p-password-strength-${strength}`}>
```
melloware, line 266: No follow the pattern above it it should be `p-password-info ${strength}` with a space in between

components/lib/password/Password.css
```diff
@@ -23,6 +24,21 @@
...
+{
+    color: red !important;
+}
+.p-password-strength-medium
+{
+    color: orange !important;
+}
+.p-password-strength-strong
+{
+    color: green !important;
```
melloware, line 43: Revert all the changes in this file the changes will go in the `theme` themselves so it can be colored per theme.

melloware (changes requested):

melloware (changes requested):
Only PrimeTek modifies the themes it comes from their theme designer.

melloware:
Going to close this PR for now. Its too far off base for the actual fix.

PR #2944, opened 2022-05-30
Update SelectItem import statement

Ruan20795:
The actual file name when pulling from npm is "selectitem" and not "SelectItem", so when running builds on case sensitive OS, the build fails:

image

### DIFF
--- a/components/lib/dropdown/Dropdown.d.ts
+++ b/components/lib/dropdown/Dropdown.d.ts
@@ -2,7 +2,7 @@ import * as React from 'react';
 import TooltipOptions from '../tooltip/tooltipoptions';
 import { CSSTransitionProps } from '../csstransition';
 import { VirtualScrollerProps } from '../virtualscroller';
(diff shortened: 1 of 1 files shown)


tamnguyen02121993:
I also have trouble same above issue. So please help me resolve it

melloware (approved):

melloware:
OK I see I am on Windows which is not case sensitive.  Thank you for reporting!

melloware:
Nevermind i will fix with another PR.

melloware:
OK I fixed the other imports as well!

Ruan20795:
@melloware I see that was tagged with milestone 8.2.0? This is clearly a bug and preventing builds on CI/CD workflows. Can this not be pushed as a bug fix?

melloware:
@Ruan20795 I don't work for PrimeTek I am just an open source committer. How and when they do releases is up to Primetek.

PR #2950, opened 2022-05-31
Fix #2948: Calendar monthpicker incorrect month on 1st click

### LINKED ISSUE #2948: Calendar: Selection Month DropDown Item at Index 1 for the first time, Selects Item at Index 2
### Describe the bug

Hey There, I am facing one issue in latest installed versions.

The issue is that I have used Calendar control which has Month and Year Dropdowns.

In Month Dropdown, I have Values from January till December

The problem is that For the First Click in this dropdown, If I select February(at Index 1), It selects March(at Index 2).

This Problem is only faced for the first click.

**Please Resolve this in Some Older Versions also!**

Please resolve this ASAP!

### Reproducer

_No response_

### PrimeReact version

8.1.1

### React version

17.x

### Language

ALL

### Build / Runtime

Create React App (CRA)

### Browser(s)

Chrome 101.0.4951

1) Click February which is at index 1 in Month Dropdown.
2) Clicking February, Selects March for the first time.
3) Again click February works fine.

### Expected behavior

1) Click February which is at index 1 in Month Dropdown.
2) Clicking February, should Select February for all the times.

melloware:
###Defect Fixes
Fix #2948: Calendar monthpicker incorrect month on 1st click

### DIFF
--- a/components/lib/calendar/Calendar.js
+++ b/components/lib/calendar/Calendar.js
@@ -1311,6 +1311,7 @@ export const Calendar = React.memo(React.forwardRef((props, ref) => {
             setCurrentMonth(month);
             createMonthsMeta(month, currentYear);
             const currentDate = new Date(getCurrentDateTime().getTime());
+            currentDate.setDate(1); // #2948 always set to 1st of month
             currentDate.setMonth(month);
             currentDate.setYear(currentYear);
 


melloware:
When using the view date since its using the Current Date Time when it sets the month and today is the 31st there is no Feb 31st so it makes it March.  You have to set the day to 1 to ensure the monthpicker is correct.

PR #2970, opened 2022-06-07
Fix #2963: Allow access to ref

### LINKED ISSUE #2963: Components: Cannot access ref
### Describe the bug

Ref is no longer accessible in Dropdown and many other components. This code used to work in 7.1

```xml
<Dropdown ref={myRef}  />
```

in 8.0 does nothing and ref is not accessible.

### Reproducer

https://codesandbox.io/s/primereact-test-forked-xztbhk?file=/src/index.js

### PrimeReact version

8.1.1

### React version

18.x

### Language

ALL

### Build / Runtime

Create React App (CRA)

### Browser(s)

ALL

1. Run the above reproducer.
2. See that `ref` has no values and can't be accessed.

From looking in the DropDown.js source code I can see the original `ref is never used.

### Expected behavior

The `ref` element to contain the value

melloware:
###Defect Fixes
Fix #2963: Allow access to ref

Many components were not forwarding the `ref` to the actual root `elementRef` of the component.

### DIFF
--- a/components/lib/autocomplete/AutoComplete.js
+++ b/components/lib/autocomplete/AutoComplete.js
@@ -13,7 +13,7 @@ export const AutoComplete = React.memo(React.forwardRef((props, ref) => {
     const [searchingState, setSearchingState] = React.useState(false);
     const [focusedState, setFocusedState] = React.useState(false);
     const [overlayVisibleState, setOverlayVisibleState] = React.useState(false);
-    const elementRef = React.useRef(null);
+    const elementRef = React.useRef(ref);
     const overlayRef = React.useRef(null);
     const inputRef = React.useRef(props.inputRef);
     const multiContainerRef = React.useRef(null);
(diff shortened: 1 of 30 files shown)


mertsincan:
Hi @melloware,
I think we need to do a separate study for this. Maybe, all components can have two refs;
- <Carousel ref -> ref.current.customWidgetVariable/Function
- <Carousel elementRef -> elementRef.current (HTML dom element)

ref: component's ref
elementRef: DOM element's ref

WDYT?

melloware:
So here is what I did I found every component where its `ref` was showing as grey and unused in VS Code.

image

Then I simply followed your example in other places where you already had an `elementRef` I just simply assigned it to the variable the user wants so its passed through as the ref like in the `DropDown` case.

But we can handle it any way you think is best I just followed how you were doing it other components!

melloware:
Actually i think you might be right I am thinking that maybe we do this...

1. `ref` is always pointed at `useImperativeHandle` in all components even if it has no widget methods to expose YET like this...

```js
    React.useImperativeHandle(ref, () => ({
        ...props
    }));
```

2. `elementRef` is always the root element of the component if its a DIV or SPAN or somethng.

mertsincan:
Yes, I'm thinking :D

melloware:
Closing this as its not the right PR.

PR #3011, opened 2022-06-29
Fix #3009: Datatable add virtualScrollerRef

### LINKED ISSUE #3009: Datatable: Add reference to virtualScroller

Please provide the possibility to get from datatable a  REF(erence) to the underlying virtualScroller Object/Module to call its exported methods      

I tried to position the datatable (with virtualscroller) to a specific Index.
I can see, that in the virtualscroller Module (virtualscroller.esm.js) there exists (exports) the function "scrollToIndex" .
But I can't get a reference to VirtualScroller to call this method/function.

datatable should export a ref to Virtualscroller

melloware:
###Defect Fixes
Fix #3009: Datatable add virtualScrollerRef


melloware:
This is consistent with how all the other components do it.

PR #3026, opened 2022-07-03
Fix #3023: Sort nulls like Excel by default at bottom

### LINKED ISSUE #3023: DataTable: Incorrect order by ASC (empty values are at the beginning)
### Describe the bug

image

When we sort table data by ASC order empty values at the beginning, these values must be at the end. In DESC everything is correct.

As for me, the problem is in `sort` -> `if (value1 == null && value2 != null) result = -1;else if (value1 != null && value2 == null) result = 1;` in such case empty values will be at the beginning.

### Reproducer

_No response_

### PrimeReact version

8.1.1

### React version

18.x

### Language

TypeScript

### Build / Runtime

Create React App (CRA)

### Browser(s)

Chrome

1. Make column sortable
2. Sort by ASC
3. Empty values at the beginning

### Expected behavior

Empty values must be at the end.

melloware:
###Feature Requests
Fix #3023: Sort nulls like Excel by default at bottom

This always sorts NULLs at the bottom which is the way Excel does it by default.

### DIFF
--- a/components/lib/datatable/DataTable.js
+++ b/components/lib/datatable/DataTable.js
@@ -861,8 +861,8 @@ export const DataTable = React.forwardRef((props, ref) => {
         return props.removableSort ? (props.defaultSortOrder === currentOrder ? currentOrder * -1 : 0) : currentOrder * -1;
     }
 
-    const compareValuesOnSort = (value1, value2) => {
-        return ObjectUtils.sort(value1, value2, 1, PrimeReact.locale);
+    const compareValuesOnSort = (value1, value2, order) => {
+        return ObjectUtils.sort(value1, value2, order, PrimeReact.locale);
     }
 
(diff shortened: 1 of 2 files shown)


mertsincan:
Thanks @melloware ;)

kbeeveer46:
Is there any way to enable the old setting where nulls are at the beginning in ascending order? I hope so. We've been telling our clients how to use the table and how it functions and now they are going to assume there are no nulls. Especially if there's 50+ pages of data. They're not going to know the nulls are on page 50 at the end. This is not something that should be changed without providing a setting to use the old method.

melloware:
@kbeeveer46 can you open a new issue please.  Since this sort mechanism is used across multiple components I am wondering if we can make it a global setting like `PrimeReact.sortNullsHigh = true;`

PR #3027, opened 2022-07-04
Fix #3019: Column sortFunction add rowData

### LINKED ISSUE #3019: Datatable: Can not get table data for Column sortFunction and sortFunction calls only by click in header
### Describe the bug

image

I'm using filtering and ordering for Data Table.
For one column I want to add a custom `sortFunction` but for it, I need table data and I can't get it anywhere, it is underhood of DataTable.

image

Could you provide data as an argument for this function?
Or provide a method to get this data. `onValueChange` calls after my custom `sortFunction` and I can't use this data.

Another problem is this function doesn't call when `sortField` and `sortOrder` are set. In my case, I get it from the back end on did mount. By default, these data are not set.

### Reproducer

_No response_

### PrimeReact version

8.0.0

### React version

18.x

### Language

TypeScript

### Build / Runtime

Create React App (CRA)

### Browser(s)

Chrome

_No response_

### Expected behavior

_No response_

melloware:
###Defect Fixes
Fix #3019: Column sortFunction add rowData.

I don't see how sortFunction can be useful unless the `rowData` is passed to it to actually do something and return the values.

### DIFF
--- a/components/lib/datatable/DataTable.js
+++ b/components/lib/datatable/DataTable.js
@@ -898,14 +898,14 @@ export const DataTable = React.forwardRef((props, ref) => {
         let value = [...data];
 
         if (columnSortable.current && columnSortFunction.current) {
-            value = columnSortFunction.current({ field, order });
+            value = columnSortFunction.current({ rowData: value, field: field, order: order });
         }
         else {
             value.sort((data1, data2) => {
                 const value1 = ObjectUtils.resolveFieldData(data1, field);
                 const value2 = ObjectUtils.resolveFieldData(data2, field);
 
-                return compareValuesOnSort(value1, value2, order);;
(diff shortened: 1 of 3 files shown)


components/lib/datatable/DataTable.js
```diff
@@ -898,14 +898,14 @@ export const DataTable = React.forwardRef((props, ref) => {
         let value = [...data];
 
         if (columnSortable.current && columnSortFunction.current) {
-            value = columnSortFunction.current({ field, order });
+            value = columnSortFunction.current({ value, field, order });
```
mertsincan, line 901: Hi @melloware,

I think you need to change `value` with `rowData: value`. Isn't it?

melloware:
I changed that line to use its specific JSON fields similar to TreeTable.

PR #3032, opened 2022-07-05
Fix #3031: Image add onError callback

### LINKED ISSUE #3031: Image: Add 'onError' property

I would love to be able to pass the 'onError' property through to the underlying img element of the Image component.

_No response_

I would like to react properly and show a fallback image if the Image component is not able to load the given image source. Right now it is only possible to show an alternative string instead, which doesn't look nice in every use case.

_No response_

### Additional context

_No response_

melloware:
###Defect Fixes
Fix #3031: Image add onError callback

Extending `span` element adds the necessary onError and other methods.  The OnError is added to the Image and not the SPAN.

### DIFF
--- a/components/lib/image/Image.d.ts
+++ b/components/lib/image/Image.d.ts
@@ -1,10 +1,8 @@
 import * as React from 'react';
 
-export interface ImageProps {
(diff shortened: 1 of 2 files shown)


SelasViktoria:
Thanks for reacting so quickly!

mertsincan:
Thanks a lot, @melloware ;) Could you please add this callback to doc?

melloware:
Yes will do!

PR #3047, opened 2022-07-09
Fix #2963: Expose ref and element

### LINKED ISSUE #2963: Components: Cannot access ref

melloware:
###Defect Fixes
Fix #2963: Expose ref and element

@mertsincan i think i got this right and tested it.  This exposes props the way PrimeReact 7 did as well as adds `elementRef` and `getElement()` exposed through `ref`.

So now you can do something like this...

```xml
<Dropdown ref={ref} ..>
```

```ts
ref.current.getElement();
```

To get the VirtualScroller of the Datatable...

```xml
<Datatable ref={ref} ..>
```

```ts
const virtualScroller = ref.current.getVirtualScroller();
```

### DIFF
--- a/components/lib/accordion/Accordion.d.ts
+++ b/components/lib/accordion/Accordion.d.ts
@@ -26,11 +26,8 @@ interface AccordionEventParams {
     index: number;
 }
 
-export interface AccordionProps {
-    id?: string;
+export interface AccordionProps extends Omit<React.DetailedHTMLProps<React.HTMLAttributes<HTMLDivElement>, HTMLDivElement>, 'ref'> {
     activeIndex?: AccordionActiveIndexType;
-    className?: string;
-    style?: object;
     multiple?: boolean;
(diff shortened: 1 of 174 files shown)


mertsincan:
Thanks @melloware ;)

PR #3060, opened 2022-07-13
Fix primefaces#2514: Fix sortFunction ignored on initial sort

emilytoy:
If `sortField` is specified, determine and set `columnSortable` and `columnSortFunction` refs when processing data.  Otherwise, the refs are false and undefined, which results in the `sortFunction` not being run initially.

###Defect Fixes
primefaces#2514

### DIFF
--- a/components/lib/datatable/DataTable.js
+++ b/components/lib/datatable/DataTable.js
@@ -1236,6 +1236,13 @@ export const DataTable = React.forwardRef((props, ref) => {
                 const sortField = (localState && localState.sortField) || getSortField();
                 const sortOrder = (localState && localState.sortOrder) || getSortOrder();
                 const multiSortMeta = (localState && localState.multiSortMeta) || getMultiSortMeta();
+                var columns = getColumns();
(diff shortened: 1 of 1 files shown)


components/lib/datatable/DataTable.js
```diff
@@ -1236,6 +1236,13 @@ export const DataTable = React.forwardRef((props, ref) => {
                 const sortField = (localState && localState.sortField) || getSortField();
                 const sortOrder = (localState && localState.sortOrder) || getSortOrder();
                 const multiSortMeta = (localState && localState.multiSortMeta) || getMultiSortMeta();
+                var columns = getColumns();
+                var sortColumn = columns.find((col) => col.props.field === sortField);
```
melloware, line 1240: use `const` instead of `var`

melloware (changes requested):

components/lib/datatable/DataTable.js
emilytoy: Done

melloware:
Thanks for the fix I had been looking at this one too!

PR #3082, opened 2022-07-20
Fix #3081: maintain p-filled properly

### LINKED ISSUE #3081: Mention, InputText, InputTextArea and Password do not maintain p-filled properly
### Describe the bug

When changing the input value, the p-filled is not updated, due to bad choice of dependencies for React.useMemo. I was able to reproduce this only during automated tests with Jest. It is not easy to reproduce this manually.

### Reproducer

_No response_

### PrimeReact version

8.3.0

### React version

17.x

### Language

ALL

### Build / Runtime

Create React App (CRA)

### Browser(s)

_No response_

_No response_

### Expected behavior

_No response_

kalinkrustev:
Fix #3081


melloware (approved):

melloware:
Instead of

```js
    const currentValue = elementRef.current && elementRef.current.value;
    const isFilled = React.useMemo(() => (
        ObjectUtils.isNotEmpty(props.value) || ObjectUtils.isNotEmpty(props.defaultValue) || ObjectUtils.isNotEmpty(currentValue)
    ), [props.value, props.defaultValue, currentValue]);
```

Would it be better to lazy evaluate the `current` rather than store it in a const?

```js
    const currentValue = elementRef.current && elementRef.current.value;
    const isFilled = React.useMemo(() => (
        ObjectUtils.isNotEmpty(props.value) || ObjectUtils.isNotEmpty(props.defaultValue) || (elementRef.current && ObjectUtils.isNotEmpty(elementRef.current.value))
    ), [props.value, props.defaultValue, elementRef.current]);
```

kalinkrustev:
The problem is that `elementRef.current.value` must be in the dependencies, otherwise it might change (without a change to elementRef.current) without triggering reevaluation of the memoized value.

melloware:
Ahh yes.

Issue #2101, opened 2021-06-09
Use KeyboardEvent.key instead of the deprecated KeyboardEvent.which and/or KeyboardEvent.keyCode.


ankitprahladsoni:
**I'm submitting a ...**  (check one with "x")
```
[x] bug report
[ ] feature request
[ ] support request => Please do not submit support request here, instead see https://forum.primefaces.org/viewforum.php?f=57
```

I initially filed a bug with `@testing-library/userEvent`, but they suggested to move it to PrimeReact.
Original issue: testing-library/user-event#684
And the suggested changes where mentioned here https://github.com/testing-library/user-event/issues/684#issuecomment-857498379

**Current behavior**
userEvent.type() doesn't work with InputNumber with decimal inputs

**Expected behavior**
userEvent.type() should work with InputNumber with decimal inputs

**Minimal reproduction of the problem with instructions**
Codesandbox: https://codesandbox.io/s/gallant-ishizaka-hv4n7?file=/src/aaa.test.js

**Please tell us about your environment:** Any

* **React version:**17.0.0

* **PrimeReact version:** 6.3.2

* **Browser:** All
 
* **Language:** All

DorukKantarcioglu:
Hello Ankit,

I think the reason why your test case fails is due to the _decimal separator_ in InputNumber being **"," (comma)** and not "."
When I ran your test case with "18,27", the test passes.

Kind regards

ankitprahladsoni:
I don't think I have provided the decimal separator as `,`. Moreover, the locale `locale="en-IN"` should not use a decimal separator as a comma anyway.
The test case on the test tabs still shows a failed test
![image](https://user-images.githubusercontent.com/8023641/123757809-2fd10c00-d873-11eb-88f9-bab1f8a184c5.png)

DorukKantarcioglu:
![image](https://user-images.githubusercontent.com/63489557/123759802-e9a69780-d8c8-11eb-8d1f-4311337e4c2b.png)

What I meant was to change the dot to a comma in the test variable

melloware:
Should this be closed since it was basically an invalid test?

melloware:
I think now that IE is dead this ticket should be implemented

habubey:
Issue tracker is used for defects only as part of our commitment to quality and continuous improvement in all areas. Enhancements are collected as valuable community feedback and managed internally so moving this enhancement ticket to our internal project management backlog.

PR #3102, opened 2022-07-24
Refactor KeyFilter

melloware:
###Defect Fixes
@mertsincan this uses `event.key` which is much more browser standard than the old `event.which` etc and is much more reliable.

It also allows us to remove all the special Safari and Opera and Mozilla handling this class was doing.

Here is a great Keyboard event viewer: https://w3c.github.io/uievents/tools/key-event-viewer.html

Also its universally supported now: https://caniuse.com/?search=event.key

Its a start of refactoring for #2101


melloware:
Also it looks like this code originated from PrimeNG so it was proabbly valid a long time ago with these IE and Opera etc fixes none of which should be necessary now.

mertsincan:
@melloware can you create a Github issue?

melloware:
Yep

melloware:
Done!

Issue #3109, opened 2022-07-26
InputNumber: Tab does not fire update events, meaning inputNumbers do not call callbacks in datatables


purxiz:
### Describe the bug

InputNumber elements used as editors in dataTables (in particular in cell edit mode) do not fire update events i.e. onCellEditComplete when the cell is tabbed out of.

### PrimeReact version

8.3.0

### React version

17.x

### Language

ALL

### Build / Runtime

Create React App (CRA)

### Steps to reproduce the behavior

Create any datatable with cell editor mode, and use an InputNumber as the input field, with an onCellEditComplete function. Observe that the onCellEditComplete function is called if "enter" is pressed, or the cell is clicked out of while editing, but not called if tab is pressed.

### Expected behavior

When tab is pressed, I expect callbacks like validators and onCellEditComplete functions to be called.

PR #3110, opened 2022-07-26
Add Tab Support to InputNumber (within Datatable)

purxiz:
When editing an inputnumber in the "cell" edit mode within a datatable, tab does not trigger the onCellEditComplete (or validators), because it is missing from the switch list. I found the same issue with the same solution in PrimeVue: https://github.com/primefaces/primevue/pull/1954/files.

###Defect Fixes
https://github.com/primefaces/primereact/issues/3109

### DIFF
--- a/components/lib/inputnumber/InputNumber.js
+++ b/components/lib/inputnumber/InputNumber.js
@@ -341,7 +341,7 @@ export const InputNumber = React.memo(React.forwardRef((props, ref) => {
                 }
                 break;
 
-            //enter
+            //enter and tab
             case 13:
             case 9:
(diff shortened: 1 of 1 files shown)


components/lib/inputnumber/InputNumber.js
```diff
@@ -343,6 +343,7 @@ export const InputNumber = React.memo(React.forwardRef((props, ref) => {
 
             //enter
```
melloware, line 344: Update comment to `//enter and tab`

melloware (changes requested):

melloware (approved):

PR #3128, opened 2022-08-02
Fix #3127: DataTableRowToggleParams TS updated

### LINKED ISSUE #3127: DataTable: onRowToggle event returns the wrong type for the data property
### Describe the bug

When clicking the row expand button, the `onRowToggle` event will log an object of the type [`DataTableExpandedRows`](https://github.com/primefaces/primereact/blob/5e60927bf91e967ff2e653a2a0d673b1214558b9/components/lib/datatable/DataTable.d.ts#L82), where the keys are the `dataKey`s of the toggled rows, and the value is a boolean that says whether the rows are toggled or not when it should log an array of data keys like the [typed value suggests](https://github.com/primefaces/primereact/blob/5e60927bf91e967ff2e653a2a0d673b1214558b9/components/lib/datatable/DataTable.d.ts#L90). 

image

By the way, it only ever shows the ids of the rows that are toggled, so the value is always `true`, hence the suggestion. As the expected type in the d.ts would make most implementations easier without the need to use the `Object. keys()` method which is usually what most people want, the keys of the toggled values. 

**But just making the `DataTableRowToggleParams['data']` equals to [`DataTableExpandedRows`](https://github.com/primefaces/primereact/blob/5e60927bf91e967ff2e653a2a0d673b1214558b9/components/lib/datatable/DataTable.d.ts#L82), would fix the issue.**

### Reproducer

https://codesandbox.io/s/epic-sunset-xmuhit?file=/src/demo/DataTableRowExpansionDemo.tsx

### PrimeReact version

8.3.0

### React version

18.x

### Language

TypeScript

### Build / Runtime

Create React App (CRA)

### Browser(s)

Chrome 103.0.5060.134

1. Click the row expand button.
2. The console will log an object of the type Record<string | number, boolean>, where the keys are the dataKeys of the toggled rows, and the value is a boolean that says whether the rows are toggled or not. The problem is that it only ever shows the ids of the rows that are toggled, so the value is always true.

So it doesn't even make sense to have an object for that. **A list containing the data keys of the toggled rows makes more sense.** 

### Expected behavior

The console will log an object of the type **`Array` containing the data keys of the toggled rows makes more sense.** Which could be the `any[]` that the type [`DataTableRowToggleParams`](https://github.com/primefaces/primereact/blob/5e60927bf91e967ff2e653a2a0d673b1214558b9/components/lib/datatable/DataTable.d.ts#L90) already specifies for the `data` parameter.

melloware:
###Defect Fixes
Fix #3127: DataTableRowToggleParams TS updated

### DIFF
--- a/components/lib/datatable/DataTable.d.ts
+++ b/components/lib/datatable/DataTable.d.ts
@@ -87,8 +87,7 @@ interface DataTableEditingRows {
     [key: string]: boolean;
 }
 
-interface DataTableRowToggleParams {
-    data: any[];
+interface DataTableRowToggleParams extends DataTableExpandedRows {
 }
 
 interface DataTableColumnResizeEndParams {


saulodias (approved):
This seems correct, now.

PR #3130, opened 2022-08-02
Fix #2913:  Password label is same as meter color

### LINKED ISSUE #2913: Password: Add same color to label of password info

habubey:
###Defect Fixes
Fix #2913: Add same color to label of password info


melloware:
@habubey never make changes in the theme.css that is controlled by PrimeTek.

melloware:
I have assigned to @mertsincan as he needs to make the theme changes.

melloware (approved):

melloware (approved):

PR #3132, opened 2022-08-03
Fix #3044: FileUpload expose onFileSelect

### LINKED ISSUE #3044: FileUpload: Ability to add file from variable

Before today, we were on 7.2.1. We have _two_ options for selecting a file: upload a local file or take a photo with your camera. When using the camera option (using react-webcam), we did the following in order to add it to the File

```typescript
        // ...
	const fileUploadRef = useRef<FileUpload & { onFileSelect(event): void }>(null)

	const onCameraFinish = (file: File) => {
		setShowCamera(false)
		const event = {
			target: {
				files: [file],
			},
		}
		// Hack: Simulate selection of a file to add it to FileUpload component
		fileUploadRef.current?.onFileSelect(event)
	}

	return (
		<>
			<Camera isActive={isShowCamera} onFinish={onCameraFinish} />
			<FileUpload
				ref={fileUploadRef}
				multiple
				customUpload
				uploadHandler={customUploader}
				// ...
			/>
		</>
	)
```

After upgrading to 8.2.0, `fileUploadRef.current?.onFileSelect` is no longer defined.

We would like a way to add files to the component without going through the normal file selection process.

We are unable to upload files gotten by other means than the normal file selection process.

Some way of adding files to FileUpload's list of files programmatically.

I have looked at the defined properties of the ref and found no suitable members. I have considered using a different component for photo uploads, but this would be a major and unacceptable downgrade. I have considered downgrading, but this is not possible as we are at the end of a very much needed upgrade process. We need React 18 support.

### Additional context

There may be a way of doing this that I'm not aware of, but it's not clear from docs or anything I've found on Google.

melloware:
###Defect Fixes
Fix #3044: FileUpload expose onFileSelect

### DIFF
--- a/components/lib/fileupload/FileUpload.d.ts
+++ b/components/lib/fileupload/FileUpload.d.ts
@@ -82,7 +82,7 @@ interface FileUploadRemoveParams {
 }
 
 interface FileUploadProps {
-    id?: string;
+    id?: void;
     name?: string;
     url?: string;
     mode?: FileUploadModeType;
@@ -130,6 +130,7 @@ export declare class FileUpload extends React.Component<FileUploadProps, any> {
     public upload(): void;
     public clear(): void;
     public formatSize(bytes: number): number;
+    public onFileSelect(event: FileUploadSelectParams): void;
     public getElement(): HTMLElement;
     public getInput(): HTMLInputElement;
 }
--- a/.babelrc
+++ b/.babelrc
@@ -1,0 +1,3 @@
+{
+  "presets": ["next/babel"]
+}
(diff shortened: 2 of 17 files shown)


components/lib/fileupload/FileUpload.d.ts
```diff
@@ -130,6 +130,7 @@ export declare class FileUpload extends React.Component<FileUploadProps, any> {
     public upload(): void;
     public clear(): void;
     public formatSize(bytes: number): number;
+    public onFileSelect(event: FileUploadSelectParams): boid;
```
Audiopolis, line 133: Should this be `void`?

melloware: Yep this should have been caught.  THe TypeScript checker was not doing its job because of a setting I just found and fixed and it then found a bunch of other issues

PR #3135, opened 2022-08-04
Fix #3134: FileUpload onBeforeSelect event

### LINKED ISSUE #3134: FileUpload: Needed Event: onBeforeSelect

When using custom template and adding multiple files I want to make some controls before file listed as an item.

_No response_

If my controls fail then I want to return false so I don't want to see the file in the file list

_No response_

### Additional context

_No response_

melloware:
###Defect Fixes
Fix #3134: FileUpload onBeforeSelect event

### DIFF
--- a/api-generator/components/fileupload.js
+++ b/api-generator/components/fileupload.js
@@ -226,6 +226,22 @@ const FileUploadEvents = [
         ]
     },
     {
+        name: 'onBeforeSelect',
+        description: 'Callback to invoke before files are selected. Return false from callback to prevent selection.',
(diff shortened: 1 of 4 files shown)


sutherlandnele:
how do i get the list of files using onBeforeSelect prop. I cannot get it from the event object that is passed to the handler method. I am trying to validate the file size and file type uploaded using toast messages and not show invalid files in the pending file list of the FileUpload control.

I am using primereact version 9.2.3 which has the onBeforeSelect prop available. please assist.

Issue #2224, opened 2021-07-29
Dropdown: Wrong event value on OnBlur


dattebayorob:
https://codesandbox.io/s/primereact-test-forked-uhhdy?file=/src/index.js:746-761

**Current behavior**

Wrong event value for onBlur leeds to issues when working with Form libraries, like Formik.
**Expected behavior**

The event.target from onBlur event should contains name, id and value

PR #3137, opened 2022-08-04
Wrong event value onBlur fixed

habubey:
###Defect Fixes
Fix-#2224--Dropdown wrong event value onBlur fixed


melloware:
I don't think this is right its passing the `inputs` onBlur event which is properly defined as a FocusEvent

```ts
onBlur?(event: React.FocusEvent<HTMLInputElement>): void;
```

melloware:
I am going to close this. I fixed it the correct way: https://github.com/primefaces/primereact/pull/3164

PR #3154, opened 2022-08-11
Fix #3153: update the current Calendar view on change (#3153)

### LINKED ISSUE #3153: Calendar: does not change "view" dynamicly
### Describe the bug

The behavior when you switch the `view` of the Calendar e.g. from 'date' to the 'month' does not work as expected.

### Reproducer

https://codesandbox.io/s/primereact-test-forked-jx8kt6

### PrimeReact version

8.3.0

### React version

17.x

### Language

TypeScript

### Build / Runtime

Create React App (CRA)

### Browser(s)

_No response_

1. Render the Calendar with the `view="date"`.
2. Change the `view` to the `month`.

**Actual result**: only the head of the picker is rendered.

### Expected behavior

The whole picker should be rendered after the `view` is changed dynamically.

kazamov:
Fix for the [https://github.com/primefaces/primereact/issues/3153](url)

### DIFF
--- a/components/lib/calendar/Calendar.js
+++ b/components/lib/calendar/Calendar.js
@@ -2452,6 +2452,10 @@ export const Calendar = React.memo(React.forwardRef((props, ref) => {
     });
 
     useUpdateEffect(() => {
+        setCurrentView(props.view);
+    }, [props.view])
+
+    useUpdateEffect(() => {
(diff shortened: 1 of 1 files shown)


melloware (approved):

melloware:
thanks for the diagnosis and the fix.  I agree with this fix.

kazamov:
@melloware thanks for the quick feedback

PR #3164, opened 2022-08-18
Fix #2224: Dropdown correct onBlur event

### LINKED ISSUE #2224: Dropdown: Wrong event value on OnBlur

melloware:
###Defect Fixes
Fix #2224: Dropdown correct onBlur event


melloware:
Need to do the delay here because the Input is actually firing on Blur before the option has been selected so we need to wait 200ms for the `inputRef` to have the correct value.

PR #3170, opened 2022-08-22
Fix #3169: MenuItem add visible property

### LINKED ISSUE #3169: MenuModel property for visibility like in Vue

PrimeVue currently has a property in the MenuModel called "visible" which allows the element to be shown or hidden as needed, it would be good if this option was also available in PrimeReact.

Not related to a problem at all.

The same property as in Vue for the React MenuModel

Two alternatives would be to use "classNames" where the "hidden" class is added as needed. Or by setting the display styles directly in the styles property of the MenuModel.

### Additional context

_No response_

melloware:
###Feature Requests
Fix #3169: MenuItem add visible property


melloware:
Updated to not even render the DOM element if not visible.

PR #3173, opened 2022-08-23
Add prettier to standardize formatting

melloware:
###Defect Fixes
Add prettier to standardize formatting


melloware:
@mertsincan i know this is a big change but this is a huge improvement for the code base and it will enforce all the code looks the same.  Once merged I will add `npm run format-check` to the Actions build to make sure it doesn't pass the build if someone has not run `npm run format` before pushing their PR.

mertsincan:
@melloware no problem! Good job ;) 
Some Changes;
- I think 'jsxSingleQuote' can be false
- you can add `npm run format-check` to build-package command(before rollup -c)
- if-else-return statements are very complex. Exp; https://github.com/primefaces/primereact/pull/3173/files#diff-650bdcfa2f350584e949de0b660020ff47e9b52f3bb9b78646ce46ab006469fdR3

melloware:
OK I updated with `jsxSingleQuote: false` and `printWidth: 250`

mertsincan:
Nice! If it is OK for you, please merge it ;)

kalinkrustev:
There is an extra semicolon here https://github.com/primefaces/primereact/blob/master/components/lib/dropdown/Dropdown.js#L621
It ends up in the rendered markup.

melloware:
@kalinkrustev fixed!  Thanks for reporting.

PR #3180, opened 2022-08-24
Fix #3179: Dropdown onBlur check for inputRef

### LINKED ISSUE #3179: Dropdown: Unmounting focused dropdown throws TypeError
### Describe the bug

This is caused by #3164, where no check is made if inputRef.current has a value.

### Reproducer

https://codesandbox.io/s/primereact-test-forked-lyjh4k

### PrimeReact version

8.4.0

### React version

18.x

### Language

ALL

### Build / Runtime

Create React App (CRA)

### Browser(s)

_No response_

Open the CodeSandbox, click the Dropdown, so that it is focused and then click the button.
This will hide the Dropdown and cause the error.

### Expected behavior

Maybe in this case the onBlur should still be called with the value properties = undefined.

melloware:
###Defect Fixes
Fix #3179: Dropdown onBlur check for inputRef

### DIFF
--- a/components/lib/dropdown/Dropdown.js
+++ b/components/lib/dropdown/Dropdown.js
@@ -86,15 +86,16 @@ export const Dropdown = React.memo(
 
             if (props.onBlur) {
                 setTimeout(() => {
+                    const currentValue = inputRef.current ? inputRef.current.value : undefined;
                     props.onBlur({
                         originalEvent: event.originalEvent,
-                        value: inputRef.current.value,
+                        value: currentValue,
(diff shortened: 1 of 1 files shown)


kalinkrustev:
Yes, I was thinking about the same fix. Another option is to keep the current value in the state with React.useState and use it when calling the handler. I am not sure if the benefits of this are worth it, as it may also lead to other complications, such as extra re-rendering on each value change.

melloware:
Yeah I feel like this is the good solution for what we are trying to solve.

PR #3207, opened 2022-08-29
Fix #3200 Add Generics to Dropdown, Cascadeselect, Listbox, Multisele…

### LINKED ISSUE #3200: TypeScript: generic options support
Add generic typing to `SelectItem` type components:

- [x] cascadeselect.d.ts
- [x] dropdown.d.ts
- [x] listbox.d.ts
- [x] multiselect.d.ts
- [x] selectbutton.d.ts

melloware:
### Feature Requests
Fix #3200 Add Generics to Dropdown, Cascadeselect, Listbox, Multiselct, Selectbutton

cc @DerDorius I resubmitted your PR but added `value` overrides to the TS definitions.  I think we need to do more testing.

### DIFF
--- a/components/lib/cascadeselect/cascadeselect.d.ts
+++ b/components/lib/cascadeselect/cascadeselect.d.ts
@@ -2,31 +2,31 @@ import * as React from 'react';
 import { CSSTransitionProps } from '../csstransition';
 import { SelectItemOptionsType } from '../selectitem/selectitem';
 
-type CascadeSelectItemTemplateType = React.ReactNode | ((option: any) => React.ReactNode);
+type CascadeSelectItemTemplateType<TOption> = React.ReactNode | ((option: TOption) => React.ReactNode);
 
 type CascadeSelectAppendToType = 'self' | HTMLElement | undefined | null;
 
-interface CascadeSelectChangeParams {
+interface CascadeSelectChangeParams<TOption> {
     originalEvent: React.SyntheticEvent;
-    value: any;
+    value: TOption;
 }
 
-interface CascadeSelectGroupChangeParams extends CascadeSelectChangeParams {}
+interface CascadeSelectGroupChangeParams<TOption> extends CascadeSelectChangeParams<TOption> {}
 
-export interface CascadeSelectProps extends Omit<React.DetailedHTMLProps<React.InputHTMLAttributes<HTMLDivElement>, HTMLDivElement>, 'onChange' | 'ref'> {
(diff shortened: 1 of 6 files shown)


DerDorius:
@melloware 
There is still an issue with the `OnChange` Type, we haven't thought of. 
You can set the  `optionValue` prop and that means the value on the Change Event is not everytime of type Any

DerDorius:
And the `optionLabel` and the `optionValue` can also be Typed with this `NestedKeyOf` Type. That gives you generic Autocomplete for all the object keys
image
image

DerDorius:
Took me 2 Hours to figure this out, but if you want to have the `onChange`  `value` correctly typed you need 2 more generics and this complex conditional typing: 

`interface SelectButtonChangeParams<TOption, TValue, TMultiple> {
    value: TMultiple extends undefined
        ? TValue extends undefined
            ? TOption extends { value: any }
                ? TOption['value']
                : TOption
            : TValue extends keyof TOption
            ? TOption[TValue]
            : any
        : TValue extends undefined
        ? TOption extends { value: any }
            ? TOption['value'][]
            : TOption[]
        : TValue extends keyof TOption
        ? TOption[TValue][]
        : any[];
    stopPropagation(): void;
    preventDefault(): void;
    target: SelectButtonChangeTargetOptions<TOption>;
}`

So in Words:
- it first Checks if `multiple `Prop is set so it goes in the Array Branch or in the normal Branch.
- then it checks if the optionValue Prop is set with the TValue Generic
- if it is:
-      it checks if `TOption `contains a `value `field (because then Primereact returns only the `value `field) 
-           if it conntains a `value `field it returns the type of that `value `field
-           if it doesnt contain a `value `field it returns the whole `TObject`
- if it isnt:
-      it checks if the `optionValue `Prop is a Key of `TOption`
-            if it is it Returns the type of the `TValue `field
-            if it isnt it returns `any `(can happen when `TValue `is a nested String like: `"test.test"`)

Few of my Tests:

- [x] `multiple` is set, no `optionValue`, and no value in the Object

expect: TOption[]
image

- [x]  `multiple` is not set,  `optionValue `is test, 
expect: TOption[TValue]
image

- [x]  `multiple `is set,  `optionValue `is not set but there is a `value `property in the Object
expect TOption['value'][]
image

So in conclusion I would not merge this PR because its in correct, I will make a new PR in the near Future. 

PS: Sry for my bad English 😊

melloware:
OK closing my PR!

PR #3239, opened 2022-09-03
Fix #3238: Remove IE Support

### LINKED ISSUE #3238: Core: Drop Internet Explorer Support
image

melloware:
### Defect Fixes
Fix #3238: Remove IE Support


rafalwojdowski:
Is it possible to drop IE support in 10.0.0 version?

melloware:
That would be my preference @rafalwojdowski

melloware:
Merging this because there are already so many other features being used in the PrimeReact code based that only work on modern browsers so no point in keeping these IE hacks in places when IE is broken in many other places.

Issue #2842, opened 2022-05-05
TabView: Dynamically created TabPanels and onTabClose closes more than one Tab


maximilianweidenauer:
### Describe the bug

I have a TabView with dynamically created TabPanels, some are closable and some are disabled. When closing one of them, an additional TabPanel gets closed or isn't visible anymore. In my example there are 5 Tabs when closing the third, the forth won't be displayed aswell. (In my demo, the activeIndex stays 0 because in my app it shouldn't switch on close, when it isn't the currently selected one)

### Reproducer

https://codesandbox.io/s/misty-worker-1lu9yy?file=/src/demo/TabViewDemo.js

### PrimeReact version

7.2.1 (tested with 8.0.1 too)

### React version

17.x

### Language

TypeScript

### Build / Runtime

Create React App (CRA)

### Browser(s)

Chrome

### Steps to reproduce the behavior

1. Close the third Tab
2. Forth Tab will disappear aswell

### Expected behavior

When I close the third Tab, only the third Tab gets removed.

melloware:
Excellent reproducer. Definitely a bug

ulasturann:
The ".splice()" method refers the same array, so it might be more suitable to use the "filter" method as shown below:
```
                        onTabClose={(e) => {
                            const newTabArray = [...components];

                            newTabArray.filter((tab, i) => i !== e.index);
                            setComponents(newTabArray);
                        }}
```

maximilianweidenauer:
My reproducer didn't exactly do, what my react project does... In my project I'm sending a request to a server and the server tells me to remove this component. The server can remove a lot of components that's why I have an extra hook for that.
But I tried to kind of simulate what it would do and I have changed this in the reproducer:
```
          onTabClose={(e) => {
            const newTabArray = [];
            components.forEach((comp) => {
              if (comp.id !== e.index) {
                newTabArray.push(comp);
              }
            });
            setComponents(newTabArray);
          }}
```

Then I'm experiencing the same issues as before. Btw. in my app I'm not checking with `comp.id !== e.index` this is just for convenience in the reproducer, it would look more like this in the custom hook I mentioned.
```
            /** New Components when component changes */
            const newComponents = buildComponents();
            /** Contains the components */
            const cl = new Array<ReactElement>();
            newComponents.forEach(nc => {
                let alreadyAdded = false
                /** Checks if the new component is already added in the current components if yes add the old component else the new one */
                components.forEach(oc => {
[...]

PR #3264, opened 2022-09-08
Add closeMode to TabView

### LINKED ISSUE #2608: TabView does not remove all dynamic tabs programmatically
```
[x] bug report
[ ] feature request
```

https://codesandbox.io/s/rough-waterfall-6tnbv?file=/src/demo/TabViewDemo.js

**Current behavior**
When closing tabs that is not an active tab until the last tab (which is active) and then we try to close the active tab, the state in which we are setting those dynamic tabs to render is not getting empty. It always has one last tab in it. This issue was not there in 7.1.0

**Expected behavior**
When closing tabs and removing them from their state value programmatically, expect the state variable array to become empty after removing the last one.

Have provided the code in sandbox.

To reproduce the issue: Keep Tab 3 active, then close Tab 2, then Tab 1. Finally close Tab 1. In the console logs, I have given length of the array from which I am rendering the tabs. It remains as 1 instead of empty. If we close the tabs in order, then it works fine.
I've downgraded to 7.1.0 locally and couldn't find this issue.
I've also got another major issue but I couldn't reproduce it in the sandbox but that too disappears in 7.1.0. Issue is that if I try to close an inactive tab, then 2 tabs are disappearing from the screen although the tab state variable shows 2 tab objects in the array.

* **React version:**
17.0.2

* **PrimeReact version:**
7.2.0

Chrome

### LINKED ISSUE #4606: TabPanel: dynamic add/close weird behavior
### Describe the bug

Unless there is something I'm missing here, I want to have a static tab panel (or a button) with a list of items and when i click then the item should be displayed on a new tab.
This is OK but when i want to close the tab, then weird things happen.

Here is a sample

https://codesandbox.io/s/eloquent-roentgen-d9ss7y?file=/src/demo/TabViewDemo.js

When i click an item to be closed, then 2 tabs are closed.
If i remove the line which updates the list of tabs, it works fine, but this means that 

`          {tabPanels.map((tab) => {
            return (
              <TabPanel closable header={tab.header} key={tab.id}>
                {tab.body}
              </TabPanel>
            );
          })}`

this is ignored and only applied at first render.
How can we update the component based on the list that we provide in map function? Does tab panel depend on another internal list?

### Reproducer

https://codesandbox.io/s/eloquent-roentgen-d9ss7y?file=/src/demo/TabViewDemo.js

### PrimeReact version

9.6.0

### React version

18.x

### Language

TypeScript

### Build / Runtime

Vite

### Browser(s)

Firefox 114

Just click the Tab 1 to be removed. Tab 2 is also removed.

### Expected behavior

Only 1 tab at a time should be closed. Removing line
`setTabPanels(tabPanels.filter((p, i) => i !== e.index - 1));` makes things work but the point is to depend on the dynamically changed list of tabs.

diogoko:
This new property fixes the problem described in #2842.

Fix #4606 
Fix #2608 

# Context

Currently, it's not possible to dynamically create closable TabPanels in TabView because the default behavior of the tab header close button is to hide the tab by index. More precisely, it adds the tab index to the hiddenTabsState array. If the onTabClose is also handled and programmatically removes the tab from the list used to generate the children TabPanel components, then 2 tabs are removed at once (the one that was closed and the one immediately following it).

This is an example of the kind of code that causes this bug:

```js
<TabView onTabClose={(e) => setDynamicTabs([...dynamicTabs.slice(0, e.index), ...dynamicTabs.slice(e.index + 1)])}>
    {dynamicTabs.map((tab) => {
        return (
            <TabPanel key={tab.id} header={tab.name} closable>
                <p>Name: {tab.name}</p>
            </TabPanel>
        );
    })}
</TabView>
```

# Description of changes

This PR adds a new closeMode property to TabView. Its value is "auto" by default; when "manual", clicking the default close button of a closable TabPanel will not add the index to the hiddenTabsState array and it's the programmer's responsibility to update the list used to generate the dynamic TabPanel components.

This is the above example with the new property:

```js
<TabView closeMode="manual" onTabClose={(e) => setDynamicTabs([...dynamicTabs.slice(0, e.index), ...dynamicTabs.slice(e.index + 1)])}>
    {dynamicTabs.map((tab) => {
        return (
            <TabPanel key={tab.id} header={tab.name} closable>
                <p>Name: {tab.name}</p>
            </TabPanel>
        );
    })}
</TabView>
```

I also updated the docs and TS definitions.

### DIFF
--- a/api-generator/components/tabview.js
+++ b/api-generator/components/tabview.js
@@ -34,6 +34,12 @@ const TabViewProps = [
         type: 'boolean',
         default: 'false',
         description: 'When enabled displays buttons at each side of the tab headers to scroll the tab list.'
+    },
+    {
+        name: 'closeMode',
+        type: 'string',
+        default: 'auto',
+        description: 'Use "manual" to programmatically control dynamic tabs visibility when the close button of a closable tab is clicked.'
     }
 ];
 
--- a/components/doc/common/apidoc/index.json
+++ b/components/doc/common/apidoc/index.json
@@ -36820,6 +36820,14 @@
                             "description": "Defines if tab can be removed."
                         },
                         {
+                            "name": "closeMode",
+                            "optional": true,
+                            "readonly": false,
+                            "type": "\"auto\" | \"manual\"",
+                            "default": "auto",
+                            "description": "How closable tabs are managed."
+                        },
+                        {
                             "name": "style",
                             "optional": true,
                             "readonly": false,
--- a/components/doc/tabview/dynamicdoc.js
+++ b/components/doc/tabview/dynamicdoc.js
@@ -1,0 +1,111 @@
+import { TabView, TabPanel } from '../../lib/tabview/TabView';
+import { Button } from '../../lib/button/Button';
+import { DocSectionCode } from '../common/docsectioncode';
+import { DocSectionText } from '../common/docsectiontext';
+import { useState } from 'react';
+
+const createClient = () => {
+    const id = Math.floor(Math.random() * 1000000);
+
+    return { id, name: `Client ${id}` };
+};
+
+export function DynamicDoc(props) {
+    const code = {
+        basic: `
+<TabView
+    closeMode="manual"
+    onTabClose={(e) => setDynamicTabs([...dynamicTabs.slice(0, e.index), ...dynamicTabs.slice(e.index + 1)])}
+>
+    {dynamicTabs.map((tab) => (
+        <TabPanel key={tab.id} header={tab.name} closable>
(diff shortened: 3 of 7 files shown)


melloware:
Thanks for the PR!  I assigned Mert to review but looks good to me.

melloware:
Can you rebase this PR there are conflicts.

diogoko:
@melloware I've rebased it from the current `master` and redone the docs in the 9.x format.

melloware:
I actually think this is fixed by https://github.com/primefaces/primereact/issues/5229

PR #3271, opened 2022-09-09
Fixes #3270: Accordion return null if idState is not set.

### LINKED ISSUE #3270: Accordion: No id causes remounting of AccordionTabs.
### Describe the bug

AccordionTab keys are tied to idState variable, which starts off null if no prop is passed in. This variable is then set with a useEffect (to appease SSR). Unfortunately this then causes the keys to be different, which in turn causes a remount. This is an issue at it causes useEffects to be triggered excessively. 

### Reproducer

https://codesandbox.io/s/epic-firefly-5yil2r

### PrimeReact version

8.5.0

### React version

18.x

### Language

ALL

### Build / Runtime

Create React App (CRA)

### Browser(s)

_No response_

1. In Sandbox, see that have 1 unmount and 2 mounts.
2. Insert an 'id' for Accordion and see 0 unmounts and 1 mount.

### Expected behavior

Regardless of id being passed in or not, the AccordianTabs should not be remounted.

harrybradshaw:
If the id prop is not set, then idState is null, which is then resolved by the useMountEffect. However this null value has been used to key AccordionTabs, which then get remounted.

This change doesn't generate any AccordionTabs if the idState is not set, resulting in only one mounting of the children.

### DIFF
--- a/components/lib/accordion/Accordion.js
+++ b/components/lib/accordion/Accordion.js
@@ -5,13 +5,13 @@ import { classNames, IconUtils, ObjectUtils, UniqueComponentId } from '../utils/
 
 export const AccordionTab = () => {};
 
+const shouldUseTab = (tab) => tab && tab.props.__TYPE === 'AccordionTab';
+
 export const Accordion = React.forwardRef((props, ref) => {
     const [idState, setIdState] = React.useState(props.id);
     const [activeIndexState, setActiveIndexState] = React.useState(props.activeIndex);
     const elementRef = React.useRef(null);
     const activeIndex = props.onTabChange ? props.activeIndex : activeIndexState;
-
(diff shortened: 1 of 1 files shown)


components/lib/accordion/Accordion.js
```diff
@@ -5,14 +5,14 @@ import { classNames, IconUtils, ObjectUtils, UniqueComponentId } from '../utils/
 
 export const AccordionTab = () => {};
 
+const shouldUseTab = (tab) => tab && tab.props.__TYPE === 'AccordionTab';
```
harrybradshaw, line 8: no need for this to be inside the component

harrybradshaw:
Another solution, if you didn't want to take the above approach, would be to tie the AccordionTab keys to a different prop.
`const keyBase = React.useRef(props.id || UniqueComponentId());`

melloware (approved):

PR #3322, opened 2022-09-14
Fix #3321: Pass "data" properties to input components

### LINKED ISSUE #3321: InputNumber: missing data attribute
### Describe the bug

InputNumber component does not properly pass down the `data` attribute as expected per #2620 

```
<span id="76" class="p-inputnumber p-component p-inputwrapper p-invalid" data-id="76">
    <input role="spinbutton" class="p-inputtext p-component p-filled p-inputnumber-input" type="text" inputmode="numeric" name="76" value="" aria-valuenow="100">
</span>
```

The `data` attribute is on the parent `span` element instead.

### Reproducer

_No response_

### PrimeReact version

8.5.0

### React version

17.x

### Language

ES6

### Build / Runtime

Next.js

### Browser(s)

_No response_

_No response_

### Expected behavior

Expected is that the `data` attribute is on the input element so that it can be easily accessed via the change event. Other Prime Input elements behave in this fashion.

melloware:
### Defect Fixes
Fix #3321: Pass "data" properties to input components

- [x] AutoComplete
- [x] Dropdown
- [x] InputNumber
- [x] InputSwitch
- [x] MultiSelect
- [x] RadioButton
- [x] Checkbox


kalinkrustev:
There is an issue with this change as it also moves the data-testid property to the hidden elements, where it is not useful for automated tests. Will you consider leaving data-testid on the parent element as it was before?

melloware:
Hmmm but don't you want your data-testid to get the actual input component and not the wrapper DIV?

melloware:
even if they are `hidden-accessible` ?  So Does #3321 make no sense then why did the user report it?

kalinkrustev:
You can see that event handlers are actually attached to the wrapper elements, not the hidden one.

kalinkrustev:
> even if they are `hidden-accessible` ? So Does #3321 make no sense then why did the user report it?

The report did not mention data-testid, maybe it is a different use case.

melloware:
Got it.  Fixing it now.  It really only applies to InputNumber because it is NOT a hidden input.

kalinkrustev:
OK, I can take a look at your fix to confirm we understand each other correctly.

melloware:
yep just fixed.  InputNumber is the only one that forwards to its InputText component

PR #3412, opened 2022-10-02
Fix #3388: CellSelection with metakey row index

### LINKED ISSUE #3388: Datatable: onSelectionChange event returns wrong rowIndex with shiftKey selection
### Describe the bug

I noticed that the onSelectionChange event of the Datatable component returns values with a bad 'rowIndex' attribute in some cases.

For example : a Datatable where multiple cell selection with MetaKey is activated and we have a pagination.
```
<Datatable
...
selectionMode="multiple"
cellSelection
metaKeySelection={true}
...
paginatorTemplate="CurrentPageReport FirstPageLink PrevPageLink PageLinks NextPageLink LastPageLink"
rows={5}
rowsPerPageOptions={[5, 10, 50]}
...
>
...
</Datatable>
```

In that case, if we go to the 2nd page and looked at the values return by the onSelectionChange event `onSelectionChange={(e) => console.log(e.value)}`, we will see that :
- with a simple click or with Ctrl+click (ctrlKey), the returned values have a correct rowIndex which match with the pagination. (in the example above, a click on the first element of the 2nd page will have rowIndex=5)
- with Shift+click, the returned values have a rowIndex which doesn't correspond with the pagination. (in the example above, a shift+click on the first element of the 2nd page will have rowIndex=0)

### Reproducer

https://codesandbox.io/embed/nervous-wind-g8lygm?fontsize=14&hidenavigation=1&theme=dark

### PrimeReact version

8.6.0

### React version

18.x

### Language

TypeScript

### Build / Runtime

Create React App (CRA)

### Browser(s)

_No response_

_No response_

### Expected behavior

A Shift+click selection should return values with a rowIndex matching the pagination.

melloware:
### Defect Fixes
Fix #3388: CellSelection with metakey row index

### DIFF
--- a/components/lib/datatable/TableBody.js
+++ b/components/lib/datatable/TableBody.js
@@ -280,7 +280,14 @@ export const TableBody = React.memo(
             DomHandler.clearSelection();
             rangeRowIndex.current = allowCellSelection() ? event.rowIndex : event.index;
             let selectionInRange = selectRange(event);
-            let selection = isMultipleSelection() ? [...new Set([...(props.selection || []), ...selectionInRange])] : selectionInRange;
+            let selection = selectionInRange;
+
+            if (isMultipleSelection()) {
+                const currentSelection = new Set([...(props.selection || [])]);
+                const newSelection = [...selectionInRange].filter((x) => !currentSelection.has(x));
+
+                selection = [...new Set([...newSelection])];
+            }
 
             if (props.onSelectionChange && selection !== props.selection) {
                 props.onSelectionChange({
@@ -357,6 +364,7 @@ export const TableBody = React.memo(
             for (let i = rowRangeStart; i <= rowRangeEnd; i++) {
                 let rowData = value[i];
                 let columns = props.columns;
+                let rowIndex = props.paginator ? i + props.first : i;
 
                 for (let j = cellRangeStart; j <= cellRangeEnd; j++) {
                     let field = columns[j].props.field;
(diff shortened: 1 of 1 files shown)


components/lib/datatable/TableBody.js
```diff
@@ -280,7 +280,14 @@ export const TableBody = React.memo(
             DomHandler.clearSelection();
             rangeRowIndex.current = allowCellSelection() ? event.rowIndex : event.index;
             let selectionInRange = selectRange(event);
-            let selection = isMultipleSelection() ? [...new Set([...(props.selection || []), ...selectionInRange])] : selectionInRange;
```
mertsincan, line 283: @melloware I didn't understand this change. 'new Set' prevents duplicate items.

melloware: If you revert it and look you will see it's allowing duplicate selections. Unlike Java in Ecmascript two identical objects are not  identical so it was allowing duplicate cells. 

See: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set

Look at the examples you will see adding two objects even if they have identical attributes will add it twice.

melloware: Tomorrow I can send you before and after screenshots.

mertsincan: In this part the rowData references are the same so duplication can be avoided. Exp;
```
var a = [{a: 1}, {b: 2}];
var b = [a[0]];

new Set([...a, ...b])
output: Set(2) {{a: 1}, {b: 2}}
```
Is there an issue or example reported about this?

melloware: We can change it to just `let selection = selectRange(event);` and get rid of the Sets entirely

PR #3482, opened 2022-10-17
Fix #3483: Splitter: Non optimal handling of splitter sizing

### LINKED ISSUE #3413: Splitter state is not saved in StrictMode
### Describe the bug

Splitter sometimes overwrites local storage data.
If using  <React.StrictMode> the data gets reset on refresh.
If removing StrictMode, then it works as intended.

### Reproducer

https://codesandbox.io/s/primereact-test-forked-p647y5?file=/src/index.js

### PrimeReact version

8.6.1

### React version

18.x

### Language

TypeScript

### Build / Runtime

Create React App (CRA)

### Browser(s)

_No response_

As shown in the codesandbox, if you use StrictMode then stored data gets overwritten on refresh.
If you remove the StrictMode tags, then you can refresh and it correctly remembers you data.

### Expected behavior

Don't overwrite data regardless of whether or not StrictMode is used.

### LINKED ISSUE #3483: Splitter: Non optimal handling of splitter sizing
### Describe the bug

There are actually 3 issues in the original code of version 8.6.1
- state is saved in storage without ever moving the splitter
- sizes are lost on re-render
- if the splitter is moved too far, sizes bigger than 100% and smaller than 0% set and even saved

### Reproducer

https://codesandbox.io/s/primereact-test-forked-wu3cyg?file=/src/index.js

### PrimeReact version

8.6.1

### React version

18.x

### Language

ALL

### Build / Runtime

Create React App (CRA)

### Browser(s)

_No response_

1) Open the reproducer URL
2) Click the button before any resizing - observe sizes are saved in local storage
3) Resize the splitter, then click the button - sizes are lost (also observed in local storage)
4) Move the splitter to the left beyond the edge - invalid sizes are set (also observed in local storage)

### Expected behavior

- Sizes should not be persisted unless resizing is performed
- Changing the `stateKey` or `stateStorage` properties on already mounted splitter should trigger reloading the sizes
- Sizes below 0% or above 100% should not be set

kalinkrustev:
Fix #3483
Fix #3413

### DIFF
--- a/components/lib/splitter/Splitter.js
+++ b/components/lib/splitter/Splitter.js
@@ -15,10 +15,15 @@ export const Splitter = React.memo(
         const prevPanelElement = React.useRef(null);
         const nextPanelElement = React.useRef(null);
         const prevPanelSize = React.useRef(null);
+        const prevPanelSizeNew = React.useRef(null);
         const nextPanelSize = React.useRef(null);
+        const nextPanelSizeNew = React.useRef(null);
         const prevPanelIndex = React.useRef(null);
-        const panelSizes = React.useRef(null);
+        const [panelSizes, setPanelSizes] = React.useState([]);
         const isStateful = props.stateKey != null;
+        const childrenLength = props.children && props.children.length;
+
+        const panelSize = (sizes, index) => (index in sizes) ? sizes[index] : (props.children[index].props.size || 100 / childrenLength);
 
         const [bindDocumentMouseMoveListener, unbindDocumentMouseMoveListener] = useEventListener({ type: 'mousemove', listener: (event) => onResize(event) });
         const [bindDocumentMouseUpListener, unbindDocumentMouseUpListener] = useEventListener({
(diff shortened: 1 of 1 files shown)


melloware:
this looks good but it looks like a build error

kalinkrustev:
Shall I also open a new issue? Linking to a closed issue does not seem to work.

melloware:
i will re-open it since its for 8.6.2 it has not gone out yet.

kalinkrustev:
I opened a new issue, as I have found additional problems with the splitter.

components/lib/splitter/Splitter.js
```diff
@@ -15,11 +15,15 @@ export const Splitter = React.memo(
...
         const prevPanelSize = React.useRef(null);
+        const prevPanelSizeNew = React.useRef(null);
         const nextPanelSize = React.useRef(null);
+        const nextPanelSizeNew = React.useRef(null);
         const prevPanelIndex = React.useRef(null);
-        const panelSizes = React.useRef(null);
-        const mounted = React.useRef(false);
+        const [panelSizes, setPanelSizes] = React.useState([]);
         const isStateful = props.stateKey != null;
+        const childrenLength = props.children && props.children.length;
```
melloware, line 24: Hmmm won't childrenLength be a `boolean` here and not an integer?  Shouldn't it be:

```js
const childrenLength = props.children ? props.children.length : 0;
```

won't dividing by 0 below return `Infinity` ?

melloware (changes requested):

components/lib/splitter/Splitter.js
kalinkrustev: It is more of a safeguard against `Cannot read properties of undefined`. I do not see a risk of division by zero, because panelSize function will never be invoked if there are zero children.

melloware (approved):

melloware:
@kalinkrustev can you reivew my PR here if you are still around: https://github.com/primefaces/primereact/issues/5916

I have confirmed that with your fix to `setPanelSize` the values are always NULL.?

PR #3535, opened 2022-10-31
Fix #3534: Splitter: widths provided to SplitterPanel.size are not correctly applied

### LINKED ISSUE #3534: Splitter: widths provided to SplitterPanel.size are not correctly applied
### Describe the bug

Widths supplied to the size property are divided by the number of SplitterPanels.

### Reproducer

https://codesandbox.io/s/primereact-test-forked-lyx7jp?file=/src/index.js

### PrimeReact version

8.7.0

### React version

17.x

### Language

ALL

### Build / Runtime

Create React App (CRA)

### Browser(s)

_No response_

Open the reproducer URL and inspect the spitpanel widths, where they will wrongly be set to
flex-basis: calc(5% - 4px);
flex-basis: calc(45% - 4px);
insted of
flex-basis: calc(10% - 4px);
flex-basis: calc(90% - 4px);

### Expected behavior

Correctly apply the provided widths.

kalinkrustev:
Fix: #3534

### DIFF
--- a/components/lib/splitter/Splitter.js
+++ b/components/lib/splitter/Splitter.js
@@ -23,7 +23,7 @@ export const Splitter = React.memo(
         const isStateful = props.stateKey != null;
         const childrenLength = props.children && props.children.length;
 
-        const panelSize = (sizes, index) => (index in sizes ? sizes[index] : (props.children[index].props.size || 100) / childrenLength);
+        const panelSize = (sizes, index) => (index in sizes ? sizes[index] : (props.children[index].props.size || 100 / childrenLength));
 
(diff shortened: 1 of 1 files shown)


melloware:
Looks like you need to run prettier format

kalinkrustev:
Yes and I think the issue was caused when I was trying to satisfy the linter. I think I got a similar error in my previous PR.

kalinkrustev:
I admit I do not have memorized the full 18 levels of JS operator precedence and when Prettier insists to remove the extra brackets, it might look "prettier", but then requires the reader (which includes me) of the source code to also memorize these 18 levels.

melloware:
You can also add `// prettier-ignore` if you don't want it touched.

kalinkrustev:
OK, I will know for the next time. But then it looks suspicious :)

melloware:
HA true!

PR #3549, opened 2022-11-01
Fix #3546: Splitter error with only 1 panel

### LINKED ISSUE #3546: Splitter: Error with only 1 panel
In PrimeReact 8.7.1, the default Splitter + SplitterPanel is throwing a "Cannot read properties of undefined (reading 'props')" error:

```js
splitter.esm.js:142 Uncaught TypeError: Cannot read properties of undefined (reading 'props')
    at panelSize (splitter.esm.js:142:1)
    at createPanel (splitter.esm.js:362:1)
    at react.development.js:1158:1
    at mapIntoArray (react.development.js:1049:1)
    at Object.mapChildren [as map] (react.development.js:1157:1)
    at createPanels (splitter.esm.js:373:1)
    at splitter.esm.js:378:1
    at renderWithHooks (react-dom.development.js:16305:1)
    at updateForwardRef (react-dom.development.js:19226:1)
    at beginWork (react-dom.development.js:21636:1)
 ```
splitter-error

 In splitter.esm.js:
```js
  var childrenLength = props.children && props.children.length;

  var panelSize = function panelSize(sizes, index) {
    return index in sizes ? sizes[index] : props.children[index].props.size || 100 / childrenLength;
  };
  ```

childrenLength is undefined, but props.children is this:
props-children

melloware:
### Defect Fixes
Fix #3546: Splitter error with only 1 panel

### DIFF
--- a/components/lib/splitter/Splitter.js
+++ b/components/lib/splitter/Splitter.js
@@ -22,8 +22,7 @@ export const Splitter = React.memo(
         const [panelSizes, setPanelSizes] = React.useState([]);
         const isStateful = props.stateKey != null;
         const childrenLength = props.children && props.children.length;
-
-        const panelSize = (sizes, index) => (index in sizes ? sizes[index] : props.children[index].props.size || 100 / childrenLength);
+        const panelSize = (sizes, index) => (index in sizes ? sizes[index] : ((props.children[index] && props.children[index].props.size) || 100) / childrenLength);
 
         const [bindDocumentMouseMoveListener, unbindDocumentMouseMoveListener] = useEventListener({ type: 'mousemove', listener: (event) => onResize(event) });
         const [bindDocumentMouseUpListener, unbindDocumentMouseUpListener] = useEventListener({


kalinkrustev:
Sorry, I did not know children could be non array. But I also wonder if specifying a size for a single panel makes sense, in which case this fix will not apply the size and instead use 100%

melloware:
Sure I was not expecting someone to use 1 panel.  You want to test and submit another PR?

kalinkrustev:
I do not know yet. I will try to check if it makes sense to have one panel width first.

kalinkrustev:
Unfortunately this fix, undid what was fixed by #3535

kalinkrustev:
I think the next fix should also include a snapshot test, so future errors can be limited.

melloware:
I have snapshot tests in place now

kalinkrustev:
Yes, but the snapshots contained incorrect values.
Check here #3646 for the bugfix including support for specifying the size of a single panel and also a few more snapshots.

PR #3561, opened 2022-11-02
Fix #3559: Component: Calendar -> keepInvalid no longer works

### LINKED ISSUE #3559: Component: Calendar
### Describe the bug

keepInvalid={true} property of Calendar no longer works

### Reproducer

https://stackblitz.com/edit/vitejs-vite-avhcdw?file=src/App.tsx

### PrimeReact version

8.7.1

### React version

17.x

### Language

ALL

### Build / Runtime

Create React App (CRA)

### Browser(s)

All

Focus calendar, input anything from keyboard

### Expected behavior

Populate inputtext from calendar with keyboard values

cerbugabriel:
Defect fix for #3559.

### DIFF
--- a/components/lib/calendar/Calendar.js
+++ b/components/lib/calendar/Calendar.js
@@ -118,7 +118,7 @@ export const Calendar = React.memo(
                 }
             } catch (err) {
                 //invalid date
-                const value = props.keepInvalid ? event.value : null;
+                const value = props.keepInvalid ? rawValue : null;
 
                 updateModel(event, value);
             }
@@ -1961,7 +1961,7 @@ export const Calendar = React.memo(
 
             let formattedValue = '';
 
-            if (value) {
+            if (value instanceof Date) {
                 try {
                     if (isSingleSelection()) {
                         formattedValue = isValidDate(value) ? formatDateTime(value) : '';
@@ -1991,6 +1991,8 @@ export const Calendar = React.memo(
                 } catch (err) {
                     formattedValue = value;
                 }
+            } else if (props.keepInvalid) {
+                formattedValue = value;
             }
 
             inputRef.current.value = formattedValue;
@@ -2511,7 +2513,7 @@ export const Calendar = React.memo(
                         }
                     }
 
-                    if (viewDate) {
+                    if (viewDate instanceof Date) {
                         validateDate(viewDate);
                         setViewDateState(viewDate);
                         setCurrentMonth(viewDate.getMonth());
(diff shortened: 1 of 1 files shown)


components/lib/calendar/Calendar.js
```diff
@@ -118,7 +118,7 @@ export const Calendar = React.memo(
                 }
             } catch (err) {
                 //invalid date
-                const value = props.keepInvalid ? rawValue : null;
+                const value = props.keepInvalid ? event.value : null;
```
melloware, line 121: I actually think the line need to be this...

```js
const value = props.keepInvalid ? null : rawValue;
```

if you try your with `keepInvalid` it works but if you remove `keepInvalid` its allowing invalid characters

melloware (changes requested):

components/lib/calendar/Calendar.js
cerbugabriel: The difference between keepInvalid true and false should only be on onBlur. 
You can input invalid characters with it true and false, but, when inBlur state value remains on keepInvalid = true, and on keepInvalid = false makes the state value null.

melloware: Just try your fix above with both `keepInvalid={true}` and `keepInvalid={false}` you will see your fix does NOT work.  Then try mine you will see it does work.

cerbugabriel: I did a couple of digging and found out why it didn't update. updateInputField didn't work properly.

components/lib/calendar/Calendar.js
```diff
@@ -1961,7 +1961,7 @@ export const Calendar = React.memo(
 
             let formattedValue = '';
 
-            if (value) {
+            if (value instanceof Date) {
```
melloware, line 1964: I don't think you can do this here because `issingleSelection` its a date but `isRangeSelection` and `isMultipleSelection` the value is not a DATE its an array of dates.

melloware (changes requested):

cerbugabriel (commented):
Move condition for keepInvalid inside isSingleSelection.
(should only work for isSingleSelection)

components/lib/calendar/Calendar.js
cerbugabriel: You are right

cerbugabriel (commented):
move props.keepInvalid to isSingleSelection

melloware (approved):

melloware:
Nice job!

PR #3597, opened 2022-11-09
improve snapshot tests

kalinkrustev:
I will put comments in the `Files changed` tab about the ideas of the changes.

### DIFF
--- a/components/lib/badge/Badge.spec.js
+++ b/components/lib/badge/Badge.spec.js
@@ -2,19 +2,17 @@ import '@testing-library/jest-dom';
 import { render } from '@testing-library/react';
 import { Badge } from './Badge';
 
-const snapshot = (props, name) => expect(render(<Badge {...props} />).container).toMatchSnapshot(name);
+import { snapshot } from '../../test';
 
 describe('Badge', () => {
-    test('check snapshots', () => {
-        snapshot({}, 'default');
-        snapshot({ value: 22 }, 'value');
-        snapshot({ size: 'large' }, 'size large');
-        snapshot({ size: 'xlarge' }, 'size xlarge');
-        snapshot({ size: 'invalid' }, 'size invalid');
-        snapshot({ severity: 'success' }, 'severity success');
-        snapshot({ severity: 'info' }, 'severity info');
-        snapshot({ severity: 'warning' }, 'severity warning');
-        snapshot({ severity: 'danger' }, 'severity danger');
-        snapshot({ severity: 'invalid' }, 'severity invalid');
-    });
+    snapshot(<Badge />, 'default');
+    snapshot(<Badge value={22} />, 'value');
+    snapshot(<Badge size="large" />, 'size large');
+    snapshot(<Badge size="xlarge" />, 'size xlarge');
+    snapshot(<Badge size="invalid" />, 'size invalid');
+    snapshot(<Badge severity="success" />, 'severity success');
+    snapshot(<Badge severity="info" />, 'severity info');
+    snapshot(<Badge severity="warning" />, 'severity warning');
+    snapshot(<Badge severity="danger" />, 'severity danger');
+    snapshot(<Badge severity="invalid" />, 'severity invalid');
 });
--- a/components/lib/badge/__snapshots__/Badge.spec.js.snap
+++ b/components/lib/badge/__snapshots__/Badge.spec.js.snap
@@ -1,6 +1,6 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
-exports[`Badge check snapshots: default 1`] = `
+exports[`Badge default 1`] = `
 <div>
   <span
     class="p-badge p-component p-badge-dot"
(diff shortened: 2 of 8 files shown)


components/lib/badge/Badge.spec.js
```diff
@@ -2,19 +2,17 @@ import '@testing-library/jest-dom';
...
-        snapshot({ size: 'large' }, 'size large');
-        snapshot({ size: 'xlarge' }, 'size xlarge');
-        snapshot({ size: 'invalid' }, 'size invalid');
-        snapshot({ severity: 'success' }, 'severity success');
-        snapshot({ severity: 'info' }, 'severity info');
-        snapshot({ severity: 'warning' }, 'severity warning');
-        snapshot({ severity: 'danger' }, 'severity danger');
-        snapshot({ severity: 'invalid' }, 'severity invalid');
-    });
+    snapshot(<Badge />, 'default');
```
kalinkrustev, line 8: To make the snapshot function easier to use across components, it expects the element, not the properties only.

components/lib/badge/__snapshots__/Badge.spec.js.snap
```diff
@@ -1,78 +1,78 @@
 // Jest Snapshot v1, https://goo.gl/fbAQLP
 
-exports[`Badge check snapshots: default 1`] = `
+exports[`Badge default 1`] = `
```
kalinkrustev, line 3: Here the 'check snapshots: ' string was not really useful, as it is clear that snapshots are being checked.

components/lib/button/Button.spec.js
```diff
@@ -2,103 +2,21 @@ import '@testing-library/jest-dom';
...
+    snapshot(<Button label={'test'} iconPos={'bottom'} visible={true} />, 'when iconPos is bottom Button is vertical');
+    snapshot(<Button visible={true} />, 'when label is empty it returns empty button');
+    snapshot(<Button badge={'test'} />, 'when badge is true it renders Button with badge');
+    snapshot(<Button />, 'when badge is null it renders Button without badge');
+    snapshot(<Button loading={'test'} />, 'when click the button if loading is true it renders Button with loading icon');
+    snapshot(<Button />, 'when click the button if loading is false it renders Button without loading icon');
+    snapshot(<Button />, 'when label is true it renders Button with default aria label');
+    snapshot(<Button label={'test'} />, 'when aria-label prop is not exist aria-label prop should be equal to label prop ');
+    snapshot(<Button aria-label={'test'} />, 'when label prop is not exist label prop should be equal to aria-label prop');
+    snapshot(<Button label={'test'} badge={'lost'} />, 'when using badge and label the aria-label should contain both values');
```
kalinkrustev, line 19: Here, the snapshot calls replaced the old code, that was repetitive. It is now more clearly visible, that some tests are doing the same thing and that test names might be a bit incorrect. Three of them are rendering just `<Button />`.

components/lib/chip/Chip.spec.js
```diff
@@ -2,40 +2,42 @@ import '@testing-library/jest-dom';
...
+    snapshot(<Chip image="http://google.com/my.png" />, 'image');
+    snapshot(<Chip image="http://google.com/my.png" imageAlt="jest" />, 'image and alt');
+    snapshot(<Chip icon="pi pi-check" />, 'icon');
+    snapshot(<Chip label="jest" />, 'label');
     test('when removable is true it returns with remove icon', () => {
         // Arrange
         const removeOn = jest.fn();
         const { container } = render(<Chip removable onRemove={removeOn} />);
+
+        expect(container).toMatchSnapshot('before remove');
```
kalinkrustev, line 18: Here is example how to pass a name to the `toMatchSnapshot` call, so that two snapshots are made in the same test. Here is makes sense to snapshot before and after the action, as the component rendered differently.

components/lib/chip/__snapshots__/Chip.spec.js.snap
```diff
@@ -59,3 +59,33 @@ exports[`Chip check snapshots: label 1`] = `
...
+  <div
+    class="p-chip p-component"
+  >
+    <span
+      class="p-chip-remove-icon pi pi-times-circle"
+      tabindex="0"
+    />
+  </div>
+</div>
+`;
```
kalinkrustev, line 76: You can observe the above how the `before remove` and `after remove` snapshots rendered different results (as expected)

components/lib/inputtext/InputText.spec.js
```diff
@@ -27,7 +27,6 @@ describe('InputText', () => {
 
         // Act
         expect(input).toBeEnabled();
-        expect(input).not.toHaveClass('p-filled');
```
kalinkrustev, line 30: This is not needed, as the snapshot will cover this check.

components/test/index.js
```diff
@@ -0,0 +1,8 @@
+import '@testing-library/jest-dom';
+import { render } from '@testing-library/react';
+
+export function snapshot(element, name) {
+    test(name, () => {
+        expect(render(element).container).toMatchSnapshot();
+    });
+}
```
kalinkrustev, line 8: The reusable function, to avoid repetitive calls to `test`, `expect`, `render` and `toMatchSnapshot`

components/lib/inputtext/InputText.spec.js
melloware: I left this here because it was explicitly what i was testing not meant to be implicitly figured out from the SNAPSHOT.

melloware (approved):
Actually I am OK with this.

PR #3598, opened 2022-11-09
pages snapshot POC

kalinkrustev:
Here is an idea: as the pages folder already contains a lot of code that renders components using different combination of properties, it seems snapshot tests can be relatively easy to activate for the majority of the components.

I tried this only for the `Button` component, but it seems many already follow the pattern, where they are rendered in a div with the classes `content-section implementation`, so the snapshot can only capture this div:
```js
expect(container.getElementsByClassName('content-section implementation')).toMatchSnapshot();
```
The change in `components/doc/common/codehighlight.js` is required, because in snapshot tests there is no window.Prism.

As I do not know how next.js works, I am not sure that the .spec.js files can be put in the `page/xxx` folders. This also causes the respective __snapshots__ folder to appear there.

Anyway, I think this is an interesting idea, that might help very easily to achieve a lot of test coverage.
Note that  the file indes.spec.js can be just copied to many of the `page/xxx` folders and it is likely to work.

### DIFF
--- a/components/doc/common/codehighlight.js
+++ b/components/doc/common/codehighlight.js
@@ -5,7 +5,7 @@ export function CodeHighlight(props) {
     const languageClassName = `language-${props.lang || 'jsx'}`;
 
     useEffect(() => {
-        window.Prism.highlightElement(codeElement.current);
+        window.Prism?.highlightElement(codeElement.current);
     }, []);
 
     return (
--- a/pages/button/__snapshots__/index.spec.js.snap
+++ b/pages/button/__snapshots__/index.spec.js.snap
@@ -1,0 +1,1147 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`Button Page content snapshot 1`] = `
+HTMLCollection [
+  <div
+    class="content-section implementation button-demo"
+  >
+    <div
+      class="card"
+    >
+      <h5>
+        Basic
+      </h5>
+      <button
+        aria-label="Submit"
+        class="p-button p-component"
+      >
+        <span
+          class="p-button-label p-c"
+        >
+          Submit
+        </span>
+      </button>
+      <button
+        aria-label="Disabled"
+        class="p-button p-component p-disabled"
+        disabled=""
+      >
+        <span
+          class="p-button-label p-c"
+        >
+          Disabled
+        </span>
+      </button>
+      <button
+        aria-label="Link"
+        class="p-button p-component p-button-link"
+      >
+        <span
+          class="p-button-label p-c"
+        >
+          Link
+        </span>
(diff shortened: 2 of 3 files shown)


kalinkrustev:
When changing the demo, it is easy to change the snapshot too. The tricky part is if you change the demo and the component together.

melloware:
I would rather Cypress or PlayWright integration test the working showcase demo with real browser than do SNAPSHOT testing against it.  I really think the SNAPSHOTs should be constrained to unit tests of the component.  This feels like it will add to much overhead on every demo change

melloware:
actually for integration tests I think we should use Playwright: https://playwright.dev/

I got a demo from a friend and its really slick and the tests look like normal Jest tests

kalinkrustev:
Yes, Playwright is good, I like it and I use it, but it has its own set of requirements for both developer and CI. Also using it is not a reason to not use Jest. I do not think the case of testing the pages can be defined as either pure unit or pure integration test and be a reason to choose the tool (or constrain to a single tool). Using Playwright will not result in significantly less effort for maintaining the tests. I think the full benefit of tests is when they test both the visual and markup result. I think Jest is easier to use for markup and Playwright for visual. This PR was more about discussing if testing via pages is something the team behind Primereact considers useful. The proposed approach is easy to start with. Maybe it is not the perfect one or the only one, but it will give a very good code coverage without a lot of effort. Therefore, it is not a lot of investment to use it and can be easily abandoned if it proves to be unsustainable. Updating the snapshots is easy to do and it is not mandatory to catch all the bugs always. If it is too much effort to analyze all the differences in all the snapshots, it is OK to not do it fully and just accept the new snapshots. It is better to have some tests for more cases, than to have none.

Anyway, let me share what are the difficulties I saw around Playwright, assuming it is used for visual tests:
1) It is officially supported only on [specific Linux distributions](https://playwright.dev/docs/release-notes#-debian-11-bullseye-support)
2) Snapshots taken under one OS [will not match](https://playwright.dev/docs/test-snapshots#:~:text=Screenshots%20differ%20between%20browsers%20and%20platforms) snapshots taken another OS
3) Snapshots taken in headless mode [will not match](https://github.com/microsoft/playwright/issues/12683) snapshots taken in headed

I have observed other edge cases too, where testing in parallel resulted in confusing trace files.
Still, I think Playwright is a very good tool, worth using. Knowing the above difficulties, teams can plan how to address them in their development and CI processes.

If there is interest, I can extend this PR with some simple Playwright test too.

kalinkrustev:
Yes, I agree integration tests should happen later, as they will be harder
to maintain in case unit tests are missing too many cases.

melloware:
Closing for now.

PR #3646, opened 2022-11-15
fix #3535 regression

kalinkrustev:
fix: #3535


melloware:
Love it!  Thanks for fixing this and adding the tests.

PR #3648, opened 2022-11-15
Fix #3647: Image fallback URL if there is an error

### LINKED ISSUE #3647: Avatar: Fallback to label or icon variant when image loading fails

It would be great if the Avatar component would have some kind of default fallback functionality. Currently, if I specify the `image` property it will always be rendered as `</img>` even if the image cannot be loaded. The only option to overcome this issue is to define a function for `onError` which I can use to change the `event.target.src` but if I don't have a fallback image and I want to use an icon just like the component would do I need to change then `event.target.outerHTML` which is complicated and will not follow PrimeReact changes.

_No response_

MUI uses the following solution to try to load images before render:
https://github.com/mui/material-ui/blob/5eb6caf89e763f346e5c57494d6aa45eff17d3b3/packages/mui-material/src/Avatar/Avatar.js#L92-L129

Based on that they decide what to render:
https://github.com/mui/material-ui/blob/5eb6caf89e763f346e5c57494d6aa45eff17d3b3/packages/mui-material/src/Avatar/Avatar.js#L150-L180

In my opinion this can be an opt-out feature for example defining a new prop `defaultFallback` with `true` as default value and change this line:
https://github.com/primefaces/primereact/blob/959f77dc9da32684ea7e053147e0fa62f2167ea2/components/lib/avatar/Avatar.js#L8
to
```javascript
if (props.image && (!prop.defaultFallback || (prop.defaultFallback && imageLoaded))) {
```

However, this solution as opt-out can be a breaking change for those who use `onError` to solve image loading issues.

_No response_

### Additional context

_No response_

melloware:
### Defect Fixes
Fix #3647: Image fallback URL if there is an error

### DIFF
--- a/api-generator/components/avatar.js
+++ b/api-generator/components/avatar.js
@@ -18,6 +18,18 @@ const AvatarProps = [
         description: 'Defines the image to display.'
     },
     {
+        name: 'imageAlt',
+        type: 'any',
+        default: 'null',
+        description: 'It specifies an alternate text for an image, if the image cannot be displayed.'
+    },
+    {
+        name: 'imageFallback',
+        type: 'string',
+        default: 'default',
+        description: 'Defines a fallback image or URL if the main image fails to load. If "default" will fallback to label then icon.'
+    },
+    {
         name: 'size',
         type: 'string',
         default: 'null',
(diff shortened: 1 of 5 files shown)


KOliver94 (approved):
Great job, I really like it! LGTM

PR #3652, opened 2022-11-15
Fix #3643: RowGroupHeaderTemplate allow custom rendering

### LINKED ISSUE #3643: DataTable: RowGrouping Header Template allow control of colspan

In the current implementation of the RowGroupHeader, the template is provided which shows template data in a single-row view. We cannot aggregate data (like Summation, Average, Max Value, etc )  for each column below the respective column headers. Can we have the ability to have a template added so that we can aggregate grouped data and show it at the RowGroupHeader level when using subHeader mode in row grouping?

It should support column resizing and column reordering as well for better view of data and summary

_No response_

In the current code I can see a single **td** is used with colspan equals to total number of column and rowGroupHeaderTemplate is accepted with puts it in the single **td**. Instead can we remove the colspan and accept template from user which would generate multiple td for each column based on provide template

The alternative is getting an element by class name used to represent the row group header class removing the colspan from the td and adding the corresponding number of td based on columns and generating the template 

### Additional context

_No response_

melloware:
### Defect Fixes
Fix #3643: RowGroupHeaderTemplate allow custom rendering

```js
const rowGroupHeaderTemplate= (data, props, options) => {
    options.customRendering = true;

    return (
      <>
        <td colSpan={2}>name xx</td>
        <td style={{ background: "lightblue" }}>price template expanded</td>
      </>
    );
  };
```

### DIFF
--- a/components/lib/datatable/TableBody.js
+++ b/components/lib/datatable/TableBody.js
@@ -813,14 +813,22 @@ export const TableBody = React.memo(
             if (isSubheaderGrouping && shouldRenderRowGroupHeader(props.value, rowData, index - props.first)) {
                 const style = rowGroupHeaderStyle();
                 const toggler = props.expandableRowGroups && <RowTogglerButton onClick={onRowToggle} rowData={rowData} expanded={expanded} expandedRowIcon={props.expandedRowIcon} collapsedRowIcon={props.collapsedRowIcon} />;
-                const content = ObjectUtils.getJSXElement(props.rowGroupHeaderTemplate, rowData, { index, props: props.tableProps });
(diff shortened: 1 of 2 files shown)


mertsincan:
@melloware I think you can send the content in rowGroupHeaderTemplate instead of `options.customRendering = true;`. Exp; https://github.com/primefaces/primereact/blob/master/components/lib/menu/Menu.js#L181

WDYT?

melloware:
I think the difference here in the table is the outer element had to be the TR...the users just want to control the TD creation but 99% of the time they don't. I am open to suggestions this was the least intrusive to legacy code!  

Let me know what you want to do.

PR #3664, opened 2022-11-17
Fix #3476: DataTable: TypeError when adding and starting editing a new row

### LINKED ISSUE #3476: DataTable: TypeError when adding and starting editing a new row
### Describe the bug

When adding a new row and immediately setting it in edit mode, an error `Cannot read properties of undefined (reading 'data')` is thrown in [BodyCell EditorCallback](https://github.com/primefaces/primereact/blob/7eb90e0ba81d8e2b59b9b58b27c99eec4d389269/components/lib/datatable/BodyCell.js#L253). `EditingMeta` was not initialized in this specific case.

### Reproducer

https://codesandbox.io/s/primereact-test-forked-224zjd?file=/src/index.js

### PrimeReact version

8.6.1

### React version

18.x

### Language

ALL

### Build / Runtime

Create React App (CRA)

### Browser(s)

_No response_

Use the reproducer URL, click to add a new row and start typing in one of the fields.

### Expected behavior

_No response_

yanxiyue:
#3476
1. The cause seems to be that `editingMeta` was not initialized in this specific case.
2. `editingMeta` will be updated by `useUpdateEffect` hook, when `BodyCell`'s `editingState` changed.
3. in this case, `BodyCell`  is not mounted yet, so, the update logic will not be executed  

 Therefore，I try change `useUpdateEffect`  to `React.useEffect`, it works well.

### DIFF
--- a/components/lib/datatable/BodyCell.js
+++ b/components/lib/datatable/BodyCell.js
@@ -463,13 +463,14 @@ export const BodyCell = React.memo((props) => {
         }
     }, [props.editingMeta]);
 
-    useUpdateEffect(() => {
+    React.useEffect(() => {
         if (props.editMode === 'cell' || props.editMode === 'row') {
             const callbackParams = getCellCallbackParams();
             const params = { ...callbackParams, editing: editingState, editingKey };
 
             props.onEditingMetaChange(params);
         }
+        // eslint-disable-next-line react-hooks/exhaustive-deps
     }, [editingState]);
 
     useUnmountEffect(() => {


components/lib/datatable/BodyCell.js
```diff
@@ -463,14 +463,16 @@ export const BodyCell = React.memo((props) => {
         }
     }, [props.editingMeta]);
 
-    useUpdateEffect(() => {
+    /* eslint-disable */
```
melloware, line 466: You should not have to disable eslint?

melloware:
@yanxiyue thanks for the fix can you also comment on the ticket so then I can assign it to you?

components/lib/datatable/BodyCell.js
yanxiyue: I didn't want to disable that, but I couldn't build it successfully, I'm not very familiar with react, please tell me how to do better

```sh
$ npm run build:lib

> primereact@8.7.3-SNAPSHOT build:lib
> NODE_ENV=production INPUT_DIR=components/lib/ OUTPUT_DIR=dist/ npm run build:package


> primereact@8.7.3-SNAPSHOT build:package
> npm run build:check && rollup -c && gulp build-resources && npm run build:api


> primereact@8.7.3-SNAPSHOT build:check
> npm run lint && npm run format:check && npm run type:check && npm run security:check && (NODE_ENV=test npm run test:check)


> primereact@8.7.3-SNAPSHOT lint
> next lint --max-warnings=0


./components/lib/datatable/BodyCell.js
473:8  Warning: React Hook React.useEffect has missing dependencies: 'editingKey', 'getCellCallbackParams', and 'props'. Either include them or remove the dependency array. However, 'props' will change when *any* prop changes, so the preferred fix is to destructure the 'props' object outside of the useEffect call and refer to those specific props inside React.useEffect.  react-hooks/exhaustive-deps
```

melloware:
@yanxiyue this should work...

```js
    React.useEffect(() => {
        if (props.editMode === 'cell' || props.editMode === 'row') {
            const callbackParams = getCellCallbackParams();
            const params = { ...callbackParams, editing: editingState, editingKey };

            props.onEditingMetaChange(params);
        }
     // eslint-disable-next-line react-hooks/exhaustive-deps
    }, [editingState]);
```

yanxiyue:
> 



> @yanxiyue this should work...
> 
> ```js
>     React.useEffect(() => {
>         if (props.editMode === 'cell' || props.editMode === 'row') {
>             const callbackParams = getCellCallbackParams();
>             const params = { ...callbackParams, editing: editingState, editingKey };
> 
>             props.onEditingMetaChange(params);
>         }
>      // eslint-disable-next-line react-hooks/exhaustive-deps
>     }, [editingState]);
> ```

got it. fixed.

melloware (approved):

PR #2744, opened 2022-04-12
Fix #2745: Multiselect inline flex

kalinkrustev:
Add `inline`, `flex` and `itemClassName` properties for `MultiSelect`, to control the panel display and make it possible to render it inline or apply flex layout to it.

image
image

Fix #2745

PR #3708, opened 2022-11-25
Fix #3302: Multiselect add overlayVisible property

### LINKED ISSUE #3302: MultiSelect: Add "overlayVisible" property

Add a `overlayVisible` property to the MultiSelect component. This property would give the option to have multiselect open when navigated to.

Currently, if multiselect has an overlay button, when user clicks overlay button they have to click again on multiselect to open. Not very user-friendly.

This is currently a feature in PrimeNG as shown here: https://www.primefaces.org/primeng/multiselect

melloware:
### Defect Fixes
Fix #3302: Multiselect add overlayVisible property

PrimeNG has this property


components/lib/multiselect/multiselect.d.ts
```diff
@@ -74,64 +74,65 @@ interface MultiSelectFilterOptions {
...
-    value?: any;
-    options?: SelectItemOptionsType;
-    optionLabel?: string;
-    optionValue?: string;
     optionDisabled?: MultiSelectOptionDisabledType;
-    optionGroupLabel?: string;
     optionGroupChildren?: string;
+    optionGroupLabel?: string;
     optionGroupTemplate?: MultiSelectOptionGroupTemplateType;
-    display?: MultiSelectDisplayType;
```
kalinkrustev, line 88: I hope merge conflicts with #2744 are not going to be too complex again.

melloware: I just merged yours. Its a great feature and I will rebase mine.

melloware:
@kalinkrustev the original use case was they wanted to have the panel open on render.   so if `overlayVisible={true}` the overlay is displayed on render.  However because of how the overlay positioning works the overlay is not in the right position and that is why the 100ms delay to allow the component to render and then position the overlay properly. 

If you have a better solution a PR is welcome.

kalinkrustev:
So we aim for the initial visibility case? This should be reflected in the docs.

melloware:
Yes for initial visibility.

PR #3717, opened 2022-11-29
Fix #3716: The show() method causes the browser to crash.

### LINKED ISSUE #3716: Messages/Toast: The messages.current.show([]) method causes the browser to crash
### Describe the bug

"_I think prime should handle empty messages. It would provide better developer experience so that we don't need to check for empty messages on our end, especially when they are created dynamically._"

### Reproducer

https://codesandbox.io/s/small-moon-djl88s?file=/src/demo/MessagesDemo.js

### PrimeReact version

8.7.2

### React version

18.x

### Language

ES6

### Build / Runtime

Next.js

### Browser(s)

_No response_

_No response_

### Expected behavior

_No response_

habubey:
### Defect Fixes
Fix #3716

### DIFF
--- a/components/lib/messages/Messages.js
+++ b/components/lib/messages/Messages.js
@@ -83,15 +83,16 @@ export const Messages = React.memo(
         return (
             <div id={props.id} ref={elementRef} className={props.className} style={props.style} {...otherProps}>
                 <TransitionGroup>
-                    {messagesState.map((message) => {
-                        const messageRef = React.createRef();
+                    {messagesState &&
+                        messagesState.map((message) => {
+                            const messageRef = React.createRef();
 
-                        return (
-                            <CSSTransition nodeRef={messageRef} key={message.id} classNames="p-message" unmountOnExit timeout={{ enter: 300, exit: 300 }} options={props.transitionOptions}>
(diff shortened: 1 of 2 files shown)


components/lib/messages/Messages.js
```diff
@@ -83,15 +83,16 @@ export const Messages = React.memo(
...
-                    {messagesState.map((message) => {
-                        const messageRef = React.createRef();
-
-                        return (
-                            <CSSTransition nodeRef={messageRef} key={message.id} classNames="p-message" unmountOnExit timeout={{ enter: 300, exit: 300 }} options={props.transitionOptions}>
-                                <UIMessage ref={messageRef} message={message} onClick={props.onClick} onClose={onClose} />
-                            </CSSTransition>
-                        );
-                    })}
+                    {messagesState &&
```
melloware, line 86: @habubey can you make the same change in Toast.js please it has the same issue.

melloware (changes requested):

melloware (approved):

components/lib/messages/Messages.js
habubey: Done 🚀

PR #3724, opened 2022-11-30
Fix Chart initialized in unmounted component

mpressmar:
The `initChart` method will in some case asynchronously create a chart object (after loading `chart.js/auto`). If during the asynchronous process the Chart component was already unmounted, the Chart will still be created, but for a `null` canvas. As a consequence, a zombie Chart is created that is never cleaned up.

**Context/Usage**

We have a simple usage of a line chart, which is rendered for each detail of a master-detail view. That means switching the selection between the entries in the master view will each time unmount the Chart component and then create a new one with different curve data.

The mentioned issue can be provoked already with a minimal Chart as this (but also with actual configuration and data):
```
        <Chart
               type="line"
               data={{labels: [], datasets: []}}
               options={{}}/>
```

**Issue**

#3725 

Oftentimes we see these two errors in the log while switching from one detail to the other:

image

image

### DIFF
--- a/components/lib/chart/Chart.js
+++ b/components/lib/chart/Chart.js
@@ -39,6 +39,7 @@ const PrimeReactChart = React.memo(
                     if (!canvasRef.current) {
                         return;
                     }
+
                     if (module) {
                         if (module.default) {
                             // WebPack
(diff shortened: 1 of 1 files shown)


melloware (approved):

melloware:
Looks like its failing the ESLINT check

```
42:21  Error: Expected blank line before this statement.  padding-line-between-statements
```

PR #3754, opened 2022-12-07
Fix #3753: Menu add `popupAlignment` property

### LINKED ISSUE #3753: Menu: Add ability to right align

It would be nice if there was an option to right align a dropdown menu to it's initiator. A good example from Bootstrap here:

https://getbootstrap.com/docs/4.0/components/dropdowns/#menu-alignment

Similar to PrimeVue request: https://github.com/primefaces/primevue/issues/3381

No

Right now it seems to always default to a best fit, so would be nice to for instance, right align first and then only best fit after that.

I've tried using an css offset which works if the menu needs to be a bit lower or something. But an offset here, would completely throw things off on the auto align. For instance on desktop it could be made to look fine, but then on mobile the offset would completely throw it off.

### Additional context

Currently the left side example is what happens, the right side example is the intended result.

image

melloware:
### Defect Fixes
Fix #3753: Menu add alignment property

**Left:** (default and current behavior)
image

**Right:**
image


Ayman9334:
we neeeeed this!

melloware:
I am adding this in another PR.

Issue #2523, opened 2021-12-14
DataTable/TreeTable: stricter generically typed API


VsevolodGolovanov:
**I'm submitting a ...**  (check one with "x")
```
[ ] bug report
[ ] feature request
[ ] support request => Please do not submit support request here, instead see https://forum.primefaces.org/viewforum.php?f=57
```

**Codesandbox Case (Bug Reports)**

**Current behavior**
The Tables have a lot of `any`s in them:
```
export interface DataTableProps {
    value?: any[];
    selection?: any | any[];
```
And many more.
`any` is not just a type - it disables static typing. TS actually [recommends not using it all](https://www.typescriptlang.org/docs/handbook/declaration-files/do-s-and-don-ts.html#any).
This makes code using Tables less resilient to PrimeReact update changes, less type safe in general (selection should consist of data entiries, but this is not checked), and makes PrimeReact API harder to understand (e.g. rowClassName and cellClassName both accept the data argument - what is it? You have to run some code to figure that out).

**Expected behavior**
Avoid `any` and bind associated types with generics. E.g.:
```
export interface DataTableProps<RowType> {
    value?: RowType[];
    selection?: RowType | RowType[];
```

**Minimal reproduction of the problem with instructions**

**Please tell us about your environment:**

* **React version:**

* **PrimeReact version:**
7.1.0

* **Browser:** [all | Chrome XX | Firefox XX | IE XX | Safari XX | Mobile Chrome XX | Android X.X Web Browser | iOS XX Safari | iOS XX UIWebView | iOS XX WKWebView ]

* **Language:** [all | TypeScript X.X | ES6/7 | ES5]

DerDorius:
![image](https://user-images.githubusercontent.com/77979816/187033086-cb64a420-70db-46ee-b847-27142873e6aa.png)
This would improve the Dev-Experience soo much

melloware:
@DerDorius feel free to submit a PR!

melloware:
I figured we could keep DataTable/TreeTable as a separate PR/ticket.

diamanthaxhimusa:
DataTable/TreeTable still need Generics right? Can I submit a PR for this?

melloware:
Yes please!

PR #3822, opened 2022-12-24
Fix: #2523 DataTable/TreeTable: stricter generically typed API

bu3alwa:
Issue: #2523

I hope this should provide generic typing to `DataTables` .  Below is a nextjs page I used to test it. Should be able to make a new page in the project to test it. Since the project is in js its hard to test. 

``` typescript
/* eslint-disable react-hooks/rules-of-hooks */
import { NextPage } from 'next';
import { useEffect, useId, useState } from 'react';
import { Column } from '../../components/lib/column/column';
import { DataTable, DataTableSelection } from '../../components/lib/datatable/datatable';

const index: NextPage = () => {
    const id = useId();

    const columns = [
        { field: 'code', header: 'Code' },
        { field: 'name', header: 'Name' },
        { field: 'category', header: 'Category' },
        { field: 'quantity', header: 'Quantity' }
    ];

    const data = [
        { code: 'test', name: 'test', category: 'test', quantity: 'test' },
        { code: 'test', name: 'test', category: 'test', quantity: 'test' },
        { code: 'test', name: 'test', category: 'test', quantity: 'test' },
        { code: 'test', name: 'test', category: 'test', quantity: 'test' }
    ];

    const [selected, SetSelected] = useState<DataTableSelection<typeof data>>();

    useEffect(() => {
        console.log({ selected });
    }, [selected]);

    return (
        <>
            <DataTable
                dataKey="id"
                value={data}
                selectionMode="single"
                cellSelection
                selection={selected}
                onSelectionChange={(e) => {
                    SetSelected(e.value);
                }}
                responsiveLayout="scroll"
            >
                {/* <Column selectionMode="single"/> */}
                {columns.map((e) => {
                    return <Column key={id + e.field} field={e.field} header={e.header} />;
                })}
            </DataTable>
        </>
    );
};

export default index;
```

Any reason why Next and the project is not in typescript. It would make it much easier to develop if everything was in ts.

Probably not perfect but its a start to iterate on.

### DIFF
--- a/components/lib/datatable/datatable.d.ts
+++ b/components/lib/datatable/datatable.d.ts
@@ -158,7 +158,7 @@ interface DataTableCellClickEventParams {
     originalEvent: React.MouseEvent<HTMLElement>;
     value: any;
     field: string;
-    rowData: any;
+    rowData: DataTableValue;
     rowIndex: number;
     cellIndex: number;
     selected: boolean;
@@ -173,7 +173,7 @@ interface DataTableRowEditSaveParams extends DataTableRowEditParams {
 }
 
 interface DataTableRowEditCompleteParams extends DataTableRowEventParams {
-    newData: any;
+    newData: DataTableValue;
     field: string;
     index: number;
 }
@@ -187,9 +187,9 @@ interface DataTableSelectParams {
 interface DataTableUnselectParams extends DataTableSelectParams {}
 
 interface DataTableExportFunctionParams {
-    data: any;
+    data: DataTableValueArray;
     field: string;
-    rowData: any;
+    rowData: DataTableValue;
     column: Column;
 }
 
@@ -242,24 +242,24 @@ interface DataTableValue extends Record<string, any> {}
 
 interface DataTableValueArray extends Array<DataTableValue> {}
 
-type DataTableRowSelection<TValue extends DataTableValueArray> = {
+type DataTableRowData<TValue extends DataTableValueArray> = {
     [K in keyof TValue]: TValue[K];
 };
 
-type DataTableRowSelectionArray<TValue extends DataTableValueArray> = DataTableRowSelection<TValue>[];
+type DataTableRowDataArray<TValue extends DataTableValueArray> = DataTableRowData<TValue>[];
 
 type DataTableCellSelection<TValue extends DataTableValueArray> = {
     cellIndex: number;
     column: Column;
     field: string;
     props: ColumnProps;
-    rowData: DataTableRowSelection<TValue>;
+    rowData: DataTableRowData<TValue>;
     rowIndex: number;
     selected: boolean;
     value: TValue[keyof TValue];
 };
 
-type DataTableSelection<TValue extends DataTableValueArray> = DataTableRowSelection<TValue> | DataTableRowSelectionArray<TValue> | DataTableCellSelection<TValue>;
+type DataTableSelection<TValue extends DataTableValueArray> = DataTableRowData<TValue> | DataTableRowDataArray<TValue> | DataTableCellSelection<TValue>;
 
 export interface DataTableProps<TValue extends DataTableValueArray> extends Omit<React.DetailedHTMLProps<React.InputHTMLAttributes<HTMLDivElement>, HTMLDivElement>, 'size' | 'onContextMenu' | 'ref' | 'value'> {
     id?: string;
@@ -279,11 +279,11 @@ export interface DataTableProps<TValue extends DataTableValueArray> extends Omit
     defaultSortOrder?: DataTableSortOrderType;
     dragSelection?: boolean;
     editMode?: string;
-    editingRows?: any[] | DataTableEditingRows;
+    editingRows?: DataTableValueArray | DataTableEditingRows;
     emptyMessage?: DataTableEmptyMessageType;
     expandableRowGroups?: boolean;
     expandedRowIcon?: string;
-    expandedRows?: any[] | DataTableExpandedRows;
+    expandedRows?: DataTableValueArray | DataTableExpandedRows;
     exportFilename?: string;
     filterDelay?: number;
     filterDisplay?: DataTableFilterDisplayType;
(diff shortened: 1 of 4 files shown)


components/lib/datatable/datatable.d.ts
```diff
@@ -212,33 +212,56 @@ interface DataTableRowExpansionTemplate {
...
-    props: DataTableProps;
+interface DataTableRowClassNameOptions<TValue extends DataTableValueArray> {
+    props: DataTableProps<TValue>;
 }
 
-interface DataTableCellClassNameOptions {
-    props: DataTableProps;
+interface DataTableCellClassNameOptions<TValue extends DataTableValueArray> {
+    props: DataTableProps<TValue>;
     rowData: any;
```
melloware, line 221: Did you miss an any here?

melloware:
Thanks I will test this out!

melloware (changes requested):

components/lib/datatable/datatable.d.ts
bu3alwa: Yes I did. I'll push the change but should I also rename `DataTableRowSelection` to something else since its used here too?

melloware: I have to look but there isn't it just a Value Array not a Selection Array?

bu3alwa: Looks like it gets passed to `ObjectUtils.getPropValue` in `BodyCell` but I am having a hard time figuring out what it's doing.

bu3alwa:
```
options:{
column: Object { "$$typeof": Symbol("react.element"), type: Column()
, key: ".$=2R2sm=2code", … }
field: "code"
frozenRow: false
props: Object { dataKey: "id", selectionMode: "single", cellSelection: true, … }
rowIndex: 0
}
```

It looks like rowData doesn't exist in options. This is what I get as an output. Not sure if there is a specific usecase where it will show up but it seems like the rest of the fields in options are also not typed. Let me know how to proceed with this.

melloware: You are right i think that definiton is wrong I think it should be this right?

```
interface DataTableCellClassNameOptions {
    props: DataTableProps;
    column: Column;
    field: string;
    frozenRow: boolean;
    rowIndex: number;
}
```
?

bu3alwa: Looks correct to me

melloware:
This is looking good.  I am going to check out your branch and do some testing with my TypeScript project here which uses the Datatable extensively: https://github.com/melloware/quarkus-monorepo

melloware (changes requested):
So this is a nice start but there a lot more `any` values you can close off.

For example a lot of times `data: any` is the rowData value the singular `DataTableValue`.   If you see anywhere its an `any` that is many values its `DataTableValueArray`.

components/lib/datatable/datatable.d.ts
```diff
@@ -358,20 +383,20 @@ export interface DataTableProps extends Omit<React.DetailedHTMLProps<React.Input
...
     onValueChange?(value: any[]): void;
-    rowClassName?(data: any, options: DataTableRowClassNameOptions): object | string;
-    rowEditValidator?(data: any, options: DataTableRowEditValidatorOptions): boolean;
+    rowClassName?(data: any, options: DataTableRowClassNameOptions<TValue>): object | string;
+    rowEditValidator?(data: any, options: DataTableRowEditValidatorOptions<TValue>): boolean;
     rowExpansionTemplate?(data: any, options: DataTableRowExpansionTemplate): React.ReactNode;
-    showRowReorderElement?(data: any, options: DataTableShowRowReorderElementOptions): boolean | undefined | null;
-    showSelectionElement?(data: any, options: DataTableShowSelectionElementOptions): boolean | undefined | null;
+    showRowReorderElement?(data: any, options: DataTableShowRowReorderElementOptions<TValue>): boolean | undefined | null;
+    showSelectionElement?(data: any, options: DataTableShowSelectionElementOptions<TValue>): boolean | undefined | null;
```
melloware, line 395: For example in the above I believe `data: any` should be `data: DataTableValue` no?

components/lib/datatable/datatable.d.ts
```diff
@@ -130,9 +130,9 @@ interface DataTableDataSelectableParams {
     index: number;
 }
 
-interface DataTableSelectionChangeParams {
+interface DataTableSelectionChangeParams<TValue extends DataTableValueArray> {
     originalEvent: React.SyntheticEvent;
-    value: any;
+    value: DataTableSelection<TValue>;
     type?: string;
     [key: string]: any;
```
melloware, line 137: `key: string` should be `TValue[keyof TValue]` no?

melloware (changes requested):

components/lib/datatable/datatable.d.ts
bu3alwa: Actually I am not sure what that is. All the values of rows are in `value`. `[key: string]: any` seems to not exist in the object from what I have seen but I am not sure if there is a case where it shows up.

components/lib/datatable/datatable.d.ts
bu3alwa: I have not used that function before but I could change it to `DataTableValue`  if you think its correct. There are multiple functions with `data: any` which I am not sure what to type. There are also cases where some types are `any[] | ...` and I am not sure what purpose those serve

melloware: In most cases if you look a the source code any Datatable function passing `data: any` is USUALLY 1 row of data from the table which is your `DataTableValue` singular definition.

melloware:
I have been messing with your PR and added some more mappings here: https://github.com/primefaces/primereact/pull/3824

bu3alwa:
I was thinking more type inference from `TValue` so I renamed `DataTableSelection` to `DataTableRowData` to use in other places so anything with. Unless those functions can pass in any value that don't exists in TValue.

melloware:
I like that idea!

melloware (approved):

melloware:
@bu3alwa i am going to submit another PR.  I am confused by some of your typing I think you have "over" typed some of these objects?

melloware:
There are only two types of record.

1. the `value` and `frozenValue` are an array of records

2. The `rowData` is a single record of #1.   It looks like you have mapped 4-6 different types which all map back to one of these two?

bu3alwa:
@melloware it's fine I was trying to get them to be more specific so you can't just pass any value that does not exist. But I was finding it pretty hard to test anything when the initial project is written in JS. Anyways, as long as there is more typing it should be a bit better now.

PR #3825, opened 2022-12-27
Fix: #2523 DataTable/TreeTable: stricter generically typed API

melloware:
### Defect Fixes
Fix: #2523 DataTable/TreeTable: stricter generically typed API

cc @bu3alwa  please review this.  Why isn't `frozenRow` the exact same type as `value` as that is how its treated in the code?

### DIFF
--- a/components/lib/datatable/datatable.d.ts
+++ b/components/lib/datatable/datatable.d.ts
@@ -125,8 +125,8 @@ interface DataTablePFSEvent extends DataTablePageParams, DataTableSortParams, Da
     [key: string]: any;
 }
 
-interface DataTableDataSelectableParams {
-    data: any;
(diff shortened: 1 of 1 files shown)


components/lib/datatable/datatable.d.ts
```diff
@@ -387,7 +387,7 @@ export interface DataTableProps<TValue extends DataTableValueArray> extends Omit
     onSort?(e: DataTablePFSEvent): void;
     onStateRestore?(state: object): void;
     onStateSave?(state: object): void;
-    onValueChange?(value: any[]): void;
+    onValueChange?(value: TValue): void;
```
melloware, line 390: onValueChange is TValue as its the exact same value as `value`

PR #3856, opened 2022-12-29
Fix #3838: InputMask flicker on focus

### LINKED ISSUE #3838: InputMask Focus Flicker
When inputmask receives focus for the first time, it flickers.

melloware:
### Defect Fixes
Fix #3838: InputMask flicker on focus

This is what was causing the flicker but I have not seen any side effects by removing it?


habubey:
@melloware I found there was a problem with "Float Label" section. The mask inside disappeared on the click after the first click.

Screenshot 2023-05-22 at 14 04 32

melloware:
You are right.  This is NOT the correct fix.

PR #3932, opened 2023-01-07
Fix #3701: Tooltip only register 1 show/hide event

### LINKED ISSUE #3701: Tooltip: registering show/hide events twice on each target
### Describe the bug

Tooltip is registering 2 show and 2 hide events on every tooltip target.

By switching from Mount Effect + Update Effect to just a single React.useEffect reduces the amount of registered listeners.  Show and Hide events were being registered **twice** on all components and this is due to the fact the functions being added are anonymous.

See MDN: https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener

> Note: If a particular anonymous function is in the list of event listeners registered for a certain target, and then later in the code, an identical anonymous function is given in an addEventListener call, the second function will also be added to the list of event listeners for that target.
> 
> Indeed, anonymous functions are not identical even if defined using the same unchanging source-code called repeatedly, even if in a loop.
> 
> Repeatedly defining the same unnamed function in such cases can be problematic. (See [Memory issues](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener#memory_issues), below.)

If you go to the current showcase and Inspect the button with a tooltip registered you will see this...

image

### Reproducer

https://primefaces.org/primereact/tooltip/

### PrimeReact version

8.7.2

### React version

17.x

### Language

ALL

### Build / Runtime

Next.js

### Browser(s)

ALL

1. Right click on Template button and inspect the DOM events
2. Notice `mouseenter` and `mouseleave` are registered twice

### Expected behavior

Only 1 show and hide event registered

melloware:
### Defect Fixes
Fix #3701: Tooltip only register 1 show/hide event

### DIFF
--- a/components/lib/tooltip/Tooltip.js
+++ b/components/lib/tooltip/Tooltip.js
@@ -1,6 +1,6 @@
 import * as React from 'react';
 import PrimeReact from '../api/Api';
-import { useMountEffect, useOverlayScrollListener, useResizeListener, useUnmountEffect, useUpdateEffect } from '../hooks/Hooks';
+import { useOverlayScrollListener, useResizeListener, useUnmountEffect, useUpdateEffect } from '../hooks/Hooks';
 import { Portal } from '../portal/Portal';
 import { classNames, DomHandler, ObjectUtils, ZIndexUtils } from '../utils/Utils';
 import { TooltipBase } from './TooltipBase';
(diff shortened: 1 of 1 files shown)


melloware:
No longer needed.

PR #4002, opened 2023-01-28
feat: replace newline character with its unicode version while exporting csv

### LINKED ISSUE #4001: DataTable: Export Function Newline Separator

Hi,

In the current DataTable if we have a column with newline character (\n) and we want to export that column, it brokes the csv layout. Because, newline character is considered as newline in the exported csv also. To fix that, we can use custom exportFunction and replace \n with it's unicode version **\u2028**.  I think doing that in the source code would make more sense because some people can have problem with that in the future.

_No response_

_No response_

_No response_

### Additional context

_No response_

EmirBoyaci:
Closes #4001

### Checks
I am currently using MacOS and Microsoft Excel. For me it works properly, I think that PR should be tested against other OS' and other applications to not break anything.

### DIFF
--- a/components/lib/datatable/DataTable.js
+++ b/components/lib/datatable/DataTable.js
@@ -1200,7 +1200,9 @@ export const DataTable = React.forwardRef((props, ref) => {
             const { field, header, exportable } = column.props;
 
             if (exportable && field) {
-                csv += '"' + (header || field) + '"';
+                const columnHeader = String(header || field).replace(/\n/g, '\u2028');
+
+                csv += '"' + columnHeader + '"';
 
                 if (i < columns.length - 1) {
                     csv += props.csvSeparator;
--- a/components/lib/datatable/DataTable.js
+++ b/components/lib/datatable/DataTable.js
@@ -1200,7 +1200,9 @@ export const DataTable = React.forwardRef((props, ref) => {
             const { field, header, exportable } = column.props;
 
             if (exportable && field) {
-                const columnHeader = String(header || field).replace(/\n/g, '\u2028');
+                const columnHeader = String(header || field)
+                    .replace(/"/g, '""')
+                    .replace(/\n/g, '\u2028');
 
                 csv += '"' + columnHeader + '"';
 
--- a/components/lib/datatable/DataTable.js
+++ b/components/lib/datatable/DataTable.js
@@ -1224,8 +1224,7 @@ export const DataTable = React.forwardRef((props, ref) => {
                         if (props.exportFunction) {
                             cellData = props.exportFunction({ data: cellData, field, rowData: record, column });
                         } else {
-                            cellData = String(cellData).replace(/"/g, '""');
-                            cellData = String(cellData).replace(/\n/g, '\u2028');
+                            cellData = String(cellData).replace(/"/g, '""').replace(/\n/g, '\u2028');
                         }
                     } else cellData = '';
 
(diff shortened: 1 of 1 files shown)


melloware:
I will test this on Windows right now.

melloware (changes requested):
Also need to do the Column Headers too. I just teested a header with `\n` and it breaks the CSV also....

```js
const columnHeader = String(header || field).replace(/\n/g, '\u2028');

csv += '"' + columnHeader + '"';
```

melloware:
But forgot to mention it does work fine on Windows!

EmirBoyaci:
> Also need to do the Column Headers too. I just teested a header with `\n` and it breaks the CSV also....
> 
> ```js
> const columnHeader = String(header || field).replace(/\n/g, '\u2028');
> 
> csv += '"' + columnHeader + '"';
> ```

Done

> But forgot to mention it does work fine on Windows!

Nice to know!

components/lib/datatable/DataTable.js
```diff
@@ -1219,7 +1221,12 @@ export const DataTable = React.forwardRef((props, ref) => {
                     let cellData = ObjectUtils.resolveFieldData(record, field);
 
                     if (cellData != null) {
-                        cellData = props.exportFunction ? props.exportFunction({ data: cellData, field, rowData: record, column }) : String(cellData).replace(/"/g, '""');
+                        if (props.exportFunction) {
+                            cellData = props.exportFunction({ data: cellData, field, rowData: record, column });
+                        } else {
+                            cellData = String(cellData).replace(/"/g, '""');
+                            cellData = String(cellData).replace(/\n/g, '\u2028');
```
melloware, line 1228: I wonder if on the second line yu need String() since its already converted to a string?

```js
cellData = String(cellData).replace(/"/g, '""');
 cellData = cellData.replace(/\n/g, '\u2028');
 ```
or can you chain them?

```js
cellData = String(cellData).replace(/"/g, '""').replace(/\n/g, '\u2028');
```
??

melloware (changes requested):

components/lib/datatable/DataTable.js
melloware: nd actually shouldn't we do the same in headers with the `"` as couldn't a header have a quote in it that would need to be escaped as well?

EmirBoyaci: Yes, that's true. I can chain, I think it is better.

EmirBoyaci: > nd actually shouldn't we do the same in headers with the `"` as couldn't a header have a quote in it that would need to be escaped as well?

To be honest, I did not think about that. Can DataTable support header with newline character?

melloware: Yeah I jus ttried it with the Showcase changing header from Category Name to `'Category " \n Name'`

EmirBoyaci: Okay so with https://github.com/primefaces/primereact/pull/4002/commits/ed5a221b7b084aed722c7679a0e8928608fef295 I think it should be fine, can you try?

melloware:
I can confirm the CSV is totally broken with `header: 'Category " \n Name'`

image

But if I make it

```js
const columnHeader = String(header || field)
                    .replace(/"/g, '""')
                    .replace(/\n/g, '\u2028');
```

CSV comes out correctly...
image

melloware (approved):
Now that is teamwork!

EmirBoyaci:
> Now that is teamwork!

Exactly! Thanks for your efforts too!

melloware:
I am going to merge this and then update my own PR with your changes: https://github.com/primefaces/primereact/pull/3980

PR #4013, opened 2023-02-01
Fix #3993: DatatableLazyEvent

### LINKED ISSUE #3993: DataTable: Page Sort Filter Event Typescript
I think the typescript doesn't now match what is happening in the code..  So in my lazy table example project I have this declared using `DatatablePFSEvent`

```ts
	const initialParams: DataTablePFSEvent = {
		first: 0,
		rows: 5,
		page: 1,
		sortField: '', // single sort only
		sortOrder: SortOrder.UNSORTED, // single sort only
		multiSortMeta: [
			{ field: 'make', order: SortOrder.ASC },
			{ field: 'model', order: SortOrder.ASC }
		],
		filters: {}
	};

	const [tableParams, setTableParams] = useState<DataTablePFSEvent>(initialParams);

	const onPage = (event: DataTablePFSEvent) => {
		setTableParams(event);
	};

	const onSort = (event: DataTablePFSEvent) => {
		setTableParams(event);
	};

	const onFilter = (event: DataTablePFSEvent) => {
		event['first'] = 0;
		setTableParams(event);
	};
```

this all works perfectly and the datatable passes the correct full event to each of page sort and filter.  But now the new declarations have broken this and paging sorting and filtering in the TypeScript only show the bare minimum even though the real JS object has the whole PFSEvent we need per.

You can see it all in action here with real lazy sorting and filterin with a real REST API here: https://github.com/melloware/quarkus-monorepo

melloware:
### Defect Fixes
Fix #3993: DatatableLazyEvent

### DIFF
--- a/components/lib/datatable/datatable.d.ts
+++ b/components/lib/datatable/datatable.d.ts
@@ -256,6 +256,15 @@ interface DataTableFilterEvent {
 }
 
 /**
+ * Custom sort,page,filter event used especially in lazy tables where the event contain the current state.
+ * @see {@link DataTableProps.onFilter}
+ * @see {@link DataTableProps.onSort}
+ * @see {@link DataTableProps.onPage}
+ * @event
+ */
+interface DataTableLazyEvent extends DataTablePageEvent, DataTableFilterEvent, DataTableSortEvent {}
+
+/**
  * Custom data selectable event.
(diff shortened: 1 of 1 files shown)


mertsincan:
Thanks @melloware, I changed it a little ;)

melloware:
Awesome as long as its fixed!

PR #4047, opened 2023-02-13
Fix #4046: Tooltip escape calls to querySelector

### LINKED ISSUE #4046: Tooltip: query selector doesn't support useId()
### Describe the bug

Tooltip does not support IDs that have been generated by React's useId, as per new standard practice for generating IDs

### Reproducer

https://codesandbox.io/s/primereact-test-forked-gk8i4k?file=/src/index.js

### PrimeReact version

8.7.3

### React version

18.x

### Language

TypeScript

### Build / Runtime

Create React App (CRA)

### Browser(s)

Firefox

_No response_

### Expected behavior

_No response_

melloware:
### Defect Fixes
Fix #4046: Escape calls to querySelector

useId from React generates an id like `:r0:` which the colons make it an invalid selector which must be escaped.

It needs to be escaped with CSS.escape: https://drafts.csswg.org/cssom/#the-css.escape%28%29-method

### DIFF
--- a/components/lib/styleclass/StyleClass.js
+++ b/components/lib/styleclass/StyleClass.js
@@ -149,7 +149,7 @@ export const StyleClass = React.forwardRef((inProps, ref) => {
                 return elementRef.current.parentElement.parentElement;
 
             default:
-                return document.querySelector(props.selector);
+                return document.querySelector(CSS.escape(props.selector));
         }
     };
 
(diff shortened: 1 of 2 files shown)


mertsincan:
Fixed in https://github.com/primefaces/primereact/commit/eeb738c5c01cdcbc3929439445dcff26d0c6e0ce

PR #4053, opened 2023-02-14
Fix #4049: Avatar check for image falsy

### LINKED ISSUE #4049: Avatar: Wrong className is getting used when image is undefined
### Describe the bug

Due to the changes in version 9.0.0 Avatar fallback is not working as intended because the wrong className is getting used. 

I'm not sure what acutally changed between 8.7.3 and 9.0.0 but right now
https://github.com/primefaces/primereact/blob/595bcf38148df727874bc66f4185e2c413904b98/components/lib/avatar/Avatar.js#L12-L13
`props.image && !imageFailed` will be `false` as `undefined` is a falsy value. This way the `onImageError` function will never be called and it will never set `imageFailed` to `true`. However
https://github.com/primefaces/primereact/blob/595bcf38148df727874bc66f4185e2c413904b98/components/lib/avatar/Avatar.js#L47
this line explicitly checks for `null` which will not catch `undefined` and as I said earlier `imageFailed` will never be set to `true`. So at the end we will use `p-avatar-image` which will create a transparent background.

### Reproducer

[CodeSandbox](https://codesandbox.io/s/primereact-avatar-issue-nxq959)

### PrimeReact version

9.0.0

### React version

18.x

### Language

TypeScript

### Build / Runtime

Vite

### Browser(s)

_No response_

1) Open attached CodeSandbox
2) Switch primereact version from 9.0.0 to 8.7.3 to see the difference.

### Expected behavior

Fallback icon Avatar should not have `p-avatar-image` in className and should not have transparent background.

melloware:
### Defect Fixes
Fix #4049: Avatar check for image falsy

### DIFF
--- a/components/lib/avatar/Avatar.js
+++ b/components/lib/avatar/Avatar.js
@@ -44,7 +44,7 @@ export const Avatar = React.forwardRef((inProps, ref) => {
     const containerClassName = classNames(
         'p-avatar p-component',
         {
-            'p-avatar-image': props.image !== null && !imageFailed,
+            'p-avatar-image': props.image && !imageFailed,
             'p-avatar-circle': props.shape === 'circle',
             'p-avatar-lg': props.size === 'large',
             'p-avatar-xl': props.size === 'xlarge',


KOliver94 (approved):
LGTM! Thanks for the quick fix. :)

PR #4216, opened 2023-04-03
Feature/frozen resize problem

balazsiisti:
If your datatable has "resizableColumns" props, for being able to manually resize the column width, and some columns has the "frozen" props to stick them in the left of the table, the frozen columns headers are overlapping themselves, causing a visibility issue.


melloware:
Can you please open an issue so this PR can be linked to the issue.

PR #4282, opened 2023-04-24
Fix #4281: Messages respect manually set id

### LINKED ISSUE #4281: Messages: Cannot pass Id property
Attempted to create an array of MessageMessages and pass Id to each item so when attempting to remove I can access the Id and know which item was removed. Id is being set automatically by the component and ignoring the Id I am setting. Codesandbox here - https://codesandbox.io/s/primereact-demo-forked-t6jm1f

melloware:
Fix #4281: Messages respect manually set id


melloware:
it looks like Toast already had these fixes.

Issue #4298, opened 2023-04-27
Calendar: time select updates the parent component state through props.value


spencer-wegner:
### Describe the bug

When using the time picker on the Calendar component, the parent state is updated by the Calendar component but it should not be. This prevents the previous value of the state variable from being accessed in a `setState()` call.

I believe the issue is in `/components/lib/calendar/Calendar.js` in `getCurrentDateTime()`.  I copy/pasted the method below. For a single selection this method returns `props.value` which can be a JavaScript Date object. Because it's an object it's returning a reference to the state variable being passed as `props.value` from the parent.

```
const getCurrentDateTime = () => {
    if (isSingleSelection()) {
        return props.value && props.value instanceof Date ? props.value : getViewDate();
    } else if (isMultipleSelection()) {
        if (props.value && props.value.length) {
            return props.value[props.value.length - 1];
        }
    } else if (isRangeSelection()) {
        if (props.value && props.value.length) {
            let startDate = props.value[0];
            let endDate = props.value[1];

            return endDate || startDate;
        }
    }

    return new Date();
};
```

Then, when the time picker is used, the `updateTime()` method is called which calls `getCurrentDateTime()`. Here in the first part of `updateTime()` the reference to the parent state variable is directly manipulated through the `setHours()`, `setMinutes()`, etc.

```
 const updateTime = (event, hour, minute, second, millisecond) => {
      let newDateTime = getCurrentDateTime();

      newDateTime.setHours(hour);
      newDateTime.setMinutes(minute);
      newDateTime.setSeconds(second);
      newDateTime.setMilliseconds(millisecond);

      ...
  };
```

I'm only using the single selection mode, but this code may impact multiple and range selection modes as well.

Note that changing the date with the calendar date selector or via manual input into the input field works fine. Only the time picker causes this issue.

### Reproducer

https://codesandbox.io/s/primereact-calendar-parent-state-bug-p8q386

### PrimeReact version

9.3.0

### React version

18.x

### Language

ES6

### Build / Runtime

Create React App (CRA)

### Steps to reproduce the behavior

1. Open the Code Sandbox linked above
2. Use the time picker to adjust the time
3. See in the console output how the previous dateTime and next dateTime value are the same

### Expected behavior

The Calendar component should not update the parent state (the variable passed to `props.value`). It should only pass the updated value via the `onChange` prop function and it's the parent component's responsibility to update the state of the date/time value passed to `props.value`.

melloware:
[...]

PR #4304, opened 2023-04-28
#4298 fix issue where calendar updates props.value on time select

spencer-wegner:
Resolves [#4298](https://github.com/primefaces/primereact/issues/4298)

Return copy of props.value Date object in getCurrentDateTime()

I also added `props.showTime` to the `useUpdateEffect()` so `updateInputField()` is called when there is a change to `props.showTime`.


melloware (approved):

melloware:
OK I refactored it a little bit but thanks for the start!

PR #4349, opened 2023-05-07
fix: updates incorrectly replacing old state instead of actual.

### LINKED ISSUE #4350: Toast/Messages: Fix updates incorrectly updating older version of the state
### Describe the bug

Without the fix, instead of being added, every update clears the state.

_Before the fix_:

https://user-images.githubusercontent.com/35939574/236701969-99135e8e-9b46-4d24-b4c6-dfb52ae51cc8.mp4

_After the fix_:

https://user-images.githubusercontent.com/35939574/236701976-834868fb-c593-4fbe-bf2e-d1f06f719b2e.mp4

### Reproducer

[codesandbox link](https://codesandbox.io/s/primereact-demo-forked-ltbpfm?file=/src/App.js)

### PrimeReact version

9.3.1

### React version

18.x

### Language

TypeScript

### Build / Runtime

Next.js

### Expected behavior

I don't know why it happens. I will try to investigate, but what I know for sure is that pr #4349 fixes it.

ecstrema:
fixes #4350


melloware:
Please open an issue with a reproducer showing your issue.

melloware (changes requested):
See comments

ecstrema:
fixes #4350

ecstrema:
I am trying to make a minimal reproduction right now

ecstrema:
repro in linked issue

melloware:
let me do some testing with this PR.  Also My guess is similar fixes need to be done to Messages as its similar to Toast.

components/lib/toast/Toast.js
```diff
@@ -19,13 +19,11 @@ export const Toast = React.memo(
 
         const show = (messageInfo) => {
             if (messageInfo) {
-                const messages = assignIdentifiers(messageInfo, true);
-
-                setMessagesState(messages);
+                setMessagesState((previousMessagesState) => assignIdentifiers(previousMessagesState, messageInfo, true));
             }
```
melloware, line 23: Can you apply similar fixes to the Messages component as it works the exact same.

melloware (changes requested):

components/lib/toast/Toast.js
ecstrema: Done, but there is a lot of duplication in these files. Some refactoring here would help.

melloware:
Ok I will test today. Yeah so far I have left them separate and not refactored just for fear there would be something different in the two but I will look today.

melloware (approved):

PR #4541, opened 2023-06-23
Fix #4518: svg images with no base size do not show up in preview

ecstrema:
Fixes #4518

PR #4542, opened 2023-06-23
Feat: show tiled background for images with transparency in image preview

ecstrema:
Goes with #4541

### DIFF
--- a/components/lib/image/Image.css
+++ b/components/lib/image/Image.css
@@ -53,6 +53,10 @@
     transition: transform .15s;
     max-width: 100vw;
     max-height: 100vh;
+    /* Show tiled background for images with transparency. */
+    background: #eee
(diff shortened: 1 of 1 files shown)


ecstrema:
The tiled background looks like this: 

image

melloware:
@ecstrema i wonder if we should open a new issue ticket.  Then I think what we can do is add a new property like `transparent=true/false` to the Image so it would add this transparency with an additional style...

```css
.p-image-preview .p-image-transparent {
    background: #eee
        url('data:image/svg+xml,<svg xmlns="http://www.w3.org/2000/svg" width="400" height="400" fill-opacity=".25" ><rect x="200" width="200" height="200" /><rect y="200" width="200" height="200" /></svg>');
    background-size: 15px 15px;
}
```

ecstrema:
Yup, that would probably be better to avoid breaking changes. 

We could also show a transparency toggle among the other actions (rotate, zoom, and close). But for all images with a background, it would do nothing which could be incomprehensible for users.

melloware:
Going to close this for now.

PR #4546, opened 2023-06-25
Fix #4134: Calendar allow manual date entry

### LINKED ISSUE #4134: Calendar: Allow typing dates manually

As a user, I would like to be able to type dates using the keyboard, and not be forced to use the UI. A common case where this is necessary is having to type your birthday: using the UI would take lots and lots of clicks, while using the keyboard would be much more immediate.

Currently, the interaction with the calendar's input box is very cumbersome. Particularly, pressing delete/backspace will wipe out the entire input, making it impossible to make any type of correction, while unexpectedly trashing the user's current selection, forcing them to reselect a date from the beginning.

Another specific case where keyboard interaction is broken is selecting the current value and pressing a number. You can reproduce this directly in https://primereact.org/calendar by:
1. Selecting a date through the UI.
2. Selecting all the text in the input box with the mouse.
3. Entering "2000" with the keyboard.
4. Realizing you have just typed "000" (the "2" was ignored).

There are likely other cases...

In my opinion, the calendar is not usable as it is. In fact, we are currently deciding against it until these problems are solved, in favor of a plain input box, because we already foresee user complaints from putting it in place.

userkks:
Fix #4134 . New props attribute allowTyping added for calendar component for this specific functionality. Set this prop and it will allow the user to type the date field also. Please review the code. This is preliminary implementation, as this attribute will create conflict with other calendar props feature (Min/Max) for example. Please let me know if this features suits you, then I can proceed to handle the rest of the scenarios also.


KacperMatusiewicz:
Hi, is this issue still active? Is it gonna be merged?

mertsincan:
Thanks a lot for your contribution! but it has more conflicts. I think a new PR can be nice. For now, closed. Thanks a lot for your understanding!

Best Regards,

PR #4547, opened 2023-06-25
fixing select all issue in multiselect component

### LINKED ISSUE #4535: MultiSelect (group) SelectAll checked when all items in first group are selected
### Describe the bug

Create a React MultiSelect with group options, when all the children under the first group are selected, the SelectAll checkbox on the header gets selected as well even though items under other groups are unselected. This issue can be easily reproduced on the Demo page
https://primereact.org/multiselect/
or
https://www.primefaces.org/primereact-v8/multiselect/

### Reproducer

_No response_

### PrimeReact version

8.x and 9.x, all versions

### React version

17.x

### Language

TypeScript

### Build / Runtime

Create React App (CRA)

### Browser(s)

_No response_

1. Navigate to React MultiSelect component demo page, either 8.x or 9.x
https://www.primefaces.org/primereact-v8/multiselect/
or
https://primereact.org/multiselect/
2. Locate the Grouped MultiSelect
3. Click to open the Grouped Multiselect
4. Select all 4 cities under Germany
5. Observe that the SelectAll checkbox is selected. 

### Expected behavior

Since only children under the first city get selected, children under other cities have not been selected, the SelectAll checkbox should not be selected. 
MultiSelect-group

shashicode:
Fixes #4535 
**Issue description:** Select All checkbox was getting auto-selected even when only a single group of items were checked.
Expected behavior: Select All checkboxes should only be selected when all the items in the multi-select groups are selected.

**Root cause:** The for loop was immediately returning while only a single group iteration was finished rather than waiting on all the groups of checkboxes as they all have to be selected for Select All to be checked. 

**Fix applied:** The code block now will not return immediately while looping through the single group, rather it will set a flag once any item from any of the groups (being checked in the loop) is unchecked. and returning finally when all the iteration on each group is finished.

### DIFF
--- a/components/lib/multiselect/MultiSelect.js
+++ b/components/lib/multiselect/MultiSelect.js
@@ -397,17 +397,21 @@ export const MultiSelect = React.memo(
                 const options = visibleOptions.filter((option) => !isOptionDisabled(option));
 
                 if (props.optionGroupLabel) {
+                    let anyOptionLeftUnchecked = false
                     for (let optionGroup of options) {
                         const visibleOptionsGroupChildren = getOptionGroupChildren(optionGroup).filter((option) => !isOptionDisabled(option));
-
-                        return !visibleOptionsGroupChildren.some((option) => !isSelected(option));
+                        
+                        if(visibleOptionsGroupChildren.some((option) => !isSelected(option)) === true) {
+                            anyOptionLeftUnchecked = true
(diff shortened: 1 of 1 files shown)


melloware:
Thanks I revered your flag to make it more readable.  I don't like returning NOT something instead of just something

PR #4550, opened 2023-06-26
fix: chips to have type string instead of any for the value since HTM…

### LINKED ISSUE #4551: Chips: Remove type any on chips value
### Describe the bug

Since the value of the InputElement is string this can just be typed as string for the value of chips.

### Reproducer

_No response_

### PrimeReact version

9.5.0

### React version

18.x

### Language

TypeScript

### Build / Runtime

Create React App (CRA)

### Browser(s)

_No response_

_No response_

### Expected behavior

_No response_

bu3alwa:
…LInputElement will return string

@melloware I tried to make it generic but since the input element is string it was throwing a type error unless you do T extends string. So I just made them strings instead. The any type is problematic as usual and causing type errors for me. 

If you need me to submit an issue I could as well. 

Fix #4551

### DIFF
--- a/components/lib/chips/chips.d.ts
+++ b/components/lib/chips/chips.d.ts
@@ -83,7 +83,7 @@ interface ChipsRemovableOptions {
     /**
      * Current value
      */
-    value: any;
+    value: string;
     /**
      * Current index
      */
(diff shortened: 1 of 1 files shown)


melloware:
Yes please submit an issue as well.

PR #4591, opened 2023-07-01
Feat #4485: add custom header prop to sidebar

### LINKED ISSUE #4485: Sidebar: Customize header

There are [passthroughs](https://primereact.org/sidebar/#pt.sidebar.options) to customize the header. However, these would just pass attributes to the header DOM.
On multiple occasions, it is actually necessary to add React elements to the header:
HeaderCustomization

_No response_

It seems the [problem](https://github.com/primefaces/primevue/issues/1939) was raised in primevue. So, I think that a similar solution might be implemented for primereact.

_No response_

### Additional context

Thank you in advance!

bu3alwa:
Passes new header prop to sidebar as a ReactNode

Fix #4485

### DIFF
--- a/components/doc/sidebar/templatedoc.js
+++ b/components/doc/sidebar/templatedoc.js
@@ -18,6 +18,12 @@ export function TemplateDoc(props) {
         </React.Fragment>
     );
 
+    const customHeader = (
+        <React.Fragment>
+            <h2 style={{ marginBottom: 0 }}>Sidebar</h2>
+        </React.Fragment>
+    );
+
     const code = {
         basic: `
 <Sidebar visible={visible} onHide={() => setVisible(false)} icons={customIcons}>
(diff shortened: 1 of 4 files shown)


components/lib/sidebar/Sidebar.js
```diff
@@ -204,6 +204,10 @@ export const Sidebar = React.forwardRef((inProps, ref) => {
         return null;
     };
 
+    const createHeader = () => {
+        return props.icons ? ObjectUtils.getJSXElement(props.header, props) : null;
```
melloware, line 208: Should this be `return props.header ...` and not `props.icon` ?

components/doc/sidebar/templatedoc.js
```diff
@@ -101,12 +117,11 @@ export default function TemplateDemo() {
         <>
             <DocSectionText {...props}>
                 <p>
-                    Additional content at the header section is provided using the <i>icons</i> property.
+                    Additional content at the header section is provided using the <i>icons</i> and <i>headers</i> property.
```
melloware, line 120:
```suggestion
                    Additional content at the header section is provided using the <i>icons</i> and <i>header</i> property.
```

melloware (changes requested):

components/lib/sidebar/Sidebar.js
```diff
@@ -259,6 +264,9 @@ export const Sidebar = React.forwardRef((inProps, ref) => {
             {
                 className: 'p-sidebar-header'
             },
+            !props.header && {
+                style: { justifyContent: 'flex-end' }
```
melloware, line 268: Instead of an inline style here just create a new CSS class like `p-sidebar-header` and then in this scenario just add another `p-sidebar-header-custom`

bu3alwa: I am not sure what ptm returns all it shows is `{}`. So it's not typed in a helpful manner. I haven't looked into it but I fixed it into a class and keeping it as mergeProps.

melloware (approved):
OK I made one small change to your PR that was easier to change than to explain.

```js
const headerClassName = classNames('p-sidebar-header', {
            'p-sidebar-custom-header': props.header
        });
```

PR #4616, opened 2023-07-06
Fix #4471: Column Resize fix for strict and non-strict mode

### LINKED ISSUE #4471: DataTable: Column resize is not working properly.
### Describe the bug

The resize of the columns of the DataTable component is not working correctly when we have updated to the latest version of the 9.5.0 primereact library.

The error is the same as the example on the official documentation page (https://primereact.org/datatable/)

The error is reproduced in both columnResizeMode.

- In "expand" mode the column place is not correct and also the resize of one column affects the rest.
[error_expand_mode.webm](https://github.com/primefaces/primereact/assets/36630670/3b3494f3-5795-46bb-a751-61f7cb922402)

- In "fit" mode, the resize directly does not work.
[error_fit_mode.webm](https://github.com/primefaces/primereact/assets/36630670/3324f777-d9a6-4d7d-bf3a-ee0032b03d98)

The most curious thing is that if we try to run the code in the sandbox, in both cases, it seems to work correctly in a simplified project.

However, the same thing happens in our project as is happening in the examples on the documentation website.

### Reproducer

_No response_

### PrimeReact version

9.5.0

### React version

18.x

### Language

ES6

### Build / Runtime

Create React App (CRA)

### Browser(s)

_No response_

_No response_

### Expected behavior

_No response_

melloware:
Fix #4471: Column Resize fix for strict and non-strict mode

cc @bu3alwa can you try this?

### DIFF
--- a/components/lib/carousel/Carousel.js
+++ b/components/lib/carousel/Carousel.js
@@ -312,7 +312,7 @@ export const Carousel = React.memo(
         }));
 
         useMountEffect(() => {
-            if (elementRef.current && !attributeSelector.current) {
+            if (elementRef.current) {
                 attributeSelector.current = UniqueComponentId();
                 elementRef.current.setAttribute(attributeSelector.current, '');
             }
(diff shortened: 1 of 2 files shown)


bu3alwa:
Tested and it works

PR #4632, opened 2023-07-09
bug fix #4263

### LINKED ISSUE #4623: AutoComplete: scroll is not working when using down key in grouped mode
### Describe the bug

The scroll does not follow the highlighted item when using the down key to navigate through the suggested list.

P.S. The attached CodeSandbox link is from your documentation page.

### Reproducer

https://codesandbox.io/s/hn7jpp

### PrimeReact version

9.6.0

### React version

18.x

### Language

TypeScript

### Build / Runtime

Vite

### Browser(s)

Chrome 114.0.5735.199(Official Build) (64-bit)

1. Press "a" in the grouped autocomplete.
2. By using the down key to navigate to the bottom of the result panel, you will notice that the scroll is not following the focused item.

### Expected behavior

_No response_

userkks:
Fix #4623 : scroll container was passed wrong

### DIFF
--- a/components/lib/autocomplete/AutoComplete.js
+++ b/components/lib/autocomplete/AutoComplete.js
@@ -446,7 +446,7 @@ export const AutoComplete = React.memo(
         };
 
         const getScrollableElement = () => {
-            return virtualScrollerRef.current ? overlayRef.current.firstChild : overlayRef.current.firstChild;
+            return overlayRef.current.firstChild;
         };
 
         const getOptionGroupLabel = (optionGroup) => {
--- a/components/lib/autocomplete/AutoComplete.js
+++ b/components/lib/autocomplete/AutoComplete.js
@@ -446,7 +446,7 @@ export const AutoComplete = React.memo(
         };
 
         const getScrollableElement = () => {
-            return virtualScrollerRef.current ? overlayRef.current.firstChild : overlayRef.current;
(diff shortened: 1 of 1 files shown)


components/lib/autocomplete/AutoComplete.js
```diff
@@ -446,7 +446,7 @@ export const AutoComplete = React.memo(
         };
 
         const getScrollableElement = () => {
-            return virtualScrollerRef.current ? overlayRef.current.firstChild : overlayRef.current;
+            return virtualScrollerRef.current ? overlayRef.current.firstChild : overlayRef.current.firstChild;
```
melloware, line 449: This makes both sides of the ternary the same and will break other scenarios.  How about...
```suggestion
            return virtualScrollerRef.current || props.optionGroupLabel ? overlayRef.current.firstChild : overlayRef.current;
```

melloware (changes requested):

components/lib/autocomplete/AutoComplete.js
userkks: Hi @melloware, you are right the both side is same for the ternary operator. But the issue is there with all kind of autocompletes not only grouped. Please check my latest changes.

melloware: let me test in the showcase.

melloware: Yep your fix seems to be working for all the showcase examples

melloware (approved):

PR #4633, opened 2023-07-09
Fix #4531: InputNumber keydown/keyup

### LINKED ISSUE #4531: InputNumber: Disable KeyUp/KeyDown default Support

Is there a way to remove default KeyUp/KeyDown event listners from primereact Input Nubmer; Ps: i want event to be propagated and treated by the parent

melloware:
Fix #4531: InputNumber keydown/keyup

Allow user to override key up and keydown to prevent default actions.


components/lib/inputnumber/InputNumber.js
```diff
@@ -448,27 +457,31 @@ export const InputNumber = React.memo(
...
+
+            const key = event.key;
 
-            if (code !== 13) {
+            if (key !== 'Enter') {
                 // to submit a form
                 event.preventDefault();
             }
 
-            const char = String.fromCharCode(code);
```
ulasturann, line 469: This definition is missing and is the cause of the error.

components/lib/inputnumber/InputNumber.js
```diff
@@ -448,27 +457,31 @@ export const InputNumber = React.memo(
...
-            }
         };
 
-        const onInputKeyPress = (event) => {
+        const onInputKeyUp = (event) => {
             if (props.disabled || props.readOnly) {
                 return;
             }
 
-            const code = event.which || event.keyCode;
```
ulasturann, line 462: This definition is missing and is the cause of the error.

ulasturann (commented):
Add the missing definitions to prevent the error.

components/lib/inputnumber/InputNumber.js
melloware: Fixed!

components/lib/inputnumber/InputNumber.js
melloware: Fixed!

PR #4640, opened 2023-07-11
feat: expose sorted and sortOrder properties for sortIcon

### LINKED ISSUE #4639: DataTable: The parameter of the function passed to `sortIcon` has the wrong type
### Describe the bug

When using a function for the `sortIcon` prop of a `<DataTable>`, the function's first parameter, `options`, is of type `IconOptions<DataTable<TValue>>`. This type knows about icons, but not about sorting.

Logging the `options` parameter to console shows it has more props than its type claims. The most important one is `sortOrder`, telling which direction a column is sorted. TS does not know about this, so `options.sortOrder` defaults to type `any`. It should be `0 | 1 | -1`

In addition to `sortOrder`, there is also a property `sorted` which the type does not know about. Its actual type appears to for some reason be `null | true`, instead of `boolean`.

### Reproducer

https://stackblitz.com/edit/vitejs-vite-s3jiwf?file=src%2FApp.tsx

### PrimeReact version

9.6.0

### React version

18.x

### Language

TypeScript

### Build / Runtime

Create React App (CRA)

### Browser(s)

_No response_

Hover `options.sortOrder` and `options.sorted` in the reproducer. They both show type `any`.

Check the console, where both props are logged. They clearly have values.

### Expected behavior

The type should know about `options.sortOrder`, which should have type  `0 | 1 | -1`.

In addition, the type should know about `options.sorted` , which for some reason should have type `null | true`

EmirBoyaci:
Closes #4639

It's my proposal, maybe we have more clear way to do that.

### DIFF
--- a/components/lib/datatable/datatable.d.ts
+++ b/components/lib/datatable/datatable.d.ts
@@ -883,6 +883,8 @@ export interface DataTablePassThroughOptions {
     tooltip?: TooltipPassThroughOptions;
 }
 
+type SortOrder = 1 | 0 | -1 | null | undefined;
+
 /**
  * Defines valid properties in DataTable component. In addition to these, all properties of HTMLDivElement can be used in this component.
  * @group Properties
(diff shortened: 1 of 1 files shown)


EmirBoyaci:
Type definition is wrong place, closing.

PR #4642, opened 2023-07-11
feat: allow additional props from IconType to expose additional props in base object

### LINKED ISSUE #4639: DataTable: The parameter of the function passed to `sortIcon` has the wrong type

EmirBoyaci:
Closes #4639

### DIFF
--- a/components/lib/datatable/datatable.d.ts
+++ b/components/lib/datatable/datatable.d.ts
@@ -1271,7 +1271,7 @@ export interface DataTableProps<TValue extends DataTableValueArray> extends Omit
     /**
      * Icon to display the current sorting status.
      */
-    sortIcon?: IconType<DataTable<TValue & { sortOrder: Omit<SortOrder, 'null' | 'undefined'>; sorted: boolean }>> | undefined;
+    sortIcon?: IconType<DataTable<TValue>, { sortOrder: Omit<SortOrder, 'null' | 'undefined'>; sorted: boolean }> | undefined;
     /**
      * Order to sort the data by default.
      */
--- a/components/lib/datatable/datatable.d.ts
+++ b/components/lib/datatable/datatable.d.ts
@@ -1271,7 +1271,7 @@ export interface DataTableProps<TValue extends DataTableValueArray> extends Omit
     /**
      * Icon to display the current sorting status.
      */
-    sortIcon?: IconType<DataTable<TValue>, { sortOrder: Omit<SortOrder, 'null' | 'undefined'>; sorted: boolean }> | undefined;
+    sortIcon?: IconType<DataTable<TValue>, { sortOrder?: SortOrder; sorted?: boolean }> | undefined;
     /**
      * Order to sort the data by default.
      */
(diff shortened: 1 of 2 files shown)


EmirBoyaci:
Not sure about naming but didn't wanted to give more time to choose what is the best naming for generics, feel free to suggest re-naming.

EmirBoyaci:
> so far I like the look of it. Let me test it out later today!

Sounds good. With adding `AdditionalProps` generic, we also add ability to define custom additional properties for specific cases as like in this.
So, in the future I think we can delete `[key: string]: any;` to make sure all properties exactly matches with the type definitions.

melloware:
You might be right i hopefully will be able to play around with your PR later today to see if I run into any issues.

components/lib/datatable/datatable.d.ts
```diff
@@ -1269,11 +1271,11 @@ export interface DataTableProps<TValue extends DataTableValueArray> extends Omit
     /**
      * Icon to display the current sorting status.
      */
-    sortIcon?: IconType<DataTable<TValue>> | undefined;
+    sortIcon?: IconType<DataTable<TValue>, { sortOrder: Omit<SortOrder, 'null' | 'undefined'>; sorted: boolean }> | undefined;
```
melloware, line 1274: Are we sure it shouldn't be: `{ sortOrder?: SortOrder; sorted?: boolean }` as optional or have you confirmed they ALWAYS have values?

EmirBoyaci: I tried with 3-4 example docs via running server on my local machine and It never returned undefined. But, I am not sure maybe we have some edge case. If you wish, we can defined as `?` to make sure we don't break it.

melloware: Yeah lets do that to be safe?

EmirBoyaci: Done

melloware (approved):

melloware:
I just made one small change and some docs but this is working great!

PR #4663, opened 2023-07-17
Fix #4662: SVG icons allow click event to pass through

### LINKED ISSUE #4662: SVG Icons: Clicking on path prevents click
### Describe the bug

Clicking on a SVG version of the icon doesn't always trigger a `click` event if you click on any painted `path` of the SVG.

Workaround:

```css
svg.p-icon {
  pointer-events: auto;
}

svg.p-icon g {
  pointer-events: none;
}
```

### Reproducer

https://stackblitz.com/edit/react-gu3eb7

### PrimeReact version

9.6.0

### React version

18.x

### Language

TypeScript

### Build / Runtime

Create React App (CRA)

### Browser(s)

_No response_

1. Click on the sort icon in the header.
2. You might have to move your mouse pointer directly over a painted parted of the SVG and keep clicking you will see the click events are "swallowed".

### Expected behavior

Clicking on the SVG icon should pass the click event through to the element below it.

See: https://stackoverflow.com/questions/22483643/svg-still-receives-clicks-even-if-pointer-events-visible-painted/

See: https://stackoverflow.com/questions/28282811/how-can-i-pass-only-clicks-through-a-svg-with-pointer-events

melloware:
Fix #4662: SVG icons allow click event to pass through

### DIFF
--- a/components/lib/iconbase/IconBase.css
+++ b/components/lib/iconbase/IconBase.css
@@ -7,6 +7,14 @@
     animation: p-icon-spin 2s infinite linear;
 }
 
+svg.p-icon {
+    pointer-events: auto;
+}
+
+svg.p-icon g {
+    pointer-events: none;
+}
+
 @-webkit-keyframes p-icon-spin {
     0% {
         -webkit-transform: rotate(0deg);


melloware:
Make this more specific with p-icon.

Issue #4608, opened 2023-07-05
ContextMenu: MenuItem is not displayed when using a custom template


PasBazz:
### Describe the bug

ContextMenuSub is not use `item.template` prop

### Reproducer

https://codesandbox.io/s/primereact-demo-forked-ksxd3y?file=/src/App.js

### PrimeReact version

9.6.0

### React version

18.x

### Language

TypeScript

### Build / Runtime

Create React App (CRA)

### Steps to reproduce the behavior

1. Create a `MenuItem` with a custom template `{ template: <div>Custom template</div>}`
2. Open context menu

PR #4682, opened 2023-07-24
Fix #4669: DataTable restore width fix

### LINKED ISSUE #4669: DataTable: The scrollbar disappears after the size is restored
### Describe the bug

If you increase the table more than the size of the parent element, a scroll appears, but the scroll is not restored after the component is remount

Reproducer

### Reproducer

https://codesandbox.io/s/primereact-demo-forked-y99rhp?file=/src/App.js

### PrimeReact version

9.6.0

### React version

18.x

### Language

TypeScript

### Build / Runtime

Create React App (CRA)

### Browser(s)

_No response_

_No response_

### Expected behavior

_No response_

PasBazz:
### Defect Fixes
Fix <ins>**[#4608](https://github.com/primefaces/primereact/issues/4669)**</ins> The problem arose due to the fact that after restoring the width, a inline style was added to the root element and after that the scroll was lost.

image

### DIFF
--- a/components/lib/datatable/DataTable.js
+++ b/components/lib/datatable/DataTable.js
@@ -382,7 +382,6 @@ export const DataTable = React.forwardRef((inProps, ref) => {
             if (props.columnResizeMode === 'expand' && tableWidthState.current) {
                 tableRef.current.style.width = tableWidthState.current;
                 tableRef.current.style.minWidth = tableWidthState.current;
-                elementRef.current.style.width = tableWidthState.current;
             }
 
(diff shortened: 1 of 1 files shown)


melloware:
Thanks for your contribution. I will test it out!

PasBazz:
I recorded the process of restoring the table width before the fix and afterwards.

**Before correction**:

Datatablet_error

**After correction**:

Datatablet_fix

melloware (approved):

PR #4703, opened 2023-07-30
Fix #3762: TreeSelect: Filtering does not select the nodes correctly

### LINKED ISSUE #3762: TreeSelect: Filtering does not select the nodes correctly
### Describe the bug

Hi,

During my work with TreeFilter + Checkbox selection + filtering of the tree I landed on a bug.

It all works fine until I try to filter the tree for a very specific (unique) value and select it.
Instead of partially selecting all the parents, all the tree nodes are completely selected, that makes the filtering unusable.
When clearing up the filter menu the tree selection gets broken - not just the selected node is displayed as "checked", but all the parents, that have to be actually only "partialChecked".

Expected: to have only one option fully selected.
Actual: the filtered tree nodes are fully selected.

Normal behavior when the filter menu is not being used:

Screen Shot 2022-12-08 at 17 52 14 PM

Behavior when the filter menu is being used (broken selection):

Screen Shot 2022-12-08 at 17 52 24 PM
Screen Shot 2022-12-08 at 17 52 33 PM

### Reproducer

https://codesandbox.io/s/funny-mountain-ylz0s0?file=/src/demo/TreeSelectDemo.js

### PrimeReact version

8.7.3

### React version

17.x

### Language

TypeScript

### Build / Runtime

Create React App (CRA)

### Browser(s)

All

1. Open the TreeFIlter menu
2. Search for a child node by using the filter
3. Clear up the filter menu
4. Expand the tree and verify the tree selection

### Expected behavior

To have only one option fully selected (the actually selected one, not all the parents up the tree).

thejus557:
fix #3762: TreeSelect: Filtering does not select the nodes correctly

https://github.com/primefaces/primereact/assets/38601497/c29ff2a5-2fa7-4c8d-8cf9-2ae198828859


melloware:
I assume you double checked Tree, TreeSelect, and TreeTable as they all re-use this functionality?

thejus557:
Let me check it again! @melloware

components/lib/tree/UITreeNode.js
```diff
@@ -374,26 +374,45 @@ export const UITreeNode = React.memo((props) => {
         }
     };
 
+    const findChildrenByKey = (data, key) => {
```
melloware, line 377: Is there any way to move this function into UITreeNode.js or a Utility class it looks like its copied 3 times?

melloware (changes requested):

thejus557:
I  double checked Tree, TreeSelect, and TreeTable. They are working as expected.

components/lib/utils/FindChildrenByKey.js
```diff
@@ -0,0 +1,15 @@
...
+        } else if (item.children) {
+            const result = findChildrenByKey(item.children, key);
+
+            if (result.length > 0) {
+                return result;
+            }
+        }
+    }
+
+    return [];
```
melloware, line 14: you can probably put this in ObjectUtils.js to reduce class sprawl...

melloware (changes requested):

components/lib/utils/FindChildrenByKey.js
thejus557: Updated

melloware (approved):

PR #4732, opened 2023-08-03
Fix primfaces #4730: Calendar: input field cannot parse year-only format

### LINKED ISSUE #4730: Calendar: input field cannot parse year-only format (e.g. 'yy')
### Describe the bug

When creating a `<Calendar>` component with `view='year'` and `'dateFormat='yy'`, if we type a valid year into the input field, the `e.value` property of the `onChange` event returns null

Internally, `parseDate` fails and throws an error

### Reproducer

https://codesandbox.io/s/primereact-test-forked-p6f5dy

### PrimeReact version

9.6.0

### React version

18.x

### Language

TypeScript

### Build / Runtime

Vite

### Browser(s)

_No response_

1. Click on the Calendar component
2. Type a valid year (e.g. 2023)
3. Observe the console (the onChange event is set up to log `e.value` to the console)
4. Confirm that it always prints out `null`

### Expected behavior

The component should be able to parse the valid year and logs it to console

akshayaqburst:
### Defect Fixes
Fix #4730: Calendar: input field cannot parse year-only format

### DIFF
--- a/components/lib/calendar/Calendar.js
+++ b/components/lib/calendar/Calendar.js
@@ -2373,6 +2373,11 @@ export const Calendar = React.memo(
                 day = 1;
             }
 
+            if (props.view === 'year') {
+                day = 1;
+                month = 1;
+            }
+
             const { dayNamesShort, dayNames, monthNamesShort, monthNames } = localeOptions(props.locale);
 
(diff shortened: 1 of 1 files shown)


melloware (approved):

melloware:
Nice fix!

Issue #4648, opened 2023-07-13
Datatable: Edit Cell update value and then Click to other row


minhtri0411:
### Describe the bug

https://codesandbox.io/s/cdemf3?file=/src/App.js

After upgrade form version 9.3.0 to 9,6.0. I can't update value for cell when I input a new value for cell and then Click to other row.

### PrimeReact version

9.6.0

### React version

18.x

### Language

TypeScript

### Build / Runtime

Create React App (CRA)

t-works:
This behavior (updating or not a cell value) is a bit unpredictable. I have similar issues. I recorded a video that demonstrates this unpredictability using the demo provided by minhtri0411
https://youtu.be/jX8bDjHpZMY

melloware:
Marked as a bug. Thanks for reproducer.

paulgrym:
I am experiencing the same problem. Any chance this will be fixed soon? 🙏

paulgrym:
Hi, I'm not sure this bug is fixed - in version 9.6.2. it still does not work. Can anyone confirm if it should work in that version?

melloware:
It is not. See the milestone on the left says "10.0.0" currently.

PR #4745, opened 2023-08-07
fix: #4648 Datatable: Edit Cell update value and then Click to other row

thejus557:
fix: #4648 Datatable: Edit Cell update value and then Click to other row


melloware (approved):

melloware:
nice fix!

thejus557:
sure @melloware  Will do it going forward!

PR #4760, opened 2023-08-10
fix: #4759 bad behaviour of Dialog component

### LINKED ISSUE #4759: Dialog: Bad behaviour when it's dragged
### Describe the bug

If the user adds a margin to the component (<Dialog class="mx-4"/>), the Dialog component will have unexpected behavior during the dragging process. When the user tries to drag the popup to the left, it moves to the right. The reason for this bug is inside the logic of the calculation coordinates of the popup during the dragging process. We can solve this problem by subtracting the margin-left from leftPos and margin-top from topPos when we calculate dialogRef.current.style.left and dialogRef.current.style.top.

### Reproducer

_No response_

### PrimeReact version

9.6.1

### React version

17.x

### Language

ES6

### Build / Runtime

Next.js

### Browser(s)

Safari(Mac os Ventura 13.4)

1.. Create a new React component and add the Dialog component to the template.
2. Add a button to the template to trigger the popup appearing.
2. Add class "mx-4" to the Dialog component.
3. Click the button to popup appears.
4. After the popup appears try to drag it to the left.
5. The popup moves to the right.

### Expected behavior

If you drag the popup, it should move in the direction of dragging regardless of the margins of this popup.

artemlukashenko:
Bad behaviour of the Dialog component during the process of dragging: https://github.com/primefaces/primereact/issues/4759
Fix #4759

### DIFF
--- a/components/lib/dialog/Dialog.js
+++ b/components/lib/dialog/Dialog.js
@@ -192,9 +192,9 @@ export const Dialog = React.forwardRef((inProps, ref) => {
                 }
             } else {
                 lastPageX.current = event.pageX;
-                dialogRef.current.style.left = leftPos + 'px';
+                dialogRef.current.style.left = leftPos - leftMargin + 'px';
                 lastPageY.current = event.pageY;
-                dialogRef.current.style.top = topPos + 'px';
+                dialogRef.current.style.top = topPos - topMargin + 'px';
             }
 
             props.onDrag && props.onDrag(event);
--- a/components/lib/dialog/Dialog.js
+++ b/components/lib/dialog/Dialog.js
@@ -174,8 +174,9 @@ export const Dialog = React.forwardRef((inProps, ref) => {
             const leftPos = offset.left + deltaX;
             const topPos = offset.top + deltaY;
             const viewport = DomHandler.getViewport();
-            const leftMargin = parseFloat(getComputedStyle(dialogRef.current)['margin-left'])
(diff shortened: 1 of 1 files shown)


artemlukashenko:
Hi! @habubey!
I think that we can solve this problem by changing the Dialog component's logic.

components/lib/dialog/Dialog.js
```diff
@@ -174,18 +174,20 @@ export const Dialog = React.forwardRef((inProps, ref) => {
             const leftPos = offset.left + deltaX;
             const topPos = offset.top + deltaY;
             const viewport = DomHandler.getViewport();
+            const leftMargin = parseFloat(getComputedStyle(dialogRef.current)['margin-left'])
+            const topMargin = parseFloat(getComputedStyle(dialogRef.current)['margin-top'])
```
melloware, line 178: I think we need to look at ` const leftPos = offset.left + deltaX;` how the offset is calculated it should account for margin?

melloware (changes requested):

components/lib/dialog/Dialog.js
melloware, line 178:
```suggestion
            const computedStyle = getComputedStyle(dialogRef.current);
            const leftMargin = parseFloat(computedStyle.marginLeft);
            const topMargin = parseFloat(computedStyle.marginTop);
```

melloware (changes requested):

melloware:
OK I committed a small refactoring to your logic.

Issue #4764, opened 2023-08-11
DataTable: cannot read properties of undefined (reading 'matchMode)


gurkiratkn:
### Describe the bug

Clicking on clear filter throws error: Cannot read properties of undefined (reading matchMode)
![image](https://github.com/primefaces/primereact/assets/103406994/43a62c12-bb6a-474f-87f3-9cc01f33996b)
![image](https://github.com/primefaces/primereact/assets/103406994/5bd17760-7367-4045-ac6e-3cdd439b6c04)

### PrimeReact version

9.6.0

### React version

17.x

### Language

ALL

### Build / Runtime

Create React App (CRA)

melloware:
Do you have a reproducer ?

melloware:
Can't reproduce: https://stackblitz.com/edit/react-wpsjsm

PR #4765, opened 2023-08-11
Update ColumnFilter.js

gurkiratkn:
fix: cannot read properties of undefined (reading matchMode)

### Defect Fixes
https://github.com/primefaces/primereact/issues/4764


melloware:
I don't think this will fix the issue you are having because PrimeReact is JS not TypeScript.

PR #4790, opened 2023-08-16
Fix #4788: Calendar typescript improvements

### LINKED ISSUE #4788: Calendar: CalendarChangeEvent type doesn't represent Calendar onChange event value when set to range
### Describe the bug

When I set the Calendar component to `selectionMode='range'` the types that come through on the `onChange` function `event.value` are: 
`Nullable<string | Date | Date[]>`
However, that means that I am not allowed to do: 
`[ new Date(), null ]`
Which seems to be a valid case when I have just one date selected and close the component:
image
image
(in this example I am updating the parent component onHide of the Calendar component, the currentValue is stored in state until this time)

I would like to suggest that the type `(Date | null)[]` be added as a valid type for the Calendar component onChange event. 

### Reproducer

_No response_

### PrimeReact version

9.6.1

### React version

18.x

### Language

TypeScript

### Build / Runtime

Create React App (CRA)

### Browser(s)

_No response_

Build out the Calendar with `range` set as the selectionMode. 
Try to set the default value to something like `[ new Date(), null ]` (to mimic the Calendar in the single date select scenario)

### Expected behavior

Allow me to set defaults of the same type as those values that are used by the Calendar component.
Maybe even separate/default types for the different selectionModes.

melloware:
Fix #4788: Calendar typescript improvements

```ts
import React, { useState } from 'react';
import { Calendar} from 'primereact/calendar';
import { Nullable } from 'primereact/ts-helpers';

export default function RangeDemo() {
  const [dates, setDates] = useState<Nullable<(Date | null)[]>>(null);

  return (
    <div className="card flex justify-content-center">
      <Calendar
        value={dates}
        onChange={(e) => setDates(e.value)}
        selectionMode="range"
        readOnlyInput
      />
    </div>
  );
}
```

### DIFF
--- a/components/lib/calendar/calendar.d.ts
+++ b/components/lib/calendar/calendar.d.ts
@@ -348,6 +348,7 @@ export interface CalendarYearOptions {
     selectable: boolean;
 }
 
+/**
  * Custom month change event.
  * @see {@link CalendarProps.onMonthChange}
  * @event
--- a/components/lib/calendar/calendar.d.ts
+++ b/components/lib/calendar/calendar.d.ts
@@ -8,12 +8,12 @@
  *
  */
 import * as React from 'react';
+import { ButtonPassThroughOptions } from '../button/button';
 import { CSSTransitionProps } from '../csstransition';
+import { InputTextPassThroughOptions } from '../inputtext/inputtext';
 import { TooltipOptions } from '../tooltip/tooltipoptions';
 import { FormEvent, Nullable } from '../ts-helpers';
 import { IconType, PassThroughType } from '../utils';
-import { ButtonPassThroughOptions } from '../button/button';
-import { InputTextPassThroughOptions } from '../inputtext/inputtext';
 
 export declare type CalendarPassThroughType<T> = PassThroughType<T, CalendarPassThroughMethodOptions>;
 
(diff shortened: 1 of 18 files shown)


melloware:
cc @bu3alwa

components/lib/calendar/calendar.d.ts
```diff
@@ -269,19 +269,11 @@ export interface CalendarState {
...
+    viewDate: Nullable<Date>;
 }
 
-/**
- * Custom change event.
- * @see {@link CalendarProps.onChange}
- * @extends {FormEvent}
- * @event
- */
-interface CalendarChangeEvent extends FormEvent<Date | Date[] | string> {}
```
bu3alwa, line 283: Was string an improper type in this component?

components/lib/calendar/calendar.d.ts
```diff
@@ -881,6 +858,62 @@ export interface CalendarProps {
     unstyled?: boolean;
 }
 
+interface CalendarPropsSingle extends CalendarBaseProps {
+    /**
+     * Specifies the selection mode.
```
bu3alwa, line 863: Maybe just minor comment improvements on these

components/lib/calendar/calendar.d.ts
melloware: I could not find anywhere a String was used to represent the `value` anywhere in the code unless I am blind and missing something?

PR #4796, opened 2023-08-17
Fix #4030: Changing Calendar input (with mask) clears the rest of the form's state

### LINKED ISSUE #4030: Calendar: Changing Calendar input (with mask) clears the rest of the form's state
### Describe the bug

Our users noticed when manually typing in a date in a `Calendar` input box, the rest of the state of the form gets cleared out.

For example, the codesandbox link has 2 inputs. One is text, and one is a `Calendar`. If you have some text in the `InputText`, then start typing in the `Calendar`, your `InputText` gets cleared out.

One thing to notice is all the inputs' states are wrapped in an object, rather than each being their own `useState`:
`var [form, setForm] = useState({ text: "", date: undefined });`

Another important note: This problem only manifests when using the `mask` attribute on the `Calendar`. If you remove `mask` this problem goes away and behavior is as expected.

### Reproducer

https://codesandbox.io/s/calendar-mask-problem-qxungp?file=/src/index.js

### PrimeReact version

8.7.3

### React version

18.x

### Language

ALL

### Build / Runtime

Create React App (CRA)

### Browser(s)

_No response_

Inside the codesandbox example: 
1. Simply type something into the first input text box
2. Then tab or click into calendar input box
3. Start typing a date by pressing a number (like 1)
4. Notice the text you initially entered in first text box is deleted

### Expected behavior

I expect the text in the InputText to not be deleted.

cxo05:
Fixes #4030 

Using the mask causes stale closure of the calendar onChange function.

Added useRef for the onChange function. Not sure if this is the best way but it works.


melloware:
Let me review this. I think there MIGHT be a better way?   But really appreciate you helping put time in on this one!

melloware (approved):

melloware:
Just tested this and it works great.  Yeah it has to do with Mask has an underlying events getting calls that switching this to a ref prevents the issue.

components/lib/calendar/Calendar.js
```diff
@@ -2517,6 +2518,10 @@ export const Calendar = React.memo(
             }
         });
 
+        React.useEffect(() => {
+            onChangeRef.current = props.onChange;
```
melloware, line 2522:
```suggestion
            // see https://github.com/primefaces/primereact/issues/4030
            onChangeRef.current = props.onChange;
```

melloware (changes requested):

melloware:
Thanks for the PR. I tested this and it fixes the issue and normal scenarios also working properly.

PR #4798, opened 2023-08-18
Fix #4797: ObjectUtils.resolveFieldData perf improvement

### LINKED ISSUE #4797: Core: ResolveFieldData Performance Improvement
Original Report: https://github.com/primefaces/primevue/issues/4296

The `resolveFieldData` utility currently uses `Object.keys` as its first check before trying to return data. The call to `Object.keys` can get quite expensive and result in bad performance. This shows when its called for a large data set for a large number of components.

image

melloware:
Fix #4797: ObjectUtils.resolveFieldData perf improvement

### DIFF
--- a/components/lib/utils/ObjectUtils.js
+++ b/components/lib/utils/ObjectUtils.js
@@ -57,6 +57,15 @@ export default class ObjectUtils {
     }
 
     static resolveFieldData(data, field) {
+        try {
+            const value = data[field];
+
+            if (value) return value;
+        } catch {
+            // Performance optimization: https://github.com/primefaces/primereact/issues/4797
+            // do nothing and continue to other methods to resolve field data
+        }
+
         if (data && Object.keys(data).length && field) {
             if (this.isFunction(field)) {
                 return field(data);


components/lib/utils/ObjectUtils.js
```diff
@@ -57,6 +57,15 @@ export default class ObjectUtils {
     }
 
     static resolveFieldData(data, field) {
+        try {
+            const value = data[field];
+
+            if (value) return value;
```
mertsincan, line 63: @melloware I dont understand this check. The value can be null, undefined, 0 etc??

melloware: @mertsincan I had to stare at it as well but its actually really clever.  What this does is try to get the `data[field]` and if the `value` is anything Truthy it returns as its found the field.  If it is `falsy` or throws an error it falls back to the original logic.   

So while this may seem like an extra call to `data[field]` and if its `0` if falls through to the default logic where it may evaluate again to `0` correctly its worth it all to avoid calling `Object.keys`.  So this covers about 85% of the time the field is found right away with no call to `Object.keys` and 15% of the time when its `falsy` it will fall back to original logic.  to me this performance improvement is worth it for the 85% scenario especually as the user pointed out if your JSON object is very large and has hundreds of keys.

PR #4818, opened 2023-08-24
PR4814: fix deeply nested replace on row edit

### LINKED ISSUE #4814: DataTable: editorCallback does not update a value if the field name of column has nested elements
### Describe the bug

Hi primereact team! Hope you are doing well! I'm working with a DataTable in a row edit mode. I have for each row an object with a nested array. I'm able to show the data by accessing the data like parentObject.innerArray.0.value. But when I update the value it adds the new field to initial object with a name "parentObject.innerArray.0.value" and not updates the actual nested array.

### Reproducer

https://codesandbox.io/s/delicate-tdd-7w9zmg?file=/src/demo/DataTableEditDemo.js

### PrimeReact version

9.5.0

### React version

18.x

### Language

TypeScript

### Build / Runtime

Next.js

### Browser(s)

_No response_

Please see the codesanbox.

### Expected behavior

I expect the field name of the column to be considered as a path and not a name.

bu3alwa:
Fix #4814 issue with deeply nested object

@Tsareff feel free to test it

### DIFF
--- a/components/lib/datatable/BodyCell.js
+++ b/components/lib/datatable/BodyCell.js
@@ -271,15 +271,14 @@ export const BodyCell = React.memo((props) => {
     const editorCallback = (val) => {
         let editingRowData = { ...editingRowDataState };
 
-        editingRowData[field] = val;
-
+        ObjectUtils.mutateFieldData(editingRowData, field, val);
         setEditingRowDataState(editingRowData);
 
         // update editing meta for complete methods on row mode
         const currentData = getEditingRowData();
 
         if (currentData) {
-            currentData[field] = val;
+            ObjectUtils.mutateFieldData(currentData, field, val);
         }
     };
 
--- a/components/lib/utils/ObjectUtils.js
+++ b/components/lib/utils/ObjectUtils.js
@@ -416,4 +416,55 @@ export default class ObjectUtils {
 
         return [];
     }
+
+    /**
+     * This function takes mutates and object with a new value given
+     * a specific field. This will handle deeply nested fields that
+     * need to be modified or created.
+     *
+     * e.g:
+     * data = {
+     *  nested: {
+     *      foo: "bar"
+     *  }
+     * }
+     *
+     * field = "nested.foo"
+     * value = "baz"
+     *
+     * The function will mutate data to be
+     * e.g:
+     * data = {
+     *  nested: {
+     *      foo: "baz"
+     *  }
+     * }
+     *
+     * @param {object} data the object to be modified
+     * @param {string} field the field in the object to replace
+     * @param {any} value the value to have replaced in the field
+     */
+    static mutateFieldData(data, field, value) {
+        if (typeof data !== 'object' || typeof field !== 'string') {
+            // short circuit if there is nothing to resolve
+            return;
+        }
+
+        const fields = field.split('.');
+        let obj = data;
+
+        for (var i = 0, len = fields.length; i < len; ++i) {
+            // Check if we are on the last field
+            if (i + 1 - len === 0) {
+                obj[fields[i]] = value;
+                break;
+            }
+
+            if (!obj[fields[i]]) {
+                obj[fields[i]] = {};
+            }
+
+            obj = obj[fields[i]];
+        }
+    }
 }
--- a/components/lib/utils/utils.d.ts
+++ b/components/lib/utils/utils.d.ts
@@ -103,6 +103,7 @@ export declare class ObjectUtils {
     static equals(obj1: any, obj2: any, field: string): boolean;
     static deepEquals(a: any, b: any): boolean;
     static resolveFieldData(data: any, field: string): any;
+    static mutateFieldData(data: object, field: string, value: any): void;
     static findDiffKeys(obj1: any, obj2: any): object;
     static reorderArray(value: any, from: number, to: number): void;
     static findIndexInList(value: any, list: any[], dataKey?: string): number;
@@ -137,7 +138,7 @@ export declare class ObjectUtils {
 }
 
 /**
- * Icon utlities for managing icon tasks.
+ * Icon utilities for managing icon tasks.
  */
 export declare class IconUtils {
     static getJSXIcon(icon: IconType<any>, iconProps: React.HTMLProps<HTMLElement>, options: any): any;
@@ -150,7 +151,7 @@ export declare class IconUtils {
 export declare function UniqueComponentId(prefix?: string): string;
 
 /**
- * ZIndex utlities for managing zindex states of different types.
+ * ZIndex utilities for managing zindex states of different types.
  */
 export declare namespace ZIndexUtils {
     export function get(el?: HTMLElement): number;


bu3alwa:
@melloware can't add you as a reviewer but let me know if this is sufficient for a fix.

tsarev-sergei-ahold:
Hi @bu3alwa! Thanks for such a quick response. Unfortunately, I receive an error "Uncaught TypeError: "properyName" is read-only". It traces to this new mutation method you've added.

Tsareff:
Hi @bu3alwa! Thanks for such a quick response! Unfortunately, I receive an error "Uncaught TypeError: "somePropertyName" is read-only". It traces to this new mutation method you've added.

bu3alwa:
> Hi @bu3alwa! Thanks for such a quick response! Unfortunately, I receive an error "Uncaught TypeError: "somePropertyName" is read-only". It traces to this new mutation method you've added.

Thanks for testing it. What I tested yesterday didn't have anything that's read-only. I am wondering what is read-only that is causing this issue. 

Are you using the same example from the reproducer link or a different one?

Tsareff:
> > Hi @bu3alwa! Thanks for such a quick response! Unfortunately, I receive an error "Uncaught TypeError: "somePropertyName" is read-only". It traces to this new mutation method you've added.
> 
> Thanks for testing it. What I tested yesterday didn't have anything that's read-only. I am wondering what is read-only that is causing this issue.
> 
> Are you using the same example from the reproducer link or a different one?

No, I'm using it in my project as a yalc linked library. As far as I understand it can't perform a mutation on currentData. Moreover, line 269 `(editingRowData = { ...editingRowDataState })` also was with an error. I was able to fix it with full copy `const editingRowData = JSON.parse(JSON.stringify(editingRowDataState));`.

bu3alwa:
Thanks I'll take a look soon. The `json.parse` method generally gets into more complications than its worth. And I think you lose `Date` as well. I'll do a clone instead of mutating the object.

bu3alwa:
ok so I am having a hard time reproducing the error you got. Can you provide and example of how you are passing the data to the column? The example you had before works well and it sound like s read only property in the object is being sent to the prop.

Tsareff:
> ok so I am having a hard time reproducing the error you got. Can you provide and example of how you are passing the data to the column? The example you had before works well and it sound like s read only property in the object is being sent to the prop.

Hi @bu3alwa! Thank you for the feedback. Let me double-check my data then.

components/lib/utils/ObjectUtils.js
```diff
@@ -416,4 +416,25 @@ export default class ObjectUtils {
...
+                obj[_prop[0]] = {};
+            }
+
+            if (_prop.length === 1) {
+                return (obj[_prop[0]] = value);
+            }
+
+            this.mutateObjectProperty(_prop.slice(1).join('.'), value, obj[_prop[0]]);
+        }
+    }
```
melloware, line 439: @bu3alwa isn't this method similar to what `ObjectUtils.resolveFieldData` is doing?

melloware (changes requested):

components/lib/utils/ObjectUtils.js
bu3alwa: Similar but my function mutates the object vs this creates a new `object | null` . And because of this line 

https://github.com/primefaces/primereact/pull/4818/files#diff-5c774f14f0f649e0edd947820d6ab8bf8ec6605b70edd15a4e18feb77e422f1cL282

I haven't used `resolveFieldData` I guess. It looks like its changing the object by reference and I haven't dug deep enough to figure out why that is or if its even needed. but I assumed its correct and created that function to mutate deeply nested properties.

Let me know if you think there is a better approach.

melloware: I have to study the problem more but it looks like resolveFieldData is exactly what they use for nested properties so `const value = ObjectUtils.resolveFieldData(json, "your.nested.prop");` would traverse the props until it gets the value?

melloware: Oh i see this is on editing.  Maybe call it `mutateFieldData` to be consistent and it looks like you can reuse a lot of that code?

components/lib/utils/ObjectUtils.js
```diff
@@ -416,4 +416,25 @@ export default class ObjectUtils {
 
         return [];
     }
+
+    static mutateObjectProperty(prop, value, obj) {
```
melloware, line 420:
```suggestion
    static mutateFieldData(data, field, value) {
```

Would keep the signature and naming consistent with this class?

bu3alwa: Makes sensee

components/lib/utils/ObjectUtils.js
bu3alwa: Not sure if that works unless I'm blind. I don't see any recursion there. If it's nested `N` deep the two for loops won't go deeper than depth 2.

melloware: in resolveFieldData it keeps looping over the split "." until it finds a value no?

```js
let fields = field.split('.');
                let value = data;

                for (var i = 0, len = fields.length; i < len; ++i) {
                    if (value == null) {
                        return null;
                    }

                    value = value[fields[i]];
                }
                ```

Comment (bu3alwa) [reply components/lib/utils/ObjectUtils.js:439]: Oh ok so I was blind I'll update it tonight or tomorrow then.

**Comment by bu3alwa** [comment]
ok, tested and its working with the new implementation. 

One thing to note that should be a separate issue when you are on the price in row edit mode example. Typing a number triggers twice when you type. There is some useEffect somewhere that's triggering twice on strict mode. Should probably be investigated on a separate ticket. My suspicion is its in InputNumber.

File: components/lib/utils/ObjectUtils.js [hunk 3]
```diff
@@ -416,4 +416,31 @@ export default class ObjectUtils {
...
+        if (typeof data !== 'object' || typeof field !== 'string') {
+            // short circuit if there is nothing to resolve
+            return;
+        }
+
+        let fields = field.split('.');
+        let obj = data;
+
+        for (var i = 0, len = fields.length; i < len; ++i) {
+            // console.log('fields',fields)
```
melloware, line 431: Remove these console.log comments

components/lib/utils/ObjectUtils.js
```diff
@@ -416,4 +416,31 @@ export default class ObjectUtils {
 
         return [];
     }
+
+    static mutateFieldData(data, field, value) {
+        // console.log(data,field,value)
```
melloware, line 421: Rmeove console.log

components/lib/utils/utils.d.ts
```diff
@@ -134,6 +134,7 @@ export declare class ObjectUtils {
     static findLast(value: any[], callback: () => any): any;
     static findLastIndex(value: any[], callback: () => any): number;
     static sort(value1: any, value2: any, order: number, locale: string | string[]): number;
+    static mutateFieldData(data: object, field: string, value: any): void;
```
melloware, line 137: Move this line up so its right next to `resolveFieldData` please.

melloware (changes requested):

components/lib/datatable/BodyCell.js
```diff
@@ -271,15 +271,14 @@ export const BodyCell = React.memo((props) => {
     const editorCallback = (val) => {
         let editingRowData = { ...editingRowDataState };
```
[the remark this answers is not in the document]
bu3alwa: I am not sure, just looking at the component in the page it seems to get nested data and when you edit it modifies the nested data. So I am assuming the implementation there works correctly but different from how `datatable` was implemented.

components/lib/utils/ObjectUtils.js
```diff
@@ -416,4 +416,31 @@ export default class ObjectUtils {
 
         return [];
     }
+
```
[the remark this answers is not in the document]
bu3alwa: There are some misspelling of `Utilities` in this file should I fix those as well in the same PR or is that too messy?

components/lib/utils/ObjectUtils.js
bu3alwa: Sorry I should have checked before I pushed

components/lib/utils/ObjectUtils.js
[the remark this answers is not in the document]
melloware: Spelling mistakes are usually ok to fix!  Weird that treetable works with no changes?

melloware (approved):

components/lib/utils/ObjectUtils.js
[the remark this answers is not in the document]
bu3alwa: I assume that the component edits only the object on its level and the `child` inside the object is recursively passed into the component so edit would always be editing depth 1

Tsareff:
Hi @bu3alwa! I checked one more time and the problem with read-only is indeed on my side. I found also another tricky thing. When I create Columns from array of arrays, the method `onEditChange` can't find `editingRowIndex` and row is always focused. In my case I just need 28 number inputs in one row, 4 inputs for each day of the week, that is why I'm using such a rendering.

Tsareff:
I've tried also to do it manually running this app locally, not my own. If you add several Columns which should edit the same object `<Column field="meta.1.code" header="Code" editor={(options) => numberEditor(options)} style={{ width: '20%' }}></Column>
                    <Column field="meta.1.code1" header="Code 1" editor={(options) => numberEditor(options)} style={{ width: '20%' }}></Column>` in the `onEditChange` method in`BodyRow` component `props.editingRows` will have an updated object, but `e.data` the previous version of it. In this case the row will be always in focused mode.

Tsareff:
Sorry, one more condition for an error. It happens when you have custom `onRowEditChange` callback. In this case the pre-built logic is not working, so I need handle the `editingRows` state like this
`    const onEditChange = ({ data = [] }) => {
        setEditingRows([...data].slice(-1));
    }`

Tsareff:
> To be clear are we talking about a new issue or is this part of the same issue?
> 
> And if you could please provide a example link so I can understand the problem better and use it to test locally.

I really can't tell for sure if it connected to this problem. For me it is, since I cannot perform update if the structure of the table consist of several rows which are getting data from nested object. 

Could you please tell me how to create a demo example using the PR code? I mean is it even possible to use codesandbox or smth similar?

bu3alwa:
> > To be clear are we talking about a new issue or is this part of the same issue?
> > And if you could please provide a example link so I can understand the problem better and use it to test locally.
> 
> I really can't tell for sure if it connected to this problem. For me it is, since I cannot perform update if the structure of the table consist of several rows which are getting data from nested object.
> 
> Could you please tell me how to create a demo example using the PR code? I mean is it even possible to use codesandbox or smth similar?

I am not sure with a specific PR but I just need it as an example code and I can copy and paste it to my local build.

Tsareff:
Hi @bu3alwa! Could you please tell me when this fix can be expected to be released?

bu3alwa:
@Tsareff I'm not an actual maintainer on here so can't give you any estimates

melloware:
Is this fix final I thought you guys were working out more scenarios?

bu3alwa:
@melloware I think the other issues were resolved on @Tsareff end, he can confirm. My PR should be ready to go from my end

melloware (approved):

Tsareff:
Hi @bu3alwa @melloware! Yes, this fix is working, all other things on my side. Maybe I'm looking at a wrong place, but I don't see the new version. Could you please tell me when can I expect to use it?

Issue #4848, opened 2023-08-30
AutoComplete: can't remove item


TheArcherST:
### Describe the bug

Can't remove item from AutoComplete component, added due external event, if this action was performed two times.

Context: I want to implement automatically undo of actions during filling a data table. Table uses this component as body for some columns. After canceling of the last action (cancel element removing), I can't remove it again by mouse click.

### Reproducer

https://codesandbox.io/s/primereact-test-forked-dvsf96?file=/src/index.js

### PrimeReact version

9.6.2

### React version

18.x

### Language

TypeScript

### Build / Runtime

Create React App (CRA)

### Steps to reproduce the behavior

1. Press on `Add item` button.
2. Remove item from autocomplete
3. Press to button `Add item` again
4. Try to remove item.

### Expected behavior

I can remove item.

melloware:
Good reproducer. Definitely looks like a bug.

akshayantony55:
I will work on this @melloware

melloware:
Assigned to you!

PR #4855, opened 2023-09-01
Fix primefaces #4848 AutoComplete: can't remove item

akshayaqburst:
### Defect Fixes
Fix primefaces #4848 AutoComplete: can't remove item

### DIFF
--- a/components/lib/autocomplete/AutoComplete.js
+++ b/components/lib/autocomplete/AutoComplete.js
@@ -152,7 +152,7 @@ export const AutoComplete = React.memo(
                 });
             }
 
-            selectedItem.current = value.length ? value : null;
+            selectedItem.current = value && value.length ? value : null;
         };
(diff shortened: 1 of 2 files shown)


akshayaqburst:
@melloware

components/lib/autocomplete/AutoComplete.js
```diff
@@ -152,7 +152,7 @@ export const AutoComplete = React.memo(
                 });
             }
 
-            selectedItem.current = value;
+            selectedItem.current = value && value.length ? value : null;
```
melloware, line 155:
```suggestion
            selectedItem.current = ObjectUtils.isNotEmpty(value) ? value : null;
```

melloware (changes requested):

melloware (approved):

Issue #4875, opened 2023-09-04
InputNumber: Doesn't allow minus sign in currency mode


VasiliPodoplielov:
### Describe the bug

Hello.
I faced problem in InputNumber when I use this component like this:
`<InputNumber
        mode="currency"
        currency="USD"
/>`
It uses `prefix` in input under the hood. So the problem is related not only to the `currency` mode but also when I use any prefix.

When I populate some value to the field and after that want to make this value negative by adding `-` minus sign - it happens nothing. After some investigation I found out that problem might be related to this part of code:
![image](https://github.com/primefaces/primereact/assets/37293464/f45188c5-af93-496c-9bc1-9a65105e0003)
because `selectionStart=1`, `prefix` has a position 0 and minus sign should be on -1 position.

Another problem, when I have a negative value with prefix inside an input, e.g. `-$123` I cant remove value from input by pressing `backspace`. I need highlight the whole value and after that press `backspace` or `delete` button.

### Reproducer

https://codesandbox.io/s/primereact-demo-forked-pj7tdw?file=/src/App.js

### PrimeReact version

9.6.2

### React version

18.x

### Language

TypeScript

### Build / Runtime

Create React App (CRA)

### Steps to reproduce the behavior

1. Type some number
2. Set cursor at the start of number(before the number)
3. Try press minus sign

### Expected behavior

It should be possible to add minus sign when cursor is at the start(before the number).
It shuold be possible to remove negative number by pressing `backspace`.

akshayaqburst:
@melloware Where can I test the recently merged changes?

melloware:
just do a git pull from `master` and run `npm run dev` it runs the full showcase with the current code.

VasiliPodoplielov:
@melloware 
I cloned master, run `npm install`, run `npm run dev`, open `http://localhost:3000/inputnumber/` and tried type `-` before an existed number.
![image](https://github.com/primefaces/primereact/assets/37293464/1e61273e-5dba-46b5-8c1e-6535777e6fe8)

akshayaqburst:
Seems there is a typo with the suggested code. I will create a new PR soon.

VasiliPodoplielov:
I added `-` and it is between dollar sign and numbers and after `blur` value in field formats and looks as expected.
Another problem - I can add `-` in any place, but after that the field is cleaned.
Case that I described in previos message - is not fixed.
[...]

PR #4881, opened 2023-09-06
Fix primefaces #4875 InputNumber: Doesn't allow minus sign in currency mode

akshayantony55:
### Defect Fixes
Fix primefaces #4875 InputNumber: Doesn't allow minus sign in currency mode

### DIFF
--- a/components/lib/inputnumber/InputNumber.js
+++ b/components/lib/inputnumber/InputNumber.js
@@ -510,7 +510,7 @@ export const InputNumber = React.memo(
         };
 
         const allowMinusSign = () => {
-            return props.min === null || props.min < 0;
+            return props.min === null || rootProps.min === null ||  props.min < 0;
         };
 
         const isMinusSign = (char) => {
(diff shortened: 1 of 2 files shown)


melloware (changes requested):
I am not sure this makes sense as `rootProps.min` will always be NULL right?

akshayaqburst:
@melloware I have slightly adjusted my code. When min is not passed to the InputNumber component, props.min has value undefined and rootProps.min has value of null.

components/lib/inputnumber/InputNumber.js
```diff
@@ -510,7 +510,7 @@ export const InputNumber = React.memo(
         };
 
         const allowMinusSign = () => {
-            return props.min === null || rootProps.min === null || props.min < 0;
+            return props.min === null || !props.min || props.min < 0;
```
melloware, line 513:
```suggestion
            return ObjetUtils.isEmpty(props.min) || props.min < 0;
```

melloware (changes requested):

melloware (approved):

Issue #4919, opened 2023-09-15
Datatable: Reorder+Filter


thejus557:
### Describe the bug

Hey, we should have a flag to prevent the reorder when filter is applied.

### PrimeReact version

9.4.0

### React version

18.x

### Language

TypeScript

### Build / Runtime

Vite

thejus557:
@melloware  If you have any idea on how to prevent the reorder when filter is applied ?.

melloware:
Hmmm I would have look but you are saying once you have filtered the table that re-ordering rows makes no sense right?

melloware:
OK I marked as new feature and for discussion with PrimeTek.  We should also check what PrimeNG or PrimeVue does in case they have already implemented such a feature.

thejus557:
Just checked the primeng, it also has the same issue.

PR #4923, opened 2023-09-16
feat: #4919 datatable: reorder+filter - disable drag and drop when filter is applied

thejus557:
feat: #4919 datatable: reorder+filter - disable drag and drop when filter is applied


melloware (commented):
Just thinking out loud I am wonder if it needs to be a property at all? You said it makes no sense to re-order when filtering?

Assigning PrimeTek to review...

thejus557:
Yes it makes no sense. But i have added it as a property to be more specific and clear on things. If it is not required. I can remove the property.

melloware:
@thejus557 i think you can remove the property as this should just be the way it works.

mertsincan:
Hi,

Thanks a lot for your contribution, but it has more conflicts. So, for now, closed.

Also, could you please create a discussion ticket to collect feedback from other users? If we are going to do this, this could be a prop option on DataTable.

PR #4967, opened 2023-09-25
MultiSelect label now uses props.values instead of props.options

### LINKED ISSUE #4944: Multiselect: Display labels of selected items in filtered component
### Discussed in https://github.com/orgs/primefaces/discussions/88

<div type='discussions-op-text'>

<sup>Originally posted by **mohamadbdeir** September 22, 2023</sup>
I was trying to use a filtered multiselect component with the data being fetched from an API instead of hardcoded.
 In my code, typing in the filter field performs an API call and fetches data to be displayed in the options.
However, I have noticed that when the options are filtered, if a selected option is filtered out so no longer available in the list of options to select, the label is showing 'null' instead of the object field. I came to the conclusion that the labels are being fetched from the list of the options instead of the list of selected items.
I think it would be better to display labels based on the list of already selected items.

opera_HldnZ3GoaN
</div>

mohamadbdeir:
### Defect Fixes
Fix #4944 

Here is the [issue](https://github.com/primefaces/primereact/issues/4944) I submitted  as well as the relevant [discussion](https://github.com/orgs/primefaces/discussions/88)  #4944 

### Feature Requests

### DIFF
--- a/components/lib/multiselect/MultiSelect.js
+++ b/components/lib/multiselect/MultiSelect.js
@@ -375,7 +375,7 @@ export const MultiSelect = React.memo(
                         }
                     }
                 } else {
-                    option = findOptionByValue(val, props.options);
+                    option = findOptionByValue(val, props.value);
                 }
             }
 


melloware (approved):
I tested this locally and nothing seems broken. I assume you tested it with your dynamic use case?

mohamadbdeir:
> I tested this locally and nothing seems broken. I assume you tested it with your dynamic use case?

Yes I tested with my use case and it works

melloware:
@mohamadbdeir found a bug with your implementation: https://github.com/primefaces/primereact/issues/4988

PR #5081, opened 2023-10-11
Fix Tree copy by doing a shallow copy of the value

### LINKED ISSUE #5080: Tree: onDrop should not attempt a copy of node values
### Describe the bug

`onDrop` does a `JSON.parse(JSON.stringify(props.value))` to copy nodes. I assume this has to do with reference equality. This has a few issues when implementing drag'n'drop:

1. Information is lost. As values are serialized/deserialized they do not always serialize back into the same object. Some cases of this include Date, Map, Set, Regexp, BigInt, Infinity, functions and custom classes
2. Not all values are serializable. Circular structures cannot be serialized to JSON, thus throwing a `TypeError`

### Reproducer

https://codesandbox.io/s/primereact-test-forked-p5vxlv?file=/src/index.js

### PrimeReact version

10.0.3

### React version

17.x

### Language

ALL

### Build / Runtime

Create React App (CRA)

### Browser(s)

_No response_

1. Open sandbox
2. Drag one node to another
3. Observer circular reference TypeError

2. Comment out node 3
3. Drag one node to another
4.  Observe console.log with different data than the original node

### Expected behavior

Drag'n'drop should not edit node data.

I'm willing to add a PR for this. I made a small test and just copying the `values` array seems to work fine. Is this expected? Or should a more deeper/recursive copy be done?

hugo-vrijswijk:
Fixes #5080

This does a _shallow_ clone of the value, instead of a deep clone. This way data is not modified. The copy is only 1 level deep. In testing, this worked fine. But someone with more knowledge of the code should review and decide if a deeper copy (excluding data) is needed.

### DIFF
--- a/components/lib/tree/Tree.js
+++ b/components/lib/tree/Tree.js
@@ -58,9 +58,19 @@ export const Tree = React.memo(
             dragState.current = null;
         };
 
+        const copyValue = (value) => {
+            if (value === null || value === undefined) {
+                return value;
+            } else if (Array.isArray(value)) {
+                return Array.from(value);
+            } else {
+                return { ...value };
+            }
+        };
+
         const onDrop = (event) => {
             if (validateDropNode(dragState.current.path, event.path)) {
-                let value = JSON.parse(JSON.stringify(props.value));
+                const value = copyValue(props.value);
                 let dragPaths = dragState.current.path.split('-');
 
                 dragPaths.pop();
@@ -89,7 +99,7 @@ export const Tree = React.memo(
 
         const onDropPoint = (event) => {
             if (validateDropPoint(event)) {
-                let value = JSON.parse(JSON.stringify(props.value));
+                const value = copyValue(props.value);
                 let dragPaths = dragState.current.path.split('-');
 
                 dragPaths.pop();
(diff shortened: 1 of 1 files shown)


melloware:
Hmm I am not about doing just a shallow copy.  I see your issue with the circular reference but aren't you better off modying your model object being used in the treenode that to do this?

hugo-vrijswijk:
I could write a deep clone that leaves the `data` property alone. That way it's still recursive but would solve my issue and for anyone using non-serializable data. What do you think? 
I unfortunately can't easily change my code to only support serializable data.

melloware:
OK yeah i would do a deep copy without the `data`  attribute as a start.

melloware:
something like this..

```js
function deepCopyWithoutData(obj) {
  if (obj === null || typeof obj !== 'object') {
    return obj;
  }

  if (Array.isArray(obj)) {
    return obj.map(item => deepCopyWithoutData(item));
  }

  const result = {};
  for (let key in obj) {
    if (key !== 'data') {
      result[key] = deepCopyWithoutData(obj[key]);
    }
  }

  return result;
}
```

hugo-vrijswijk:
Thanks, I made a sort of similar implementation that handles some edge-cases. JavaScript is messy 😅

hugo-vrijswijk:
@melloware Is this PR good to merge like this? I've tested it both in the docs and my own project and it works on both

melloware (approved):

PR #5121, opened 2023-10-19
Fix #4181 -AvatarGroup: Avatar Hover tooltip

### LINKED ISSUE #4181: AvatarGroup -> Avatar Hover tooltip
### Describe the bug

When hovering the AvatarGroup - Avatar, the tooltip sets to 100% of the viewport, sometimes for a flash, other times it stays like that.

This is an image for what happened to me
Captura de ecrã 2023-03-24 094510

### Reproducer

https://stackblitz.com/edit/react-55hj5g-jskkgn?file=src%2Fdemo.js

### PrimeReact version

9.2.1

### React version

18.x

### Language

TypeScript

### Build / Runtime

Next.js

### Browser(s)

_No response_

Hover the avatars and the bug will flash sometimes and in other cases stay 

### Expected behavior

_No response_

buraksaglam089:
Fix #4181 .

### DIFF
--- a/components/lib/tooltip/Tooltip.js
+++ b/components/lib/tooltip/Tooltip.js
@@ -149,8 +149,10 @@ export const Tooltip = React.memo(
                 if (isAutoHide()) {
                     elementRef.current.style.pointerEvents = 'none';
                 }
+                
+                const mouseTrackCheck = isMouseTrack(currentTargetRef.current) || position === 'mouse';
 
-                if (((isMouseTrack(currentTargetRef.current) || position == 'mouse') && !containerSize.current) || (isMouseTrack(currentTargetRef.current) || position == 'mouse')) {
+                if ((mouseTrackCheck && !containerSize.current) || mouseTrackCheck) {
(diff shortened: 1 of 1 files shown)


components/lib/tooltip/Tooltip.js
```diff
@@ -150,7 +150,7 @@ export const Tooltip = React.memo(
                     elementRef.current.style.pointerEvents = 'none';
                 }
 
-                if ((isMouseTrack(currentTargetRef.current) || position == 'mouse') && !containerSize.current) {
+                if (((isMouseTrack(currentTargetRef.current) || position == 'mouse') && !containerSize.current) || (isMouseTrack(currentTargetRef.current) || position == 'mouse')) {
```
melloware, line 153: Since we are calling isMouseTrack method twice here break it out to a const on the line above.

melloware (changes requested):

habubey (approved):

melloware (approved):

PR #5215, opened 2023-11-01
Added custom classNameMergeFunction configuration to ptOptions

### LINKED ISSUE #5144: ConfirmDialog pt prop style doesn't work with tailwind classes

### LINKED ISSUE #5564: Calendar input pt
### Describe the bug

Hey there,

Not sure if it is a bug, but I can’t get the calendar input pt work.

i tried to copy the basic tailwind pt from documentation, but I get the fn can’t apply error.

`input: {
            root: ({ props }) => ({
                className: classNames('font-sans text-base text-gray-600 dark:text-white/80 bg-white dark:bg-gray-900 p-3 border border-gray-300 dark:border-blue-900/40 transition-colors duration-200 appearance-none', 'hover:border-blue-500', {
                    'rounded-lg': !props.showIcon,
                    'border-r-0 rounded-l-lg': props.showIcon
                })
            })
        }`

Withouth the props it’s working, but I would like to apply a style for disabled option.

Thanks

### Reproducer

https://stackblitz.com/edit/react-ceahef?file=src%2FApp.js

### PrimeReact version

10.2.1

### React version

18.x

### Language

ES5

### Build / Runtime

Next.js

### Browser(s)

_No response_

_No response_

### Expected behavior

_No response_

### LINKED ISSUE #5570: Panel: Passthrough not working for header
### Describe the bug

The headerTemplate is not receiving the headerClassname specified in the passthrough

### Reproducer

https://codesandbox.io/p/sandbox/primereact-bug-headerclassname-not-included-in-panel-8s4xgf

### PrimeReact version

10.2.1

### React version

17.x

### Language

TypeScript

### Build / Runtime

Vite

### Browser(s)

_No response_

1. Create a <Panel />
2. Use templateHeader for the panel
3. Change the headerClassname

### Expected behavior

headerTemplate should receive the classnames specified in the passthrough

bweis:
This is the next iteration on #5209

Fix #5144 
Fix #5564
Fix #5570

### New Context

Some CSS Class systems require a careful merging of classnames since there is no way to know what classnames have higher priority.  This new configuration option will allow consumers of the library to assume the responsibility for merging these classnames.  (ie: tailwind-merge for Tailwind)

After [discussing in the discord](https://discord.com/channels/787967399105134612/787968635414118421/1169264043056447538), we determined this composable solution would be a better choice as to not constrain ourselves to tailwind support.  

### Problem that this is solving

This is an issue in broader CSS. But when consuming a component lib its not of large impact generally because you would have the component lib's CSS defined before your local, so your local CSS would take priority.  However, when using tailwind this is not the case and is well documented by tailwind itself.  Because of this many tailwind users rely on tailwind merge to resolve these conflicts.

Copying the preset and customizing it to your own liking is an partial option, however this does not solve the use case in which you either use the tailwind preset or your own and then later in your application want to customize a component using the pt={} props.  This will never work with tailwind classes.  

If a user uses the Tailwind preset or a custom PT config at the provider.  But then wants to override a class of a specific component, this will not be possible for them to do with the existing APIs.

 This change aims to fix that and is necessary for proper tailwind (or other css library) support.

As it stands before this PR, the paginator's bg color can never be reliably changed:

```javascript
const customPT = {
  dropdown: {
      root: { className: 'bg-blue-500' },
    },
}
return (
  <PrimeReactProvider value={{ unstyled: true, pt: customPT}>
    <Paginator 
      pt={{
        RPPDropdown: { root: { className: 'bg-red-500' } },
      }}
    />
  </PrimeReactProvider>
);
```

With the proposed non-breaking changes in this PR you will be able to properly use talwind css in primereact.  Without it tailwind support is not reliable for anyone that wants to modify any styling in components themselves beyond the provider passthrough. 

### Notable changes

- The `mergeProps` function is now consumed via a hook `useMergeProps` so that context can be provided automatically. 
- The original `mergeProps` function has been made "package private" as to only be used by the `ComponentBase.js`.
- Documentation has been added for the usage of this new configuration.

### Remaining questions/notes

- the `usePassthrough` "hook" is a misnomer and should really be renamed as `mergePassthrough` since it is not a hook in a traditional sense, more-so a utility function.

### Previous PR Context

> ###
> Previously, when attempting to override a tailwind class, the class merging algorithm would not take into account the priority of the tailwind base classes.  A utility called tailwind-merge does this for you, along with maintaining the merging of standard classes that might be defined.
> 
> This has been brought up in a few other discussions. (See: #5144 and similar) Since it seems that there has been a big push to support tailwind, we decided to just hack away at this. 
> 
> Obviously it would be very nice to not have to drill context into every component like that, however I decided to go that way since the president was set and I didn't want to influence design decisions as a first time contributer.  In the event that you would like to avoid prop drilling, we could create a wrapped hook such as `useMergeProps(props...)` that returns a wrapped around the existing `mergeProps()` function. Ultimately, every file that was touched would still need to be touched.
> 
> ### Note
>  It works 100% if you rebase it on the tip of 10.0.7.  Currently some styling issues exist on the top of master that seem to break components such as `Editor.js`.
>  
> ### Important changes
> - `components/lib/utils/MergeProps.js` now checks for the useTailwind context flag to determine what merging strategy to use.
> - The usePassthrough hook has also been updated to use the new merge functionality when the Provider has been enabled to do so.  This works independently of the per component overrides.
> ```javascript
> const CustomTailwind = usePassThrough(
>   {
>     dropdown: {
>       root: { className: 'w-36 md:w-36 bg-red-500 md:w-64' },
>     },
>   },
>   {
>     mergeSections: true,
>     mergeProps: true,
>     useTailwind: true,
>   },
> );
> ```
> ```javascript
> <PrimeReactProvider value={{ unstyled: true, pt: CustomTailwind, useTailwind: true }}>
>   <AppRouter />
> </PrimeReactProvider>
> ```
> 
> - Documentation has been updated as have the types associated with the change.
> 
> 
> ### Automated CodeMod
> Just as an FYI, incase you want to change the shape of the API, I automated the code change with the following regex:
> FInd `\bmergeProps\(((?:[^;]+)|(?:\n))\);` ->  Replace `mergeProps([$1], { useTailwind: context.useTailwind });`


melloware:
This looks clever to me.

bweis:
Is there any other context that I can add here to move this forward?

melloware (approved):

melloware:
@gucal it would be great if this was included in `10.0.10` or probably be `10.1.0` to be safe with semver as this is a feature a lot of users are asking for.

components/lib/utils/MergeProps.js
```diff
@@ -1,5 +1,22 @@
...
+        const options = {
+            ...(context?.ptOptions?.classNameMergeFunction && { classNameMergeFunction: context.classNameMergeFunction })
+        };
+
+        return _mergeProps(props, options);
+    };
+};
+
+// *
+export function _mergeProps(props, options = {}) {
```
melloware, line 17: @bweis why rename this `_mergeProps` won't this break backwards compatibility for users using this published method?

melloware (changes requested):

components/lib/utils/MergeProps.js
bweis: I figured that this would cause either a minor bump, or be reserved for a major bump.  The `_mergeProps` function fundamentally works without any context.  Conceivably, you would want to either enforce subscribers to the API to use `useMergeProps` "hook".  Which for the record, is not a hook.  If we wanted a real "equivalent" API then `useMergeProps` should also be configured to return a custom `-mergeProps` function with the context at creation to make it actually a hook.  Rather than a simple "getter" utility.

melloware: I totally understand I just worry about public renames of already published methods.

gucal:
Thank you for your effort. 
We will review and get back to you as soon as possible. We plan to complete it at the next or later milestone.

Thanks!

MarcusResell:
This is exactly what would unblock us from moving to version 10 and also utilise unstyled mode with tailwind. Is there any update on the timeline for this? Unfortunate that the release for the css variables were pushed a long with this bottleneck for Tailwind and unstyled mode.  We will unwillingly have to stick with scss themes in our remake

melloware (approved):

melloware:
@bweis any chance you could update the conflicts?  If not I understand since its been a long time to merge this PR but I think I am ready to merge it for 10.4.0 it fixes so many TailWind issues

melloware:
Implemented with: https://github.com/primefaces/primereact/pull/5755

PR #5242, opened 2023-11-05
fix: Primefaces #5241, AutoComplete: Primitive value 0 does not display correctly

### LINKED ISSUE #5241: AutoComplete: Primitive value 0 does not display correctly
### Describe the bug

Dropdown component does not display the correct label when the selected option has a primitive value of 0.

Can also be reproduced on the PrimeReact website - https://primereact.org/autocomplete/#dropdown

### Reproducer

_No response_

### PrimeReact version

10.0.9

### React version

17.x

### Language

TypeScript

### Build / Runtime

Create React App (CRA)

### Browser(s)

_No response_

https://primereact.org/autocomplete/#dropdown

Step 1: Have a AutoComplete with primitive values from 0 to 10
Step 2: Click open autocomplete suggestions
Step 3: Click option with value 0
Step 4: Option is not selected

### Expected behavior

Option 0 should be selected.

akshayaqburst:
### Defect Fixes
fix #5241 AutoComplete: Primitive value 0 does not display correctly

### DIFF
--- a/components/lib/autocomplete/AutoComplete.js
+++ b/components/lib/autocomplete/AutoComplete.js
@@ -159,7 +159,7 @@ export const AutoComplete = React.memo(
         };
 
         const formatValue = (value) => {
-            if (value || value === 0) {
+            if (ObjectUtils.isNotEmpty(value)) {
                 if (typeof value === 'string') {
                     return value;
                 } else if (props.selectedItemTemplate) {
--- a/components/lib/autocomplete/AutoComplete.js
+++ b/components/lib/autocomplete/AutoComplete.js
@@ -159,7 +159,7 @@ export const AutoComplete = React.memo(
         };
 
         const formatValue = (value) => {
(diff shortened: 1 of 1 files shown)


components/lib/autocomplete/AutoComplete.js
```diff
@@ -159,7 +159,7 @@ export const AutoComplete = React.memo(
         };
 
         const formatValue = (value) => {
-            if (value) {
+            if (value || value === 0) {
```
melloware, line 162: I think you can use ObjectUtils.IsNotEmpty(value)

melloware (changes requested):

components/lib/autocomplete/AutoComplete.js
akshayaqburst: Done, thanks. @melloware

melloware (approved):

PR #5279, opened 2023-11-10
#5278 Added plain prop into classes and updated demos

nitrogenous:
Added plain prop into classes and updated demos #5278


melloware (approved):

melloware:
Thanks for the assist!

nitrogenous:
> Thanks for the assist!

Always!

PR #5319, opened 2023-11-14
Fix #5312: Tooltip left behind on slow machine

### LINKED ISSUE #3965: Tooltip: Stuck after hovering when used in complex page
### Describe the bug

When in a complex page (one that makes the CPU start to struggle a little), tooltips sometimes miss the event that hides them after hovering them, therefore remaining on screen when they should not.

### Reproducer

https://codesandbox.io/s/primereact-test-forked-eijgs6?file=/src/index.js

### PrimeReact version

8.7.3

### React version

18.x

### Language

ALL

### Build / Runtime

Create React App (CRA)

### Browser(s)

_No response_

1. Move the mouse from "one" to "ten", and back to "one" again in a straight line
2. Observer how some tooltips stay visible
3. If it doesn't happen, repeat step 1. It may be tricky to catch this consistently, so, alternatively, one can try moving the mouse slower or faster over the list, following patterns other than a straight line, or cranking up the number of elements in the list below (whose only purpose is to slow things down) to a number challenging for your machine. The goal is to get to a point like this:
image

### Expected behavior

_No response_

### LINKED ISSUE #5312: Tooltip: tooltip does not close when cpu throttled x6
### Describe the bug

when I load cpu with performance insights 6x tooltip does not close.  

### Reproducer

_No response_

### PrimeReact version

9.5.0

### React version

18.x

### Language

TypeScript

### Build / Runtime

Create React App (CRA)

### Browser(s)

google

cpu throttle x6 with google perfomance

### Expected behavior

_No response_

melloware:
Fix #3965
Fix #5312: Tooltip left behind on slow machine

I tested this with my Google Chrome set to 6x slow and it seems to be working

### DIFF
--- a/components/lib/tooltip/Tooltip.js
+++ b/components/lib/tooltip/Tooltip.js
@@ -469,9 +469,7 @@ export const Tooltip = React.memo(
         }, [props.content]);
 
         useUnmountEffect(() => {
-            clearTimeouts();
-            unloadTargetEvents();
-
+            hide();
             ZIndexUtils.clear(elementRef.current);
         });
 


melloware:
I want to do some more testing on this one.

PR #5322, opened 2023-11-15
fix: strength meter popup does not close on escape

### LINKED ISSUE #5321: Password: Strength meter does not close on escape key press
### Describe the bug

Strength meter popup does not close on escape key press

### Reproducer

https://primereact.org/password/#meter

### PrimeReact version

10.0.9

### React version

17.x

### Language

TypeScript

### Build / Runtime

Create React App (CRA)

### Browser(s)

Chrome

1. Go to https://primereact.org/password/#meter
2.  Enter a password.
3. Press escape once strength meter opens.

### Expected behavior

Strength meter popup should close on escape key press.

navedqb:
Fixes #5321

### DIFF
--- a/components/lib/password/Password.js
+++ b/components/lib/password/Password.js
@@ -164,7 +164,7 @@ export const Password = React.memo(
         };
 
         const onKeyup = (e) => {
-            let keyCode = e.keyCode || e.which;
+            const keyCode = e.code;
 
             if (props.feedback) {
                 let value = e.target.value;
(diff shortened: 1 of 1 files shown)


components/lib/password/Password.js
```diff
@@ -201,7 +205,7 @@ export const Password = React.memo(
                 setMeterState(meter);
                 setInfoTextState(label);
 
-                if (!!keyCode && !overlayVisibleState) {
+                if (!!keyCode && keyCode !== 27 && !overlayVisibleState) {
```
melloware, line 208:
```suggestion
                const keyCode = e.code;

                if (!!keyCode && keyCode !== 'Escape' && !overlayVisibleState) {
                    show();
                }
```

melloware: You can delete the line `let keyCode = e.keyCode || e.which;` also

melloware (approved):

PR #5326, opened 2023-11-16
feat:#5327:CascadeSelect:add panelHeaderTemplate for cascadeselect

kl-nevermore:
### Feature Requests
feat:#5327:CascadeSelect:add panelHeaderTemplate for cascadeselect

### DIFF
--- a/components/lib/cascadeselect/CascadeSelect.js
+++ b/components/lib/cascadeselect/CascadeSelect.js
@@ -371,7 +371,24 @@ export const CascadeSelect = React.memo(
             },
             ptm('panel')
         );
+        const createHeader = () => {
+            if (!props.panelHeaderTemplate) return null;
+            const headerProps = mergeProps(
+                {
+                    className: cx('header')
+                },
+                ptm('header')
+            );
+            const content = <div {...headerProps}></div>;
 
+            const defaultOptions = {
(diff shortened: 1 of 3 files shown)


melloware (changes requested):
Not sure I agree with this change. This component doesn't have a header in the other prime libs Vue or NG does it?

What is the purpose of a header?

kl-nevermore:
Custom headers, such as treeselect and dropdown, to accommodate custom search and other functions

melloware:
Yes but those have a built in filter mechanism: https://primereact.org/treeselect/#filter  IT makes no sense for CascadeSelect?  I am just wondering what your use case is?

melloware:
It would also be missing `panelClassName` and `panelFooterTemplate` as well if you are trying to make it like TreeSelect or Dropdown no?

kl-nevermore:
> It would also be missing `panelClassName` and `panelFooterTemplate` as well if you are trying to make it like TreeSelect or Dropdown no?

### You're right I didn't consider other scenarios, this PR is not complete



> Yes but those have a built in filter mechanism: https://primereact.org/treeselect/#filter IT makes no sense for CascadeSelect? I am just wondering what your use case is?

### when I use it normally, I need multiple panels 
image


### when search i can combine itemsTemp to impl my case
image

Issue #5349, opened 2023-11-20
PanelMenu: Menu items with no sub-items toggling border


akshayaqburst:
### Describe the bug

PanelMenu: Menu items with no sub-items toggling border

### PrimeReact version

10.1.0

### React version

17.x

### Language

TypeScript

### Build / Runtime

Create React App (CRA)

### Steps to reproduce the behavior

1. Go to Panel Menu
2. Use a basic Panel Menu in your project
3. Try building a menu item which has no sub items present in it.
4. Click on the newly created menu item which has no sub items.
5. It shows an empty space .

![image](https://github.com/primefaces/primereact/assets/123446355/2fbc529b-7b9f-416c-bd93-c6d0b395c51e)

### Expected behavior

Menu items with no sub-items should not toggle.

After fix - 
![image](https://github.com/primefaces/primereact/assets/123446355/c2f3ef81-4510-4bcb-a3bd-30ae9b44a446)

melloware:
I think it should be handled like this: https://github.com/primefaces/primevue/pull/4844

The toggler should not even be shown:

![image](https://github.com/primefaces/primereact/assets/4399574/2f4d2e6e-9907-4de5-ab41-e9fac8f4d224)

dy-xiaodong2022:
> I think it should be handled like this: [primefaces/primevue#4844](https://github.com/primefaces/primevue/pull/4844)
> 
> The toggler should not even be shown:
> 
> ![image](https://user-images.githubusercontent.com/4399574/284386104-2f4d2e6e-9907-4de5-ab41-e9fac8f4d224.png)

Hi, thanks for your reference but I **don’t** think that remove the empty space or remove the icon is a good idea.
I think it should always show the sub-items menu and the icon if there is a `items` prop given, even it is empty,
Sometimes user may want to show an **empty list**.
If they don’t want the list, they **should not add** the `items` props.
Another way, maybe adding a tip text on the empty place is the best way to solve this problem.

dy-xiaodong2022:
> > I think it should be handled like this: [primefaces/primevue#4844](https://github.com/primefaces/primevue/pull/4844)
> > The toggler should not even be shown:
> > ![image](https://user-images.githubusercontent.com/4399574/284386104-2f4d2e6e-9907-4de5-ab41-e9fac8f4d224.png)
> 
[...]

PR #5351, opened 2023-11-20
fix: #5349, PanelMenu: Menu items with no sub-items toggling border

akshayaqburst:
### Defect Fixes
fix: #5349, PanelMenu: Menu items with no sub-items toggling border

### DIFF
--- a/components/lib/panelmenu/PanelMenu.js
+++ b/components/lib/panelmenu/PanelMenu.js
@@ -209,7 +209,7 @@ export const PanelMenu = React.memo(
 
             const menuContentProps = mergeProps(
                 {
-                    className: cx('menuContent')
+                    className: cx('menuContent', { item })
                 },
                 getPTOptions(item, 'menuContent')
             );
--- a/components/lib/panelmenu/PanelMenuBase.js
+++ b/components/lib/panelmenu/PanelMenuBase.js
@@ -8,7 +8,7 @@ const classes = {
     headerAction: 'p-panelmenu-header-link',
     panel: ({ item }) => classNames('p-panelmenu-panel', item.className),
     header: ({ active, item }) => classNames('p-component p-panelmenu-header', { 'p-highlight': active, 'p-disabled': item.disabled }),
-    menuContent: 'p-panelmenu-content',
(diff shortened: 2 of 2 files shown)


melloware:
I think it should be handled like this: https://github.com/primefaces/primevue/pull/4844

The toggler should not even be shown:

image

melloware (changes requested):
See comments

dy-xiaodong2022:
> I think it should be handled like this: [primefaces/primevue#4844](https://github.com/primefaces/primevue/pull/4844)
> 
> The toggler should not even be shown:
> 
> image

Hi, thanks for your reference but I don’t think that remove the empty space or remove the icon is a good idea.
I think it should always show the sub-items menu and the icon if there is a items prop given, even it is empty,
Sometimes user may want to show an empty list.
If they don’t want the list, they should not add the items props.
Another way, maybe adding a tip text on the empty place is the best way to solve this problem.
So this pr should be **closed**.

dy-xiaodong2022:
> > I think it should be handled like this: [primefaces/primevue#4844](https://github.com/primefaces/primevue/pull/4844)
> > The toggler should not even be shown:
> > image
> 
> Hi, thanks for your reference but I don’t think that remove the empty space or remove the icon is a good idea. I think it should always show the sub-items menu and the icon if there is a items prop given, even it is empty, Sometimes user may want to show an empty list. If they don’t want the list, they should not add the items props. Another way, maybe adding a tip text on the empty place is the best way to solve this problem. So this pr should be **closed**.

That’s mean

With `items` props but empty list:
Should show an empty space or show with some text like “This list is empty.”

With no `items` props but border changed when click:
This should be fixed like my Pull Request in primevue https://github.com/primeface/primevue/pull/4844

dy-xiaodong2022 (commented):
This change shouldn’t be merged or accepted.

akshayaqburst:
@melloware Can I do a fix as per your comment? I see the other person's PR is not merged, so not sure if I should follow his comment. Let me know, thanks.

melloware:
I will fix it.

PR #5397, opened 2023-11-25
Fix #5233: Cell editor ColumnEditorOptions param fix

### LINKED ISSUE #5233: Datatable: Cell editor ColumnEditorOptions param is broken
### Describe the bug

This PR #4745  introduced a bug in the Datatable cell editor options param.
After the first edit of a cell the editingRowData will be set to undefined, so the next time you edit the same cell the ColumnEditorOptions will contain undefined for the rowData.

Simple repro is below, edit a cell once, observe the console log, then edit it a second time.

```
import { Column, ColumnEditorOptions } from 'primereact/column';
import { DataTable } from 'primereact/datatable';
import { InputText } from 'primereact/inputtext';

const items = [
  {
    id: 1,
    name: 'xxx',
    type: 'a',
  },
  {
    id: 2,
    name: 'zzz',
    type: 'b',
  },
];

const TestPage = () => {
  const onCellEditComplete = e => {
    let { rowData, newValue, field, originalEvent: event } = e;
    console.log(`onCellEditComplete ${newValue} ${JSON.stringify(rowData)}`);

    if (newValue.trim().length > 0) rowData[field] = newValue;
    else event.preventDefault();
  };
  const textEditor = (options: ColumnEditorOptions) => {
    console.log(
      `Although value ${options.value} will be ok on second call, rowData ${JSON.stringify(
        options.rowData
      )} will be undefined`
    );
    return (
      <InputText
        value={options.value}
        onChange={ev => {
          if (options.editorCallback) {
            options.editorCallback(ev.target.value);
          }
        }}
      ></InputText>
    );
  };

  return (
    <>
      <DataTable value={items}>
        <Column header="Name" field="name" editor={textEditor} onCellEditComplete={onCellEditComplete}></Column>
        <Column header="Type" field="type" editor={textEditor} onCellEditComplete={onCellEditComplete}></Column>
      </DataTable>
    </>
  );
};
```
In my app I use a custom editor template, where I need to access other fields from the row as well, that feature has broken with the update.
Adding this return here https://github.com/primefaces/primereact/blob/master/components/lib/datatable/BodyCell.js#L190 and setting the state solves my issue, not sure if it has any interference with the original bug.
``` return newRowData;```

### Reproducer

[stackblitz](https://stackblitz.com/edit/react-btmtrb?file=src%2Findex.js,src%2FApp.js)

### PrimeReact version

10.0.7

### React version

18.x

### Language

TypeScript

### Build / Runtime

Next.js

### Browser(s)

_No response_

1. click and edit a cell
2. click elsewhere so the cell switches back to view mode
3. click the cell again
4. observe the console log

### Expected behavior

options.rowData should never be undefined

### PR DESCRIPTION
**Pull request by melloware**
Fix #5233: Cell editor ColumnEditorOptions param fix

### DIFF
--- a/components/lib/datatable/BodyCell.js
+++ b/components/lib/datatable/BodyCell.js
@@ -190,6 +190,8 @@ export const BodyCell = React.memo((props) => {
             } else {
                 event.preventDefault();
             }
+
+            return newRowData;
         });
     };
 

### CODE REVIEW

**Comment by melloware** [comment]
Waiting on response from OP

File: components/lib/datatable/BodyCell.js [hunk 1]
```diff
@@ -189,6 +189,8 @@ export const BodyCell = React.memo((props) => {
                 closeCell(event);
             } else {
                 event.preventDefault();
+
+                return newRowData;
```
vpetya, line 193: @melloware I don't think this is right, but take my comment with a grain of salt, I did not spent too much time with this code. I think you should return newRowData regardless of the validator return true or false, otherwise you set editingRowDataState to undefined in the valid case and nothing will set it to the correct value next time to switch back to editor mode.

melloware: Ok i thought so your comment confused me. I will fix.

PR #5416, opened 2023-11-27
Fix 5415 Action buttons should disabled if any item is not selected

nitrogenous:
#5415

### DIFF
--- a/components/lib/orderlist/OrderListControls.js
+++ b/components/lib/orderlist/OrderListControls.js
@@ -189,7 +189,7 @@ export const OrderListControls = React.memo((props) => {
         ptm('moveBottomButton')
     );
 
-    const isMoveDisabled = !props.selection || !props.selection.length ? true : false;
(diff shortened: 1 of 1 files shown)


melloware (approved):

components/lib/orderlist/OrderListControls.js
```diff
@@ -189,12 +189,14 @@ export const OrderListControls = React.memo((props) => {
         ptm('moveBottomButton')
     );
 
+    const isMoveDisabled = !props.selection || !props.selection.length ? true : false;
```
melloware, line 192:
```suggestion
    const isMoveDisabled = ObjectUtils.isEmpty(props.selection);
```

melloware (changes requested):

melloware (approved):

components/lib/orderlist/OrderListControls.js
nitrogenous: I missed that thank you for fixing it

PR #5456, opened 2023-11-30
Fix 5455 updating demo environments to vite

### LINKED ISSUE #5455: Updating demo environments to Vite

nitrogenous:
Fix #5455 updating demo environments to vite

### DIFF
--- a/components/doc/common/codeeditor/index.js
+++ b/components/doc/common/codeeditor/index.js
@@ -1,13 +1,20 @@
 import sdk from '@stackblitz/sdk';
-import { getCRA, getNextJS } from './templates';
+import { getVite } from './templates';
 
 const useCodeSandbox = (props) => {
     const getSandboxParameters = (sourceType) => {
-        const { files, dependencies, sourceFileName } = props.template === 'cra' ? getCRA(props, sourceType) : getNextJS(props, sourceType);
+        const sandboxDependencies = {
+            vite: '^4.0.0',
+            '@vitejs/plugin-react': '^4.0.0'
+        };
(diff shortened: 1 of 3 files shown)


components/doc/common/codeeditor/index.js
```diff
@@ -1,13 +1,20 @@
...
+            '@vitejs/plugin-react': '^4.0.0'
+        };
+        const { files, dependencies, sourceFileName } = getVite({ dependencies: sandboxDependencies, ...props }, sourceType);
 
         files['sandbox.config.json'] = {
             content: {
-                infiniteLoopProtection: false
+                infiniteLoopProtection: false,
+                container: {
+                    node: '16'
```
melloware, line 16: Should we use Node 18 or 20 since those are LTS now?

melloware (changes requested):

components/doc/common/codeeditor/index.js
nitrogenous: Codesandbox not supporting 18 or 20. Unfortunately this is the lates version that we can use. This is why we are also using the vite 4.0.0 for codesandbox

nitrogenous: When i set the node version higher than 16 it sets the node version to the default value which is 10


image

nitrogenous: Codesandbox have this buckets called DevBox and it supports Node 18 or 20 we can basically use whatever we want but I couldn't find any examples about creating DevBox buckets via API

melloware: OK no problem thanks for checking!

melloware (approved):

gucal (approved):

PR #5502, opened 2023-12-02
fix:CascadeSelect missing some values in pt

### LINKED ISSUE #5479: CascadeSelect: missing some values in pt
I can get more states in variable  classes of  CascadeSelect(otherComp)Base.js,but only  get metadata in pt
Should there be consistency here?

### Example
i can get ` option, isOptionGroup, activeOptionState` in https://github.com/primefaces/primereact/blob/7f0c2c024cb00d46de9963767ff831cee436f5be/components/lib/cascadeselect/CascadeSelectBase.js#L28

only get `state` and `props` in pt props 
https://stackblitz.com/edit/react-nz1uos?file=src%2FApp.js,src%2Findex.js

kl-nevermore:
Fix #5479

### DIFF
--- a/components/lib/cascadeselect/CascadeSelect.js
+++ b/components/lib/cascadeselect/CascadeSelect.js
@@ -338,7 +338,7 @@ export const CascadeSelect = React.memo(
                     ref: labelRef,
                     className: cx('label', { label })
                 },
-                ptm('label')
+                ptm('label', { label })
             );
 
             return <span {...labelProps}>{label}</span>;
@@ -443,7 +443,7 @@ export const CascadeSelect = React.memo(
                     onClick: (e) => onClick(e)
                 },
                 otherProps,
-                ptm('root')
+                ptm('root', { focusedState, overlayVisibleState })
             );
 
             return (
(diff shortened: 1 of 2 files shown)


kl-nevermore:
I can't impl some styles in unstyled using pt if I can't get consistent values

### example
```
// in classes can add style for active item,but pt can not 
  'p-cascadeselect-item-active p-highlight': activeOptionState === option
```
I referred to some comp like dropdown use ptm pass value to pt
Other components also have this problem
It should be necessary for ptm and cx to be consistent?
If this is correct, the types file should also be updated

melloware:
That is a good question. It seems CX and PTM are definitely treated differently I am not quite sure why. And yes the Types would have to be updated that is what @Dalorzo has been helping with.

melloware:
@kl-nevermore @Dalorzo see my PR fixing these issues:

https://github.com/primefaces/primereact/pull/5507

So for PTM you always have 3 objects `context, state, props`

```ts
export interface CascadeSelectPassThroughMethodOptions {
    props: CascadeSelectProps;
    state: CascadeSelectState;
    context: CascadeSelectContext;
}
```

So for your examples above in passthrough you will use.

`label` - use `context.label` to get the labels

`item` - use `state.grouped` and `state.active` to determine if the item is active or in a group.

'list and sublist` - now have the proper `context` value which includes all API options like ripple and `inputfilled` etc.

PR #5552, opened 2023-12-08
fix:InputNumber: Text selection not working on input focus when using a s…

### LINKED ISSUE #5537: InputNumber: Text selection not working on input focus when using a suffix or prefix
### Describe the bug

When rendering `<InputNumber>` with a prefix or suffix, tabbing to that input will not cause the text to be selected

### Reproducer

https://codesandbox.io/p/sandbox/primereact-test-forked-662yfm

### PrimeReact version

10.2.1

### React version

18.x

### Language

TypeScript

### Build / Runtime

Create React App (CRA)

### Browser(s)

Chrome 119

1. Open the forked codesandbox above
2. You'll see 3 inputs (the first without a prefix/suffix, and the other two with one)
3. Click on first input
4. Use tab / shift+tab to focus on the various inputs

### Expected behavior

Expected: When focused, the inputs have their text selected
Actual: Only the first input behaves this way. If there is a prefix or suffix, the text does not get selected

kl-nevermore:
Fix #5537

### DIFF
--- a/components/lib/inputnumber/InputNumber.js
+++ b/components/lib/inputnumber/InputNumber.js
@@ -963,6 +963,7 @@ export const InputNumber = React.memo(
             props.onFocus && props.onFocus(event);
 
             if ((props.suffix || props.currency || props.prefix) && inputRef.current && !isFocusedByClick.current) {
+                // GitHub #1866,#5537
                 let inputValue = inputRef.current.value;
                 let prefixLength = (prefixChar.current || '').length;
                 let suffixLength = (suffixChar.current || '').length;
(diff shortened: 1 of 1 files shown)


components/lib/inputnumber/InputNumber.js
```diff
@@ -1067,6 +1078,7 @@ export const InputNumber = React.memo(
                     onKeyPress={onInputKeyUp}
                     onInput={onInput}
                     onClick={onInputClick}
+                    onMouseDown={onInputMouseDown}
```
melloware, line 1081: Should this be `onPointerDown` instead so it works for touch devices as well?

melloware (changes requested):

components/lib/inputnumber/InputNumber.js
kl-nevermore: Thanks! I learned

PR #5602, opened 2023-12-14
feat:Calendar:add disabledDate

### LINKED ISSUE #5581: Calendar:about disabled logic
### monthPickerValues  `data-p-disabled` always true
https://github.com/primefaces/primereact/blob/master/components/lib/calendar/Calendar.js#L3654
### yearPickerValues `data-p-disabled` always false
https://github.com/primefaces/primereact/blob/master/components/lib/calendar/Calendar.js#L3697

###  is this an issue or a feature? if it will be fixed, Is it a good choice to accept a new prop like example
```tsx
disabledDate?: (year: number, month: number, day: number) => boolean;

<Calendar
    value={date}
    onChange={(e) => setDate(e.value)}
    disabledDate={(y, m, d) => {
        return y > 1999 && m > 1;
    }}
/>
```

### Reproducer

https://codesandbox.io/p/sandbox/primereact-test-forked-6y4y7n?file=%2Fsrc%2Findex.js

### PrimeReact version

latest

### React version

18.x

### Language

TypeScript

### Build / Runtime

Create React App (CRA)

### Browser(s)

_No response_

_No response_

### Expected behavior

_No response_

kl-nevermore:
Fix #5581 

- If any date can be selected in currentYear or currentMonth, then the year and month should also be selectable
- If any month can be selected, then the year should also be selectable
- if all date disabled ,then the year and month should also be disabled
- if all month disabled ,then the year and date should also be disabled
- if all year disabled ,then the mouth and date should also be disabled

I will  add test cases recently.

`checkDateIsDisabled` cost time (i7-12700 windows10 chrome)

<table>
  <tr>
    <td>view</td>
    <td>1x</td>
    <td>4x</td>
     <td>6x</td>
  </tr>
  <tr>
    <td>year</td>
    <td>2.3ms</td>
    <td>11.4ms</td>
<td>18.5ms</td>
  </tr>
  <tr>
    <td>month</td>
    <td>1.1ms</td>
    <td>3.1ms</td>
<td>4.6ms</td>
  </tr>
</table>


components/lib/calendar/Calendar.js
```diff
@@ -3685,6 +3689,28 @@ export const Calendar = React.memo(
             return null;
         };
 
+        const checkDateIsDisabled = (month, year) => {
+            const daysCountInAllMonth = month === -1 ? new Array(12).fill(0).map((_, i) => getDaysCountInMonth(i, year)) : [getDaysCountInMonth(month, year)];
```
kl-nevermore, line 3693: should use `monthPickerValues()`  replaced

components/lib/calendar/calendar.d.ts
```diff
@@ -566,6 +566,10 @@ interface CalendarBaseProps {
      * @defaultValue false
      */
     disabled?: boolean | undefined;
+    /**
+     * Specify the date that cannot be selected.
+     */
+    disabledDate?: ((day: number, month: number, year: number) => boolean) | undefined;
```
melloware, line 572: Why not just use a `disabledDates` array with a single valyue?

components/lib/calendar/Calendar.js
melloware: Not sure what this means?

kl-nevermore: `new Array(12).fill(0)` ->`monthPickerValues()`

components/lib/calendar/calendar.d.ts
kl-nevermore: Sorry I don't understand Can you give me an example? thanks!

melloware: Well it looks like you are trying to disable a "single" day right?

Why not just use `disabledDates?: Date[] | undefined;`

So you can just do...

```js
const disabledDay = new Date(2023, 10, 15);
const disabledDates = [disabledDay]; // array of 1 day

<Calendar disabledDates={disabledDates} />
```

melloware: Yes I was just trying to prevent adding new props.

components/lib/calendar/Calendar.js
```diff
@@ -3685,6 +3685,25 @@ export const Calendar = React.memo(
             return null;
         };
 
+        const checkDateIsDisabled = (month, year) => {
```
melloware, line 3688: This looks great I think for consistency lets just rename `checkDateIsDisabled` to `isDateDisabled` since that seems to be a pattern used through this codebase.

melloware (changes requested):

components/lib/calendar/Calendar.js
kl-nevermore: I wanted to do it before, but it was used

melloware: Oh let me look.

kl-nevermore: Maybe it can be called isYearDisabled and isMonthDisabled like ng

melloware: OK I just refactored `isDayDisabled` on your branch to be one single method.  I found it confusing that it was two differnet methods for isDateDisabled and isDayDisabled.  but i think the good name for it should be `isMonthYearDisabled`

kl-nevermore: Okay,I'll go home later and update the code

PR #5608, opened 2023-12-18
fix:InputNumber's issue when Locale set de-DE

### LINKED ISSUE #5548: InputNumber: Cursor stuck when using minFractionDigits and comma as decimal separator
### Describe the bug

When using the `InputNumber` component with `minFractionDigits` set to 2 or more, and a locale that uses a comma as a decimal separator, the cursor gets stuck on the first character when setting the decimal value.

### Reproducer

https://codesandbox.io/p/sandbox/primereact-demo-forked-5cs9lm

### PrimeReact version

10.2.1

### React version

18.x

### Language

TypeScript

### Build / Runtime

Next.js

### Browser(s)

_No response_

1. Set `minFractionDigits` to 2 or more in `InputNumber` component
2. Set locale to one that uses comma as decimal separator. Eg: de-DE, pt-BR
3. Try to set a decimal value

### Expected behavior

The cursor should move freely, allowing the user to input decimal values without any hindrance.

### LINKED ISSUE #5589: InputNumber: (Germany) second step of currency can't changed
First, delete all the input, type some numbers, and then type a comma. The second step of the currency will not change anymore.

### LINKED ISSUE #5607: InputNumber: Incorrect duplication of fractional part of number
### Describe the bug

related https://github.com/primefaces/primevue/issues/4392

the value is also wrong after defocusing
20231218171800_rec_

### Reproducer

https://codesandbox.io/p/sandbox/primereact-demo-forked-jxfcvr?file=%2Fsrc%2FApp.js%3A45%2C32

### PrimeReact version

latest

### React version

18.x

### Language

TypeScript

### Build / Runtime

Create React App (CRA)

### Browser(s)

_No response_

_No response_

### Expected behavior

_No response_

kl-nevermore:
Fix #5548
Fix #5589 
Fix #5607

### DIFF
--- a/components/lib/inputnumber/InputNumber.js
+++ b/components/lib/inputnumber/InputNumber.js
@@ -75,13 +75,16 @@ export const InputNumber = React.memo(
             _group.current = getGroupingExpression();
             _minusSign.current = getMinusSignExpression();
             _currency.current = getCurrencyExpression();
-            _decimal.current = getDecimalExpression();
-            _decimalSeparator.current = getDecimalSeparator();
+            _decimal.current = getDecimalExpression(); // regular expression /[,]/g, /[.]/g, /[]/g
+            _decimalSeparator.current = getDecimalSeparator(); // current decimal separator  '.', ','
             _suffix.current = getSuffixExpression();
             _prefix.current = getPrefixExpression();
             _index.current = (d) => index.get(d);
         };
 
+        /**
+         * get decimal separator in current locale
+         */
         const getDecimalSeparator = () => {
             return new Intl.NumberFormat(_locale, { useGrouping: false }).format(1.1).trim().replace(_numeral.current, '');
         };
--- a/components/lib/inputnumber/InputNumber.js
+++ b/components/lib/inputnumber/InputNumber.js
@@ -40,6 +40,7 @@ export const InputNumber = React.memo(
         const _minusSign = React.useRef(null);
         const _currency = React.useRef(null);
         const _decimal = React.useRef(null);
+        const _decimalSeparator = React.useRef(null);
         const _suffix = React.useRef(null);
         const _prefix = React.useRef(null);
         const _index = React.useRef(null);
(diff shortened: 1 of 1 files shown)


components/lib/inputnumber/InputNumber.js
```diff
@@ -75,11 +76,16 @@ export const InputNumber = React.memo(
             _minusSign.current = getMinusSignExpression();
             _currency.current = getCurrencyExpression();
             _decimal.current = getDecimalExpression();
+            _decimalSeparator.current = getDecimalSeparator();
```
melloware, line 79: I don't understand the difference between `decimalSeparater.current` and `decimal.current` I think `decimal.current` was supposed to be the current decimal place regardless of whether its a `.` or a `,` ?

kl-nevermore: `decimal.current` is a regular
 `/[,]/g`, `/[.]/g`, `/[]/g`
```
const getDecimalExpression = () => {
            const formatter = new Intl.NumberFormat(_locale, { ...getOptions(), useGrouping: false });

            return new RegExp(`[${formatter.format(1.1).replace(_currency.current, '').trim().replace(_numeral.current, '')}]`, 'g');
        };
```

components/lib/inputnumber/InputNumber.js
```diff
@@ -75,11 +76,16 @@ export const InputNumber = React.memo(
...
             _suffix.current = getSuffixExpression();
             _prefix.current = getPrefixExpression();
             _index.current = (d) => index.get(d);
         };
 
+        const getDecimalSeparator = () => {
```
melloware, line 85: OK then can you add a comment in this method exactly what it is doing with examples?

components/lib/inputnumber/InputNumber.js
```diff
@@ -40,6 +40,7 @@ export const InputNumber = React.memo(
         const _minusSign = React.useRef(null);
         const _currency = React.useRef(null);
         const _decimal = React.useRef(null);
+        const _decimalSeparator = React.useRef(null);
```
melloware, line 43: i think we should add a comment next to each field about what they hold like...

```js
const _decimal = React.useRef(null); // regular expression /[,]/g, /[.]/g, /[]/g
const _decimalSeparator = React.useRef(null); // ?????

**Review by melloware — CHANGES_REQUESTED** [review]

File: components/lib/inputnumber/InputNumber.js (hunk 2 above)
Comment (kl-nevermore) [reply components/lib/inputnumber/InputNumber.js:85]: Sure,this function is get decimal separator in current locale(`','`,`'.'`)

The reason for these problems is when setting a comma-separated decimal like Germany,the value ​​are like `1231.1` divided by `.`, but the `_decimal .current` is `/[,]/g` 
So what I want is to store a current decimal separator, if the value is a decimal, replace the last `.` with `_decimalSeparator.current`

File: components/lib/inputnumber/InputNumber.js (hunk 3 above)
Comment (kl-nevermore) [reply components/lib/inputnumber/InputNumber.js:43]: Do these need to be commented together?
```js
const _minusSign = React.useRef(null);
const _currency = React.useRef(null);
const _decimal = React.useRef(null);
```

melloware: That would be great.

melloware (approved):

Issue #5613, opened 2023-12-18
TreeSelect: TreeSelect component is not supporting tooltips and is an issue in multiple select mode.


akshayaqburst:
### Describe the bug

TreeSelect: TreeSelect component is not supporting tooltips and is an issue in multiple select mode.
![image](https://github.com/primefaces/primereact/assets/123446355/2b1e6de8-1385-4847-8be4-743d4cdb80fc)

### PrimeReact version

10.2.1

### React version

17.x

### Language

TypeScript

### Build / Runtime

Create React App (CRA)

### Steps to reproduce the behavior

1. Go to https://primereact.org/treeselect/#multiple.
2. Select more than 3 items. 
3. Some items are getting cuttoff. 
4. Try to pass in a tooltip prop

Tooltip prop is not supported.

### Expected behavior

Tooltip prop should be supported in TreeSelect.

PR #5615, opened 2023-12-18
fix: #5613, TreeSelect: TreeSelect component is not supporting tooltips and is an issue in multiple select mode.

akshayaqburst:
### Defect Fixes
fix: #5613, TreeSelect: TreeSelect component is not supporting tooltips and is an issue in multiple select mode.


Malihe752:
Hi
I have received the latest version of PrimeReact (9.6.4) But I dont have these changes in my code. How can I use tooltip in my treeSelect? my treeSelect doesnt have any tooltip prop. I appreciate any help.

melloware:
@Malihe752 if you look at the Ticket it is marked "10.3.0" so it was fixed in 10.3.0 or higher which is why its not in your 9.6.4 version.

PR #5532, opened 2023-12-06
DataTable:converted to data- lookups instead of className lookups

kl-nevermore:
Fix #5524 
Fix #5489

PR #5626, opened 2023-12-21
feat:CascadeSelect support scroll in container

### LINKED ISSUE #2229: CascadeSelect: dropdown container creates scrolls in page instead of list container
**I'm submitting a ...**
```
[ x] bug report
[ ] feature request
```

https://codesandbox.io/s/exciting-villani-vqex4?file=/src/demo/CascadeSelectDemo.js:6294-6507

**Current behavior**
The issue with cascade select happens when the list is too long for the page to display in available space below the component, so it creates scrolls to the whole page instead of creating scroll just for the list container of the dropdown. When the list have many elements, and those elements have multiple levels.

image

**Expected behavior**
I found a custom fix to limit height of the cascade select container to prevent cascade to create scrolls in page and limits the max-height to the space in screen available below the component, the right CascadeSelect uses this strategy, but still, only the second container for the child items are limited and using the scroll as intended, the first level of items is limited to the space but the items are going below the limit without a scroll-y appearing.
We probably need a portal to fix this problem to use only available space in screen to display the dropdown preventing to create scrolls in the whole page, and just create scrolls in the items containers of the list in each level of children items.

image

1. Place a CascadeSelec in the page (preferably vertically centered or the the bottom of the screen).
2. Add multiple "options" and each option with children, add enough of them so the dropdown list container goes below the screen space available.
3. You can see scroll bars for page are created, instead of adding scroll bars to the container of the list for each level in the DropDown.

Browser Chrome v92.0.4515.107, OS Windows 10 20h1
"react": "^16.10.2",
"primeflex": "^2.0.0",
"primeicons": "^4.1.0",
"primereact": "^6.4.1",

* **React version:**
16.10.2

* **PrimeReact version:**
6.4.1

* **Browser:**
- Chrome v92.0.4515.107
- Firefox v90.0.2

 
* **Language:**
Javascript

kl-nevermore:
Fix #2229 
- [x] code
- [x] test
- [x] pt & tw

## pc
20231221155417_rec_

### before
20231221160735_rec_

### after
20231221151204_rec_

### DIFF
--- a/components/lib/cascadeselect/CascadeSelect.js
+++ b/components/lib/cascadeselect/CascadeSelect.js
@@ -235,28 +235,28 @@ export const CascadeSelect = React.memo(
                 const selector = `${attributeSelectorState}_panel`;
                 const innerHTML = `
 @media screen and (max-width: ${props.breakpoint}) {
-    .p-cascadeselect-panel[${selector}] .p-cascadeselect-items-wrapper > ul {
+    [data-pc-section="panel"][${selector}] [data-pc-section="wrapper"] > ul {
         max-height: ${props.scrollHeight};
         overflow: ${props.scrollHeight ? 'auto' : ''};
     }
 
-    .p-cascadeselect-panel[${selector}] .p-cascadeselect-sublist-wrapper {
+    [data-pc-section="panel"][${selector}] [data-pc-section="sublistwrapper"] {
         position:relative;
         left:0;
     }
 
-    .p-cascadeselect-panel[${selector}] .p-cascadeselect-sublist {
+    [data-pc-section="panel"][${selector}] [data-pc-section="sublist"] {
         overflow: hidden !important;
     }
 
-    .p-cascadeselect-panel[${selector}] .p-cascadeselect-item-active  .p-cascadeselect-sublist {
+    [data-pc-section="panel"][${selector}] [data-pc-section="item"][data-p-active="true"] > [data-pc-section="sublistwrapper"] > [data-pc-section="sublist"] {
         left: 0;
         box-shadow: none;
         border-radius: 0;
         padding: 0 0 0 calc(var(--inline-spacing) * 2); /* @todo */
     }
 
-    .p-cascadeselect-panel[${selector}] .p-cascadeselect-group-icon:before {
+    [data-pc-section="panel"][${selector}] [data-pc-section="optiongroupicon"]:before {
         content: "\\e930";
     }
 }
--- a/components/lib/cascadeselect/CascadeSelect.js
+++ b/components/lib/cascadeselect/CascadeSelect.js
@@ -235,28 +235,28 @@ export const CascadeSelect = React.memo(
                 const selector = `${attributeSelectorState}_panel`;
                 const innerHTML = `
 @media screen and (max-width: ${props.breakpoint}) {
-    [data-pc-section="panel"][${selector}] [data-pc-section="wrapper"] > ul {
+    .p-cascadeselect-panel[${selector}] .p-cascadeselect-items-wrapper > ul {
         max-height: ${props.scrollHeight};
         overflow: ${props.scrollHeight ? 'auto' : ''};
     }
 
-    [data-pc-section="panel"][${selector}] [data-pc-section="sublistwrapper"] {
+    .p-cascadeselect-panel[${selector}] .p-cascadeselect-sublist-wrapper {
         position:relative;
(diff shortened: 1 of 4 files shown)


components/lib/cascadeselect/CascadeSelectSub.js
```diff
@@ -18,14 +18,14 @@ export const CascadeSelectSub = React.memo((props) => {
...
-        const parentItem = elementRef.current.parentElement;
+        const parentItem = elementRef.current.parentElement.parentElement;
         const containerOffset = DomHandler.getOffset(parentItem);
         const viewport = DomHandler.getViewport();
         const sublistWidth = elementRef.current.offsetParent ? elementRef.current.offsetWidth : DomHandler.getHiddenElementOuterWidth(element);
         const itemOuterWidth = DomHandler.getOuterWidth(parentItem.children[0]);
 
         if (parseInt(containerOffset.left, 10) + itemOuterWidth + sublistWidth > viewport.width - DomHandler.calculateScrollbarWidth()) {
-            elementRef.current.style.left = '-100%';
+            elementRef.current.parentElement.style.left = -100 * props.level + '%';
```
kl-nevermore, line 28: Not sure this is a good solution, I feel it would be better to use breakpoints

components/lib/cascadeselect/CascadeSelect.js
```diff
@@ -235,28 +235,28 @@ export const CascadeSelect = React.memo(
                 const selector = `${attributeSelectorState}_panel`;
                 const innerHTML = `
 @media screen and (max-width: ${props.breakpoint}) {
-    .p-cascadeselect-panel[${selector}] .p-cascadeselect-items-wrapper > ul {
```
melloware, line 238: Uusually these `p-` classes are left like this so they can be overridden by the PrimeReact SASS Themes: https://github.com/primefaces/primereact-sass-theme/blob/main/theme-base/components/input/_cascadeselect.scss

kl-nevermore: Do you think it can be modified like this?
```js
isUnstyled() ? [data-pc-section="panel"][${selector}] : .p-cascadeselect-panel[${selector}]
```

kl-nevermore: If the breakpoints need to use the className,
I should revert className in  #5586 and #5532

kl-nevermore: @melloware  
I have reverted
Do you think unstyled mode should be implemented breakpoints by the user?


> I'm not sure if in unstyled mode, Is this logic should be implemented by the user?
> 
> Do you think it should be modified like this?
> 
> ```tsx
> isUnstyled()  ? [data-pc-section="panel"][${selector}] : .p-cascadeselect-panel[${selector}]
> ```

melloware: No in the Dialog fix its OK because its creating an inline width style.

components/lib/cascadeselect/CascadeSelectBase.js
```diff
@@ -75,11 +76,6 @@ const styles = `
         min-width: 100%;
     }
     
-    .p-cascadeselect-panel {
-        position: absolute;
-        top: 0;
-        left: 0;
-    }
```
melloware, line 82: These changes probably mean we need SASS changes for the PrimeTEk themes: https://github.com/primefaces/primereact-sass-theme

melloware (changes requested):

components/lib/cascadeselect/CascadeSelectBase.js
kl-nevermore: I checked [_cascadeselect.scss](https://github.com/primefaces/primereact-sass-theme/blob/main/theme-base/components/input/_cascadeselect.scss) and  [themes](https://github.com/primefaces/primereact-sass-theme/tree/main/themes), there are no position related css

components/lib/cascadeselect/CascadeSelect.js
kl-nevermore: Thank you for your review!
I will submit a PR to revert the breakpoints of Datatable

melloware (approved):

melloware:
Assigned to PrimeTek so they are aware and can decide to make this change in PrimeVue and NG.

kl-nevermore:
My fork project has some problems. I will resubmit recently.
@melloware

PR #5647, opened 2023-12-26
Accept array as PT value

kyjus25:
### Defect Fixes
PrimeReact required the use of an internal `classNames` utility for array class joining. Hoist this out so it matches the Vue implementation. Remains backwards compatible.

https://github.com/primefaces/primereact/blob/master/components/lib/passthrough/tailwind/index.js#L613
https://github.com/primefaces/primevue/blob/master/components/lib/passthrough/tailwind/index.js#L741

### Feature Requests

### DIFF
--- a/components/lib/componentbase/ComponentBase.js
+++ b/components/lib/componentbase/ComponentBase.js
@@ -519,13 +519,10 @@ export const ComponentBase = {
 
             const getPTClassValue = (...args) => {
                 const value = getOptionValue(...args);
-                if (Array.isArray(value)) return { className: classNames(...value) }
+                if (Array.isArray(value)) return { className: classNames(...value) };
                 if (ObjectUtils.isString(value)) return { className: value };
-                if (value?.hasOwnProperty('className') && ObjectUtils.isString(value.className)) {
-                    return { className: classNames(value.className)}
-                }
                 if (value?.hasOwnProperty('className') && Array.isArray(value.className)) {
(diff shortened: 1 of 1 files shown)


SoyDiego (approved):

components/lib/componentbase/ComponentBase.js
```diff
@@ -519,8 +519,12 @@ export const ComponentBase = {
...
             const getPTClassValue = (...args) => {
                 const value = getOptionValue(...args);
-
-                return ObjectUtils.isString(value) ? { className: value } : value;
+                if (Array.isArray(value)) return { className: classNames(...value) };
+                if (ObjectUtils.isString(value)) return { className: value };
+                if (value?.hasOwnProperty('className') && Array.isArray(value.className)) {
+                    return { className: classNames(...value.className) };
+                }
+                return value;
```
melloware, line 527: I thought you mentioned PrimeVue already has this but that is not what I am seeing?

See:  https://github.com/primefaces/primevue/blob/c1a78ef9222d0e7d3f353e13dcab43611c2ea884/components/lib/basecomponent/BaseComponent.vue#L129-L132

melloware (changes requested):

melloware:
i prefer it match the Vue implementation I did not realize Vue supported that out of the box so this is a good change if React does not support it natively to keep them in sync.

PR #5674, opened 2024-01-02
fix:Ripple and Tabview bugs in Next.js App Router

### LINKED ISSUE #5127: PrimeReactProvider: Error next14 app router
### Describe the bug

Captura de pantalla 2023-10-20 a la(s) 11 19 30
Captura de pantalla 2023-10-20 a la(s) 11 21 21
Captura de pantalla 2023-10-20 a la(s) 11 20 09
Captura de pantalla 2023-10-20 a la(s) 11 20 39

### Reproducer

_No response_

### PrimeReact version

10.0.3

### React version

18.x

### Language

TypeScript

### Build / Runtime

Next.js

### Browser(s)

chrome 118

_No response_

### Expected behavior

_No response_

### LINKED ISSUE #5395: TabView: Doesn't render in NextJS 14
i'm using nextjs 14.0.3, prime react 10.1.1, i have follow walkthrough on your website but it seem not working for me, when i use tabview, tabpanel it return for me a blank div, what i have wrong, plz help me with this 

image

image

kl-nevermore:
Because I use Next14 in my project, So I fixed the forked project
Should we wait for PrimeTek to fix it or use a temporary solution first?

Fix #5395 

Fix #5127  refer to https://github.com/vercel/next.js/discussions/17443#discussioncomment-637879

### DIFF
--- a/components/lib/ripple/Ripple.js
+++ b/components/lib/ripple/Ripple.js
@@ -78,20 +78,18 @@ export const Ripple = React.memo(
             getTarget: () => targetRef.current
         }));
 
-        useUpdateEffect(() => {
-            if (isRippleActive) {
-                // SSR next.js > 13
-                setMounted(true);
-            }
-        }, [isRippleActive]);
+        useMountEffect(() => {
+            // SSR next.js > 13
+            setMounted(true);
+        });
 
         useUpdateEffect(() => {
-            if (isRippleActive && isMounted && inkRef.current) {
+            if (isMounted && inkRef.current) {
                 targetRef.current = getTarget();
                 setDimensions();
                 bindEvents();
             }
-        }, [isRippleActive, isMounted]);
+        }, [isMounted]);
 
         useUpdateEffect(() => {
             if (inkRef.current && !targetRef.current) {
--- a/components/lib/ripple/Ripple.js
+++ b/components/lib/ripple/Ripple.js
@@ -79,7 +79,7 @@ export const Ripple = React.memo(
         }));
 
         useMountEffect(() => {
-            // SSR next.js > 13
+            // for App Router in Next.js ^14
             setMounted(true);
         });
 
--- a/components/lib/ripple/Ripple.js
+++ b/components/lib/ripple/Ripple.js
@@ -6,6 +6,7 @@ import { RippleBase } from './RippleBase';
 
 export const Ripple = React.memo(
     React.forwardRef((inProps, ref) => {
+        const [isMounted, setMounted] = React.useState(false);
         const inkRef = React.useRef(null);
         const targetRef = React.useRef(null);
         const context = React.useContext(PrimeReactContext);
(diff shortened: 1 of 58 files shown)


components/lib/utils/ObjectUtils.js
```diff
@@ -222,7 +222,13 @@ export default class ObjectUtils {
     static isValidChild(child, type, validTypes) {
         /* eslint-disable */
         if (child) {
-            const childType = this.getComponentProp(child, '__TYPE') || (child.type ? child.type.displayName : undefined);
+            let childType = this.getComponentProp(child, '__TYPE') || (child.type ? child.type.displayName : undefined);
+
+            // SSR next.js >= 13 (app router)
+            if (!childType && child?.type?._payload?.value) {
```
kl-nevermore, line 228: It might be better to refactor the component

melloware: I like this fix for now its the least intrusive

components/lib/ripple/Ripple.js
```diff
@@ -77,13 +78,20 @@ export const Ripple = React.memo(
             getTarget: () => targetRef.current
         }));
 
-        useMountEffect(() => {
-            if (inkRef.current) {
+        useUpdateEffect(() => {
```
melloware, line 81: Should this just be `React.useEffect` and not update effect so it happens no matter what?  Right now its only happening on Update AFTER the component is mounted does this work in React.Strict and non-Strict mode?

melloware (changes requested):

components/lib/ripple/Ripple.js
kl-nevermore: I used to think that if context.ripple changes, Ripple will re-render 
But I just tested it and it won’t render [demo](https://codesandbox.io/p/sandbox/primereact-demo-forked-66cj8z?file=%2Fsrc%2Findex.js%3A14%2C17)
If this is the design, I think it can simplify like
```tsx
useEffect(()=>{
   setMounted(true);
},[])
 useUpdateEffect(() => {
            if ( isMounted && inkRef.current) {
                targetRef.current = getTarget();
                setDimensions();
                bindEvents();
            }
      
 }, [ isMounted]);
```

kl-nevermore: I haven't tested the production environment yet next12 13 14 and  vite

melloware: OK even simpler.

```js
useMountEffect(()=>{
   setMounted(true);
})
```

as that is exactly what useMountEffect does.

kl-nevermore:
If use Next.js 13.x, we need to 
```tsx
import { PrimereactProvider } from 'primereact/api/api.esm'
import { Button } from 'primereact/button/button.esm'
...
```
Looks like a Next.js issue
I tested strictMode and non-strictMode in Next.js 14.x 12.x and vite project

kl-nevermore:
I just test it working on 13.5.7-canary.37, But a new problem is coming

components/lib/utils/ObjectUtils.js
```diff
@@ -222,7 +222,13 @@ export default class ObjectUtils {
...
+                childType = child.type._payload.value.find((v) => v === type);
```
kl-nevermore, line 229: There not compatible if you enable the experimental appDir in ~13.2

melloware (approved):

PR #5693, opened 2024-01-05
Fix #5692: Content is optional

### LINKED ISSUE #5692: TypeScript: 10.3.0 Components complaining `content` is required
Typscript is now complaining the `content` is now required when it should not be in 10.3.0

image

melloware:
Fix #5692: Content is optional

This broke my apps upgrading to 10.3.0


nitrogenous (approved):
CR OK

PR #5732, opened 2024-01-10
Fix #4089: Orderlist require dataKey

### LINKED ISSUE #4089: OrderList: Behavior when using spread syntax in onChange
### Describe the bug

Hello.
When updating state using spread syntax with the value of [event.value] passed when the onChange event of OrderList is fired, the following bug occurs.
- Focus is lost when selecting.
- When the down arrow is pressed continuously, "undefined" is stored in the array and becomes an invalid value.

### Reproducer

https://codesandbox.io/s/primereact-demo-forked-i55nkl?file=/src/demo.js

### PrimeReact version

8.3.0

### React version

17.x

### Language

TypeScript

### Build / Runtime

Next.js

### Browser(s)

_No response_

1. select one of the products shown.
2. press the down arrow once (the focus is removed at this point)
3. Press the down arrow again (the value of the item at the top is undefined)

image

### Expected behavior

Ideally, no focus is lost and no undefined is stored in the array.

melloware:
Fix #4089: Orderlist require dataKey

cc @kl-nevermore

### DIFF
--- a/components/doc/orderlist/basicdoc.js
+++ b/components/doc/orderlist/basicdoc.js
@@ -29,7 +29,7 @@ export function BasicDoc(props) {
 
     const code = {
         basic: `
-<OrderList value={products} onChange={(e) => setProducts(e.value)} itemTemplate={itemTemplate} header="Products"></OrderList>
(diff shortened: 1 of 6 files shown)


kl-nevermore:
Maybe it can be released in the next major version?
I guess `PickList`(some Components with sorting) may also have this problem
I'll check it later

kl-nevermore:
PR #5737

Issue #4994, opened 2023-10-01
DataTable: Filter Menu Icon disappeared after sorting


marinashe:
### Describe the bug

If a column has the property `showFilterMenu={false}` (in my example, it's the first column, "Name"), after I sort another column (for example, the second column, "Country"), the Filter Menu Icon disappears for a third column (which is the third column, "Company").

https://github.com/primefaces/primereact/assets/19283907/7a414a41-acb6-49f3-aa3a-1c96ea2a06bf

### Reproducer

https://codesandbox.io/s/primereact-demo-forked-vtn4kg

### PrimeReact version

10.0.2

### React version

17.x

### Language

TypeScript

### Build / Runtime

Create React App (CRA)

### Steps to reproduce the behavior

1. Create a DataTable with multiple columns that support filtering and sorting.
2. Set the property showFilterMenu={false} for the first column.
3. Sort the second column.
4. Observe that the filter menu icon in the third column has disappeared.

### Expected behavior

The filter menu icons shouldn't disappear.

melloware:
Weird the SVG is in the DOM it just not somehow being displayed.

![image](https://github.com/primefaces/primereact/assets/4399574/6acb7580-37c7-487b-810b-c54fa652d7cc)

marinashe:
maybe it can help...
What I found, that if change clip-path in disappeared icon to another, icon will appear again.
For example (see video), if I will change #pr_icon_clip_5 to #pr_icon_clip_15, icon will appear.

https://github.com/primefaces/primereact/assets/19283907/dfb0ee8d-8170-49b1-90e7-0e687c3cc991

melloware:
Thanks for the tip I got to the bottom of it.

PR #5803, opened 2024-01-18
Fix #5802: Allows clear filter button to render without filterMenu button

### LINKED ISSUE #5802: Datatable Column Filter: Clear Button will not show if filter menu button is hidden
### Describe the bug

In datatable, the clear filter button will now render unless filter button is also rendered

It appears that there may have been an erroneous change in this commit: https://github.com/primefaces/primereact/commit/deec2b2db26ebcfc98f56ba6b9cd2c31d6ac3b69 

### Reproducer

https://stackblitz.com/edit/vqsqhi?file=src%2FApp.jsx

### PrimeReact version

10.3.3

### React version

18.x

### Language

TypeScript

### Build / Runtime

Create React App (CRA)

### Browser(s)

_No response_

Scroll down to the name column and see that showClearButton is set to true and showFilterMenu is false

### Expected behavior

If showClearButton is set to true and showFilterMenu is false, the clear button should still show

cassenhp:
Fix #5802

Fixes Issue where clear button will not be rendered if menu button is not rendered. In original issue fix listed below, I think that the check implemented in this issue was unintentionally placed in the createClearButton method

In reference to this issue:
https://github.com/primefaces/primereact/issues/4994

And this commit:
https://github.com/primefaces/primereact/commit/deec2b2db26ebcfc98f56ba6b9cd2c31d6ac3b69

### DIFF
--- a/components/lib/datatable/ColumnFilter.js
+++ b/components/lib/datatable/ColumnFilter.js
@@ -563,10 +563,6 @@ export const ColumnFilter = React.memo((props) => {
     };
 
     const createClearButton = () => {
-        if (!showMenuButton()) {
-            return null;
-        }
-
         const filterClearIconProps = mergeProps(
             {
                 'aria-hidden': true


components/lib/datatable/ColumnFilter.js
```diff
@@ -563,10 +563,6 @@ export const ColumnFilter = React.memo((props) => {
     };
 
     const createClearButton = () => {
-        if (!showMenuButton()) {
```
melloware, line 566: This should check !showClearButton which doesn't exist and needs to be created next to the showMenuButton.

melloware (changes requested):

components/lib/datatable/ColumnFilter.js
lbathp: > This should check !showClearButton which doesn't exist and needs to be created next to the showMenuButton.

Doesn't [this line](https://github.com/primefaces/primereact/pull/5803/files/85a85ca43f303a67a5f91781c7aae86624602ec7#diff-b816aae7f5071fd81bcae0d885dbe0b2bbae790ce43618e2a26e2b52e6f076adR575) (line 579/575) already check if the clear button should be rendered or not?

melloware: It does but its now not consistent with the MenuButton logic.  It should check it first and short circuit the whole method. Its cleaner and consistent. I am big on consistency!

melloware: I can fix it if you don't have time.  Just let me know.

lbathp: @melloware I know @cassenhp is sick and I do not have access to their fork - so that would be appreciated. Otherwise I can create my own fork and make the changes needed.

melloware: @lbathp i will take care of it!  Thanks!

PR #5839, opened 2024-01-25
Enhancement: DataTable accessibility

nitrogenous:
https://github.com/primefaces/primereact/issues/5799

### DIFF
--- a/components/lib/datatable/BodyRow.js
+++ b/components/lib/datatable/BodyRow.js
@@ -162,37 +162,30 @@ export const BodyRow = React.memo((props) => {
             const { target, currentTarget: row } = event;
 
             switch (event.code) {
-                //down arrow
                 case 'ArrowDown':
                     onArrowDownKey(row, event);
                     break;
 
-                //up arrow
                 case 'ArrowUp':
                     onArrowUpKey(row, event);
                     break;
 
-                //home
                 case 'Home':
                     onHomeKey(row, event);
                     break;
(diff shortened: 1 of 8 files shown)


components/lib/datatable/BodyRow.js
```diff
@@ -152,44 +164,37 @@ export const BodyRow = React.memo((props) => {
             switch (event.which) {
                 //down arrow
                 case 40:
```
melloware, line 166: We should switch those to event.code like 'ArrowDown' etc. the old event.which is deprecated

components/lib/datatable/HeaderCell.js
```diff
@@ -169,7 +169,7 @@ export const HeaderCell = React.memo((props) => {
     };
 
     const onKeyDown = (event) => {
-        if (event.key === 'Enter' && event.currentTarget === elementRef.current && DomHandler.getAttribute(event.currentTarget, 'data-p-sortable-column') === 'true') {
+        if ((event.which == 13 || event.which == 32) && event.currentTarget === elementRef.current && DomHandler.getAttribute(event.currentTarget, 'data-p-sortable-column') === 'true') {
             onClick(event);
```
melloware, line 173: I don't think we want to use event.which anymore it should all be event.code right?

melloware (changes requested):

gucal (approved):

PR #5852, opened 2024-01-29
Fix #5849: Splitter improve ARIA accessibility

### LINKED ISSUE #5849: Splitter: missing/incorrect accessibility features
### Describe the bug

I refer to the detailed specs outlined here: https://www.w3.org/WAI/ARIA/apg/patterns/windowsplitter/

the incorrect one is the use of aria-orientation which while the component prop refers to the orientation of the panels themselves, the aria-orientation should be the orientation of the separator, i.e., always the inverse of the component orientation.

the missing are:

role="separator" on the separator
aria-valuemin and aria-valuemax on the separator
aria-controls on the separator with a value that points to the primary panel
props should be added to the component to allow setting of aria-label and/or aria-labeledby on the separator

and then ideally add to the keyboard handling support for the Home, End and Enter keys

### Reproducer

https://primereact.org/splitter/

### PrimeReact version

10.3.3

### React version

17.x

### Language

ALL

### Build / Runtime

Vite

### Browser(s)

_No response_

The behavior is reproducible in any example of the Splitter component, including the samples in Prime React documentation at https://primereact.org/splitter/

### Expected behavior

that the separator includes the following attributes in the rendered HTML:

1. aria-orientation with a value that represents the direction of the separator itself (i.e., the inverse of the Splitter prop orientation)
2. aria-valuemin with a value that represents the minimum size of the primary panel (0 default)
3. aria-valuemax with a value that represents the maximum size of the primary panel (100 default)
4. role with the value separator
5. aria-controls with a value that points to a unique ID of the primary panel

The <Splitter> component should have two 2 props, aria-label and aria-labelledby (note that the spec spelling uses labelled and not labeled), the values of these should be passed through to identically named attributes on the separator

Implement additional keyboard functionality:

1. Home - maximizes the primary panel
2. End - minimizes the primary panel
3. Enter - toggles the primary panel between minimum and maximum sizes

melloware:
Fix #5849: Splitter improve ARIA accessibility

### DIFF
--- a/components/doc/splitter/accessibilitydoc.js
+++ b/components/doc/splitter/accessibilitydoc.js
@@ -50,6 +50,24 @@ export function AccessibilityDoc() {
                                 </td>
                                 <td>Moves a vertical splitter to the right.</td>
                             </tr>
+                            <tr>
+                                <td>
+                                    <i>home</i>
+                                </td>
+                                <td>Maximizes the primary panel.</td>
+                            </tr>
+                            <tr>
+                                <td>
+                                    <i>end</i>
+                                </td>
+                                <td>Minimizes the primary panel.</td>
+                            </tr>
+                            <tr>
+                                <td>
+                                    <i>enter</i>
+                                </td>
+                                <td>Toggles the primary panel between minimum and maximum sizes.</td>
+                            </tr>
                         </tbody>
                     </table>
                 </div>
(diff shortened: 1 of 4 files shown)


sharvell-collegeboard (approved):
this looks awesome, thanks for amazingly fast work!!!

melloware:
Thanks fixed `role` with https://github.com/primefaces/primereact/pull/5853/files

Issue #5858, opened 2024-01-30
Menu: first menu item selected on popup


peconomou929:
### Describe the bug

If I create a menu component with submenu headers and menu items, then clicking on any of the submenu headers will activate the first menu item in the menu. This is shown in the image, taken directly from the PrimeReact website.

<img width="215" alt="Screenshot 2024-01-30 at 7 04 27 PM" src="https://github.com/primefaces/primereact/assets/65582486/9cd12a45-e2d4-4f1b-902a-e9ed3e870074">

Take a look at the steps to reproduce the behaviour below.

### Reproducer

https://primereact.org/menu/#popup

### PrimeReact version

10.3.3

### React version

18.x

### Language

TypeScript

### Build / Runtime

Vite

### Steps to reproduce the behavior

1. Go to [this subheading](https://primereact.org/menu/#group) of the documentation for the PrimeReact Menu.
2. In the example, click on the header "Profile."
3. Observe as the menu item "New" is activated (given a background color).

You will see two changes associated with this in the HTML:
1. if you look at the menu item "New," the attribute "data-p-focused" is set to true.
2. if you look at the whole menu element, it is given an attribute "aria-activedescendants". 

### Expected behavior

In the menu component, a menu items is activated only when the user clicks on the item itself.

melloware:
Looking here at accessibility: https://primereact.org/menu/#accessibility  the tab key states "Add focus to the first item if focus moves in to the menu. If the focus is already within the menu, focus moves to the next focusable item in the page tab sequence."

But i wonder if that first element is meant to be focused or not?

peconomou929:
I see, okay. I didn't notice that.

But still, if there is already a menu item in focus, and I click on one of the headers, the focus is brought to the first item. Is this the desired behaviour? Even taking the note about accessibility into consideration, this still feels like a bug, or at least unexpected behaviour.

akshayaqburst:
@melloware I feel like @peconomou929 is raising a valid bug and have created a fix for the same. PR for review - https://github.com/primefaces/primereact/pull/5861.

Thanks,

melloware:
Ok I will test your PR today. I will also check what PrimeVue does.

melloware:
PrimeVue behaves the same way: https://primevue.org/menu/

I think this is intentional because according to ARIA specs when a popup is displayed it should be given focus or the first item should be given.

According to ARIA specs: https://www.w3.org/WAI/ARIA/apg/patterns/menu-button/

[...]

PR #5861, opened 2024-01-31
fix:primereact #5858, Menu: first menu item selected when any header is clicked

akshayaqburst:
### Defect Fixes
fix:primereact #5858, Menu: first menu item selected when any header is clicked


melloware:
See comments on why the current behavior is valid and correct.

PR #5871, opened 2024-02-01
Fix primefaces#5863: Calendar enabledDates fixed

sdwillbrand:
@melloware there was a bug when only ```enabledDates``` is present, it would *disable* them instead of only enabling them. If only ```enabledDates```  is present, then the first two if-branches would not be entered. ```isDisabled``` would be set to ```true``` if the date is in the list of enabled dates. This would ultimately set ```isDisabled``` to true which would disable the dates.

### DIFF
--- a/components/lib/calendar/Calendar.js
+++ b/components/lib/calendar/Calendar.js
@@ -2308,8 +2308,11 @@ export const Calendar = React.memo(
             // last check for enabled dates to force dates enabled
             if (props.enabledDates) {
                 const isEnabled = props.enabledDates.some((d) => d.getFullYear() === year && d.getMonth() === month && d.getDate() === day);
-
-                isDisabled = isDisabled ? !isEnabled : isEnabled;
+                if (isEnabled) {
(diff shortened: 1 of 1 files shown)


components/lib/calendar/Calendar.js
```diff
@@ -2406,14 +2409,14 @@ export const Calendar = React.memo(
...
-                        iFormat++;
-                    }
+                if (matches) {
+                    iFormat++;
+                }
 
-                    return matches;
-                },
+                return matches;
+            },
```
melloware, line 2419: What is this change?

melloware (changes requested):

sdwillbrand:
@melloware should I open another issue or are you going to open it again?

melloware:
i reopened it and linked this PR.

melloware:
looks like you need to run `npm run format` as well.

melloware:
Running the build now...

sdwillbrand:
@melloware Thank you for your support!

PR #5926, opened 2024-02-08
Improve TypeScript typings for zIndex option

dchenk:
I'm not entirely sure this change is correct, but the idea here is to change the TypeScript typings for global `zIndex` option by making all properties optional. This way you can specify the z-index for just one component type (e.g., `modal`) if that's all you need to customize.


melloware:
I think this is correct.

melloware (approved):

PR #5995, opened 2024-02-17
BodyRows merge rowClassName properly #5983

### LINKED ISSUE #5983: DataTable: rowClassName don't take prevalence over passthrough tailwind
### Describe the bug

The rowClassName attributes works and add the class `bg-primary` to row but the pt define some class `bg-white` that are applied after our customization.
We now have `bg-primary bg-white`

The step 4 of https://primereact.org/tailwind/#unstyledmode doesn't help (my twMerge is never called)

### Reproducer

https://stackblitz.com/edit/d8kwfr?file=src%2Fmain.jsx,src%2FApp.jsx

### PrimeReact version

10.5.1

### React version

18.x

### Language

ALL

### Build / Runtime

Vite

### Browser(s)

ALL

1. I took the exemple in https://primereact.org/datatable/#conditional_style
2. in main.jsx
```diff
+import Tailwind from 'primereact/passthrough/tailwind';
```

```diff
-    <PrimeReactProvider>
+    <PrimeReactProvider value={{ pt: Tailwind }}>
```

### Expected behavior

Class added with local customization should be visible instead of global classes.
My twMerge is used to resolve a conflict
I don't have to rewrite `bodyrow` to hide this bug. (this is my actual workaround)

Et7f3:
Fix #5983

Thanks for your quick reply and code indication @melloware,

[Even though it works](https://github.com/primefaces/primereact/issues/5983#issuecomment-1946902417)! I digged a bit and found the issue to happened elsewhere (the `mergeProps` function).

What is done in this PR:
- add the `classNameMergeFunction` props because `context.classNameMergeFunction` was undefined (and from docs it seems to be a left over of an old doc where it wasn't nested)
- re-order class by customisation: custom last so tailwind-merge works. I could keep the old line instead of `classNames(cx('bodyRow', { rowProps: props }), rowClassName)` but I think it could allow to override `bg-row-odd` but I am not sure.

This PR doesn't have test because I don't know how to write them (I am not a front end guy). I just tested by modifying the documentation example and see if it works.

before/after:
```diff
-bg-white text-gray-600 dark:bg-gray-900 transition duration-200 focus:outline focus:outline-[0.15rem] focus:outline-blue-200 focus:outline-offset-[-0.15rem] dark:text-white/80 dark:focus:outline dark:focus:outline-[0.15rem] dark:focus:outline-blue-300 dark:focus:outline-offset-[-0.15rem]`
+         text-gray-600 dark:bg-gray-900 transition duration-200 focus:outline focus:outline-[0.15rem] focus:outline-blue-200 focus:outline-offset-[-0.15rem] dark:text-white/80 dark:focus:outline dark:focus:outline-[0.15rem] dark:focus:outline-blue-300 dark:focus:outline-offset-[-0.15rem] bg-primary
```

PS: you know you can copy permalink so you can directly jumps to code
https://github.com/primefaces/primereact/blob/0083b8dea35159f6338e9cc003b4f991a067e555/components/lib/datatable/BodyRow.js#L471

### DIFF
--- a/components/lib/datatable/BodyRow.js
+++ b/components/lib/datatable/BodyRow.js
@@ -468,7 +468,7 @@ export const BodyRow = React.memo((props) => {
         {
             role: 'row',
             tabIndex: tabIndex,
-            className: classNames(rowClassName, cx('bodyRow', { rowProps: props })),
+            className: classNames(cx('bodyRow', { rowProps: props })),
             style: style,
             onMouseDown: (e) => onMouseDown(e),
             onMouseUp: (e) => onMouseUp(e),
@@ -491,7 +491,10 @@ export const BodyRow = React.memo((props) => {
             'data-p-highlight': props.selected,
             'data-p-highlight-contextmenu': props.contextMenuSelected
         },
-        getBodyRowPTOptions('bodyRow')
+        getBodyRowPTOptions('bodyRow'),
+        {
+            className: classNames(rowClassName)
+        }
     );
 
     return <tr {...rowProps}>{content}</tr>;
--- a/components/lib/hooks/useMergeProps.js
+++ b/components/lib/hooks/useMergeProps.js
@@ -10,7 +10,7 @@ export const useMergeProps = () => {
 
     return (...props) => {
         const options = {
-            ...(context?.ptOptions?.classNameMergeFunction && { classNameMergeFunction: context.ptOptions.classNameMergeFunction })
+            classNameMergeFunction: context?.ptOptions?.classNameMergeFunction
         };
 
         return _mergeProps(props, options);
--- a/components/lib/hooks/useMergeProps.js
+++ b/components/lib/hooks/useMergeProps.js
@@ -10,7 +10,7 @@ export const useMergeProps = () => {
 
     return (...props) => {
         const options = {
-            ...(context?.ptOptions?.classNameMergeFunction && { classNameMergeFunction: context.classNameMergeFunction })
+            ...(context?.ptOptions?.classNameMergeFunction && { classNameMergeFunction: context.ptOptions.classNameMergeFunction })
         };
 
         return _mergeProps(props, options);


components/lib/hooks/useMergeProps.js
```diff
@@ -10,7 +10,7 @@ export const useMergeProps = () => {
 
     return (...props) => {
         const options = {
-            ...(context?.ptOptions?.classNameMergeFunction && { classNameMergeFunction: context.classNameMergeFunction })
+            ...(context?.ptOptions?.classNameMergeFunction && { classNameMergeFunction: context.ptOptions.classNameMergeFunction })
```
melloware, line 13:
```suggestion
            ...(context?.ptOptions?.classNameMergeFunction)
```

Should it just be the above?  I am not sure what its appending classNameMergeFunction to itself again now?

components/lib/datatable/BodyRow.js
```diff
@@ -465,10 +465,11 @@ export const BodyRow = React.memo((props) => {
     const content = createContent();
     const tabIndex = getTabIndex();
     const rowProps = mergeProps(
+        getBodyRowPTOptions('bodyRow'),
```
melloware, line 468: Does this really need to be first now?

melloware (changes requested):

components/lib/datatable/BodyRow.js
Et7f3: It is needed because otherwise we have bg-primary before bg-white and tailwind merge keep last.

components/lib/hooks/useMergeProps.js
Et7f3: Should be syntactically equivalent. Let me check.

components/lib/datatable/BodyRow.js
Et7f3: Or I can keep in second place and add rowClassName at third place so we minimise the change.

melloware (approved):

melloware:
@EfrainTlapale this broke normal non-passthrough styling see #6194

melloware:
Nevermind I see what broke it and it wasn't this change @Et7f3 !

PR #6012, opened 2024-02-19
Fix #6000: Improved closeOnEscape handling

### LINKED ISSUE #6000: useDisplayOrder: clears incorrect
### Describe the bug

The useDisplayOrder hooks clears out an incorrect id when unmounting, so very simple Dialogs configurations throw "Unexpected: global esc key listener with priority [300, x] already exists." errors", even if not using the 'closeOnEscape' option (if they are not destroyed in the reverse order they become visible).

https://github.com/primefaces/primereact/blob/d5625feea80e871ca662a72d450dcd0659967739/components/lib/hooks/useDisplayOrder.js#L22

Here, the information is deleted taking the generated displayOrder as an index, but that will always be off by 1.

In the attached example, te following happens in this order:
- Dialog "one" is created visible, is asigned displayOrder 1.
- Dialog "two" becomes visible, is asigned displayOrder 2.
- Dialog "one" is hidden. Because of this bug, the information for Dialog "two" is removed, as the value of groupToDisplayedElements[group] is [dialogOne, dialogTwo], and deleted the value with index 1.
- Dialog "three" becomes visible, is asigned displayOrder 2.

Finally, because there are two visible Dialogs with displayOrder 2, the application crashes.

The fix should be pretty trivial, just delete the item with index `newDisplayOrder - 1`. Also, i think the global Esc handler should not be added if `closeOnEscape` is `false`.

### Reproducer

https://stackblitz.com/edit/vitejs-vite-e88fmg

### PrimeReact version

10.5.1

### React version

17.x

### Language

ES6

### Build / Runtime

Create React App (CRA)

### Browser(s)

_No response_

Just load the page, and look for the "Unexpected: global esc key listener with priority [300, 2] already exists." error.

### Expected behavior

Should not crash.

melloware:
Fix #6000: Improved closeOnEscape handling

1. Removed the unnecessary check if (!(group in groupToDisplayedElements)) since we're checking if groupToDisplayedElements[group] exists directly.
2. Simplified the calculation of newDisplayOrder by using the push method, which returns the new length of the array.
3. Removed the deletion of elements by index and replaced it with finding the index of the element using indexOf and then removing it using splice.
4. Removed the cleanup logic to reduce the length of the array since it seems unnecessary considering we are setting setDisplayOrder(undefined).

This version should be cleaner and more concise while maintaining the same functionality.

### DIFF
--- a/components/lib/hooks/useDisplayOrder.js
+++ b/components/lib/hooks/useDisplayOrder.js
@@ -9,21 +9,22 @@ export const useDisplayOrder = (group, isVisible = true) => {
 
     React.useEffect(() => {
         if (isVisible) {
-            if (!(group in groupToDisplayedElements)) {
+            if (!groupToDisplayedElements[group]) {
                 groupToDisplayedElements[group] = [];
             }
 
-            const newDisplayOrder = groupToDisplayedElements[group].length + 1;
+            const newDisplayOrder = groupToDisplayedElements[group].push(uid);
 
-            groupToDisplayedElements[group].push(uid);
             setDisplayOrder(newDisplayOrder);
 
             return () => {
-                delete groupToDisplayedElements[group][newDisplayOrder];
-                const lastOrder = groupToDisplayedElements[group].findLastIndex((el) => el !== undefined);
+                delete groupToDisplayedElements[group][newDisplayOrder - 1];
 
                 // Reduce array length, by removing undefined elements at the end of array:
-                groupToDisplayedElements[group].splice(lastOrder + 1);
+                const lastIndex = groupToDisplayedElements[group].length - 1;
+                const lastOrder = groupToDisplayedElements[group].findLastIndex((el) => el !== undefined);
+
+                if (lastOrder !== lastIndex) groupToDisplayedElements[group].splice(lastOrder + 1);
 
                 setDisplayOrder(undefined);
             };


Manizuca (commented):
Hi!

This looks cleaner. It still has a problem though.

Consider this case:

* useDisplayOrder is called for a Dialog with uid "pr-1". It returns 1.
* useDisplayOrder is called for a Dialog with uid "pr-2". It returns 2.
* useDisplayOrder is called for a Dialog with uid "pr-3". It returns 3.

At this point, `groupToDisplayedElements[group]` will be `["pr-1", "pr-2", "pr-3"]`.

* Now, consider the case the Dialog with uid "pr-2" unmounts. Since `Array.splice()` removes the given index, and moves the items after that. That means that after unmounting `groupToDisplayedElements[group]` will be `["pr-1", "pr-3"]`.

* Finally, useDisplayOrder is called for a Dialog with uid "pr-4". It returns 3, since there are currently 2 items at the array. At this point, you will have 2 items with displayOrder 3, which could cause issues with the global esc key listener.

I think the cleanup function should be like:
```js
const index = groupToDisplayedElements[group].indexOf(uid);
delete groupToDisplayedElements[group][index]; // or just newDisplayOrder - 1

// Reduce array length, by removing undefined elements at the end of array:
const lastIndex = groupToDisplayedElements[group].length - 1:
const lastOrder = groupToDisplayedElements[group].findLastIndex((el) => el !== undefined);
if (lastOrder !== lastIndex)  groupToDisplayedElements[group].splice(lastOrder + 1);
```

melloware:
OK PR updated.

Manizuca (approved):
LGTM

PR #6040, opened 2024-02-24
Fix #6039: splitter bug fixes 

### LINKED ISSUE #6039: Splitter: Accessibility updates
### Describe the bug

There are a few small bugs in the Splitter component:

- role="separator" is on the wrong element, it should be on the gutter handle element where the other aria attributes are located (in particular, the aria-orientation attribute)
- the documentation indicates that the step property belongs to SplitterPanel, however when you look at the code it's clear that really it goes on Splitter itself, this is simply a bug in the typescript typing file that results in the documentation appearing incorrect
- if you use minSize on a SplitterPanel, when you try to use Home/End/Enter keys for maximize/minimize behavior, then the arrow keys stop working

I've actually coded fixes for this in my local cloned repo that seem to be working great BUT I am unable to push a branch to the repo to create a PR as GitHub says I do not have write access.  If I could be given that access I'd be glad to push and PR OR alternatively would be glad to somehow share the code changes I made with somone who can do it.

### Reproducer

_No response_

### PrimeReact version

10.5.1

### React version

17.x

### Language

TypeScript

### Build / Runtime

Create React App (CRA)

### Browser(s)

_No response_

_No response_

### Expected behavior

_No response_

shawnharvell:
Fix #6039 

for issue #6039

- move step typing to correct element

- move role=separator to correct element (per spec should be on the same element that contains aria-orientation and aria-value* attribs) - see https://www.w3.org/TR/wai-aria-1.2/#separator

- fix bug where if home/end/enter were used in a splitter that has a minSize, after pressing those buttons, the arrow keys would no longer work

for this issue https://github.com/primefaces/primereact/issues/6039

### DIFF
--- a/components/lib/splitter/Splitter.js
+++ b/components/lib/splitter/Splitter.js
@@ -253,24 +253,24 @@ export const Splitter = React.memo(
                 }
 
                 case 'Home': {
-                    resizePanel(index, 100, minSize);
+                    resizePanel(index, 100 - minSize, minSize);
 
                     event.preventDefault();
                     break;
                 }
 
                 case 'End': {
-                    resizePanel(index, minSize, 100);
+                    resizePanel(index, minSize, 100 - minSize);
 
                     event.preventDefault();
                     break;
                 }
 
                 case 'Enter': {
-                    if (prevSize.current > 100 - (minSize || 5)) {
-                        resizePanel(index, minSize, 100);
+                    if (prevSize.current >= 100 - (minSize || 5)) {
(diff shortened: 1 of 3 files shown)


melloware (approved):
I tested these locally and they look great

shawnharvell:
@melloware do I need to do anything more at this point, I was assuming it's all in the core team's hands now but thought I should double check.  Thanks!

melloware:
Yep the core team will review and merge it for me it looks perfect.

nitrogenous (approved):

PR #6060, opened 2024-02-29
Fix #6055: DomHandler incorrect case for computed styles

### LINKED ISSUE #6055: DomHandler.js: getComputedStyle getPropertyValue should be in hyphen case.
### Describe the bug

`scrollInView` in `DomHandler` does not scroll to the very top of dropdown list if that list has top border or padding.
This is happening because `getPropertyValue` gets values using camel case (`borderTopWidth`) instead of hyphen case (`border-top-width`)
Additionally, there the same problem in `getScrollableParents` function with `overflowX` and `overflowY`.

### Reproducer

_No response_

### PrimeReact version

10.5.1

### React version

17.x

### Language

TypeScript

### Build / Runtime

Create React App (CRA)

### Browser(s)

any

1. Make a custom theme where dropdown lists have either top border or top padding.
2. Try to keyboard navigate such list with more items that fits into dropdown.
3. When navigating up list is not scrolled to the very top.

### Expected behavior

Changing `getPropertyValue` arguments to hyphen case resolves the problem.

melloware:
Fix #6055: DomHandler incorrect case for computed styles


nitrogenous (approved):
nıce

PR #6081, opened 2024-03-04
Fix: Primereact uses document.getElementById(undefined) for appending…

### LINKED ISSUE #6079: Primereact uses `document.getElementById(undefined)` for appending component styles
Primereact components styles are being appended into `document.getElementById(undefined)` if some exists on the page. 
Take a look inside `useStyle` hook this line:
https://github.com/primefaces/primereact/blob/a2a2f11716fc726bd929e4ad775fe2a1e000e55c/components/lib/hooks/useStyle.js#L24
```js
// First on line 13
const { /** ... */ id = undefined, /** ... *// } = options; // Here we can get undefined value

// And later on line 24 inside load function:
styleRef.current =
  styleContainer.querySelector(`style[data-primereact-style-id="${name}"]`) // When hook is called first time
  || document.getElementById(id) // document.getElementById(undefined) is used here... :(
  || document.createElement('style');
```

PS: Im using primereact inside Chrome Extension. My content script is being executed on various pages on the internet and uses primereact inside. I don't feel comfortable, removing `document.getElementById(undefined)` elements from the page to fix this bug. Please, guys, add some undefined check before using `getElementById `selector in this hook. :)

### Reproducer

https://stackblitz.com/edit/vitejs-vite-ztm2wz

### PrimeReact version

10.5.1

### React version

18.x

### Language

TypeScript

### Build / Runtime

Vite

### Browser(s)

_No response_

1. Create `<div id="undefined"></div>` element on the page
2. Primereact now appends styles to this element instead of expected one

### Expected behavior

Primereact keeps using `document.head` or provided `styleContainer` for appending styles.

kl-nevermore:
Fix #6079

### DIFF
--- a/components/lib/hooks/useStyle.js
+++ b/components/lib/hooks/useStyle.js
@@ -21,7 +21,7 @@ export const useStyle = (css, options = {}) => {
 
         const styleContainer = context?.styleContainer || document.head;
 
-        styleRef.current = styleContainer.querySelector(`style[data-primereact-style-id="${name}"]`) || document.getElementById(id) || document.createElement('style');
+        styleRef.current = styleContainer.querySelector(`style[data-primereact-style-id="${name}"]`) || document.querySelector(`#${id}`) || document.createElement('style');
 
         if (!styleRef.current.isConnected) {
             styleRef.current.type = 'text/css';


components/lib/hooks/useStyle.js
```diff
@@ -21,7 +21,7 @@ export const useStyle = (css, options = {}) => {
 
         const styleContainer = context?.styleContainer || document.head;
 
-        styleRef.current = styleContainer.querySelector(`style[data-primereact-style-id="${name}"]`) || document.getElementById(id) || document.createElement('style');
+        styleRef.current = styleContainer.querySelector(`style[data-primereact-style-id="${name}"]`) || document.querySelector(`#${id}`) || document.createElement('style');
```
yikstinis, line 24: This change won't fixe the issue. New selector `document.querySelector('#undefined')` will select the same element with `id="undefined"`.. :(

yikstinis: I belive, you something like that:
```js
const getCurrentStyleRef = () => {
  const existingStyleContainerChild = styleContainer.querySelector(
    `style[data-primereact-style-id="${name}"]`,
  );
  if (existingStyleContainerChild) {
    return existingStyleContainerChild;
  }
  // First, check that id is defined!
  if (id !== undefined) {
    // And select element by defined id on the page after:
    const existingElementWithId = document.getElementById(id);
    if (existingElementWithId) {
      return existingElementWithId;
    }
  }
  return document.createElement('style');
};
styleRef.current = getCurrentStyleRef();
```

kl-nevermore: Yes, I made a mistake, sorry. I forgot add during the testing with #

PR #6087, opened 2024-03-05
Fix: Editor: onTextChange handler does not reflect updated props

### LINKED ISSUE #6067: Editor: onTextChange handler does not reflect updated props
### Describe the bug

The onTextChange handler of the PrimeReact Editor component appears to be caching the initial value of props passed to its containing component, leading to inconsistencies in behavior when those props change.

### Reproducer

https://stackblitz.com/edit/ylsaqj?file=src%2FApp.jsx

### PrimeReact version

10.5.1

### React version

18.x

### Language

TypeScript

### Build / Runtime

Vite

### Browser(s)

_No response_

1) Attempt to access a prop value inside onTextChange handler of `<Editor />`.

from the stackblitz example:

1) click on the add button a couple of times, to update the count.
2) type in both the textarea and in the Editor.
3) check the console: onChange can access the current value. OnTextChange reports the initial value.

### Expected behavior

You should be able to access the current prop values. Instead, props return their initial value.

kl-nevermore:
Fix #6067

### DIFF
--- a/components/lib/editor/Editor.spec.js
+++ b/components/lib/editor/Editor.spec.js
@@ -9,6 +9,7 @@ describe('Editor', () => {
     // https://github.com/primefaces/primereact/issues/6067
     test('onTextChange handler does not reflect updated props', async () => {
         const consoleSpy = jest.spyOn(console, 'log').mockImplementation();
+
         function BasicDemo() {
             const [state, setState] = useState(1);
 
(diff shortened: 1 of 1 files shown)


melloware:
Hmmm I like simple but i also like how powerful that useEvent hook is an could be useful in the future?

kl-nevermore:
https://github.com/reactjs/react.dev/pull/5373
Yes, it is needed many times, but there have been some changes and it has not been updated in a long time.

kl-nevermore:
use simple writing seems to be no problem, but `onTextChange` will not be executed under jest.

melloware:
OK then lets go with the simplest changes?

kl-nevermore:
resubmitted a PR and fixed test case

PR #6104, opened 2024-03-07
Fix #6101: Slider: event.value in onSlideEnd returns null on touch devices

### LINKED ISSUE #6101: Slider: Null event.value within the onSlideEnd
### Describe the bug

On **Chrome** based browser and **Touch mode**, when utilizing the Slider component along with the onSlideEnd event, it's observed that the event.value is always null.

### Reproducer

https://codesandbox.io/p/sandbox/primereact-slider-new-value-bug-sv3gwp?file=%2Fsrc%2FApp.js

### PrimeReact version

10.5.1

### React version

18.x

### Language

ES6

### Build / Runtime

Create React App (CRA)

### Browser(s)

Chrome

1. Open the sandbox with a Chrome based browser
2. Open the device toolbar and choose a touch based device
3. Move the slider -> The value of the event is always null

### Expected behavior

The value of the event must be filled with the current value of the slider.

kl-nevermore:
Fix #6101

- The event of `touchend` does not contain `touches`
- Use `changedTouches` with reference to `MUI`
-  query via [caniuse](https://caniuse.com/?search=changedTouches) is not supported on `safari`
    -  but tested in iOS17 and iOS15.8, it worked

### DIFF
--- a/components/lib/slider/Slider.js
+++ b/components/lib/slider/Slider.js
@@ -19,6 +19,7 @@ export const Slider = React.memo(
         const initY = React.useRef(0);
         const barWidth = React.useRef(0);
         const barHeight = React.useRef(0);
+        const touchId = React.useRef();
         const value = props.range ? props.value || [props.min, props.max] : props.value || props.min || 0;
         const horizontal = props.orientation === 'horizontal';
         const vertical = props.orientation === 'vertical';
@@ -69,6 +70,8 @@ export const Slider = React.memo(
 
                 props.onSlideEnd && props.onSlideEnd({ originalEvent: event, value: newValue });
 
+                touchId.current = undefined;
+
                 unbindDocumentMouseMoveListener();
                 unbindDocumentMouseUpListener();
                 unbindDocumentTouchMoveListener();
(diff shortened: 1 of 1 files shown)


components/lib/slider/Slider.js
```diff
@@ -156,11 +163,32 @@ export const Slider = React.memo(
             barHeight.current = elementRef.current.offsetHeight;
         };
 
+        const trackFinger = (event) => {
```
melloware, line 166:
```suggestion
        const trackTouch= (event) => {
```

melloware:
OK then just rename to `trackTouch` and I think we are good.

kl-nevermore:
maybe?
```tsx
 const trackTouch = (_event) => {
            const event = Array.from(_event.changedTouches ?? []).find((t) => t.identifier === touchId.current) || _event;

            return {
                pageX: event.pageX,
                pageY: event.pageY
            };
        };
```

melloware:
Oooh I like it!

melloware:
or reverse it called the param `event` and the private variable `const _event` ?

kl-nevermore:
Updated

melloware (approved):

PR #6164, opened 2024-03-15
Fix #6163: p-hidden-accessible modernization

### LINKED ISSUE #6163: Core: p-hidden-accessible modernization
From: https://stackoverflow.com/a/62109988/502366

See also: https://webaim.org/techniques/css/invisiblecontent/#techniques

Boostrap: https://github.com/twbs/bootstrap/blob/main/scss/mixins/_visually-hidden.scss#L9-L16

melloware:
Fix #6163: p-hidden-accessible modernization

From: https://stackoverflow.com/a/62109988/502366

See also: https://webaim.org/techniques/css/invisiblecontent/#techniques

Boostrap: https://github.com/twbs/bootstrap/blob/main/scss/mixins/_visually-hidden.scss#L9-L16

### DIFF
--- a/components/lib/componentbase/ComponentBase.js
+++ b/components/lib/componentbase/ComponentBase.js
@@ -4,14 +4,15 @@ import { ObjectUtils, classNames, mergeProps } from '../utils/Utils';
 
 const baseStyle = `
 .p-hidden-accessible {
-    border: 0;
-    clip: rect(0 0 0 0);
-    height: 1px;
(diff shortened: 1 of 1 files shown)


melloware:
will do but `clip-rect` is for IE and legacy browsers which are not supported anymore right?  The new `clip-path: inset(50%)` is for modern browsers?

melloware:
OK changes made

PR #6189, opened 2024-03-20
MenuBar length spelled wrong

melloware:
MenuBar length spelled wrong


melloware:
Just saw this as i was researching another issue.

Issue #6144, opened 2024-03-13
MultiStateCheckBox: htmlFor prop doesn't work


gorsheninmv:
### Describe the bug

htmlFor prop doesn't work.
An `input` html element seems not to be in there, so `for` attribute doesn't bind to any element.

The behavior is reproduced in the demo examples:

**Combobox works as expected**:
![checkbox](https://github.com/primefaces/primereact/assets/32216772/48968e4a-2bd1-40b3-ac79-9fb56ee9d3e0)

**MultiStateCombobox doesn't work**:
![multiselectcheckbox](https://github.com/primefaces/primereact/assets/32216772/d9802127-cfe8-4887-8eae-3321a8bc2b06)

### PrimeReact version

10.3.3

### React version

18.x

### Language

TypeScript

### Build / Runtime

Vite

charithAmila:
Hi @melloware 
Is anyone working on this? I would like to raise a PR.

melloware:
Is it me or in the showcase the checkbox never checks at all for MultiStateCheckbox?

charithAmila:
> Is it me or in the showcase the checkbox never checks at all for MultiStateCheckbox?

https://github.com/primefaces/primereact/pull/6195#issuecomment-2012669805

nitrogenous:
We are planning to deprecate MultiState and TriState checkboxes thus we will not no longer be giving support for these components. We will share an example about how you can achieve the same behaviours with only checkbox

PR #6195, opened 2024-03-21
fix/#6144 htmlFor prop doesn't work in MultiStateCheckBox

charithAmila:
### Defect Fixes
Fix: #6144 htmlFor prop doesn't work in  MultiStateCheckBox

### DIFF
--- a/components/lib/multistatecheckbox/MultiStateCheckbox.js
+++ b/components/lib/multistatecheckbox/MultiStateCheckbox.js
@@ -121,25 +121,13 @@ export const MultiStateCheckbox = React.memo(
             if (props.id) {
                 const label = document.querySelector(`label[for="${props.id}"]`);
 
-                const clickCheckbox = () => {
-                    if (DomHandler.isExist(elementRef.current)) {
-                        elementRef.current.click();
-                    }
-                };
-
                 if (label) {
                     label.addEventListener('click', () => {
-                        clickCheckbox();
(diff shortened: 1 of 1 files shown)


melloware:
Did you check if PrimeVue has the same issue or has the same
Solution?

melloware:
It looks like more than just this label is broken on the showcase the MultiState checkbox doesn't ever seem to get checked?

melloware (changes requested):
I htink we need to investigate why the icons are not appearing also.

charithAmila:
I will take a look.

charithAmila:
> I htink we need to investigate why the icons are not appearing also.

The icons do not appear because both the background and icon colors are same. 

Dark mode preview:
Screen Shot 2024-03-21 at 11 30 42 AM

Suggestion: 
We should add a default color to the icon or parent element (div)

Results:
1. Added primary color to the parent element.
Screen Shot 2024-03-21 at 11 17 35 AM

2. Added primary color to the icon.
Screen Shot 2024-03-21 at 11 23 47 AM

melloware:
OK I fixed the icons in my PR: https://github.com/primefaces/primereact/pull/6209

melloware (approved):

mertsincan:
Hi,

Similar to https://github.com/primefaces/primereact/pull/6210#issuecomment-2017517413

We'll have added new structures for the next version. For now, closed. Thanks a lot for your PR!

PR #6214, opened 2024-03-22
Fix #6151: fixed maxDate prop validation issue

### LINKED ISSUE #6151: Calendar: AM/PM does not work properly if minDate and maxDate validations are provided as prop
### Describe the bug

When we are using minDate and maxDate validation as prop in the two separate input (which is shown in the demo as From and To fields), then changing the AM to PM does not work. The detailed steps have been provided below to recreate this issue.

### Reproducer

https://stackblitz.com/edit/stackblitz-starters-aemkru?file=src%2FApp.tsx

### PrimeReact version

10.5.1

### React version

17.x

### Language

ES6

### Build / Runtime

Create React App (CRA)

### Browser(s)

Edge 121

1. Go to the above stackblitz demo
2. Change the time in the "From:" field (For example, put time as 02:30 PM)
3. Change the time in the "To:" field (For example, put the time as 06:30 PM)
4. Now, change the "From" field PM to AM and then AM to PM just after it.
5. You can see that while changing it from PM to AM works, but reversing it to original state, which is from AM to PM does not work.

While changing it to its original state - AM to PM still holds the validation true, that is, the FROM time should be less than TO field max time (02:30 AM / 02:30 PM < 06:30 PM)

### Expected behavior

The From field time should be able to change from AM to PM
FROM: 02:30 AM -> 02:30 PM should work if we have provided TO time as 06:30 PM.

shubhsaur:
Fix #6151: fixed maxDate prop validation issue

### DIFF
--- a/components/lib/calendar/Calendar.js
+++ b/components/lib/calendar/Calendar.js
@@ -801,7 +801,7 @@ export const Calendar = React.memo(
             const currentHour = currentTime.getHours();
             const newHour = currentHour >= 12 ? currentHour - 12 : currentHour + 12;
 
-            if (validateHour(convertTo24Hour(newHour, !(currentHour > 11)), currentTime)) {
+            if (validateHour(convertTo24Hour(newHour, currentHour > 11), currentTime)) {
                 updateTime(event, newHour, currentTime.getMinutes(), currentTime.getSeconds(), currentTime.getMilliseconds());
             }
 


shubhsaur:
Fixed #6151: fixed maxDate prop validation issue

melloware (approved):

shubhsaur:
Hello @melloware looks like some lint checks are failing for this, have checked on my local by running lint formatter but still getting the same error, is this fine?

melloware:
They just changed the lint check but don't worry about it. PrimeTek is going to fix them all

mertsincan:

PR #6217, opened 2024-03-23
Fix #5998: fixed minute values updated by hour changes with stepMinute

### LINKED ISSUE #5998: Calendar: When using timeonly and stepMinute prop together, minute is not showing correctly in the input field
### Describe the bug

When we are clicking on the calendar input field and then click on the up or down arrow once for changing hour in the time picker overlay, the minute shown or reflected in the input field and overlay is different and it's also not in the steps of 30 but looks like it is taking the mins from current local time.

Note: By changing only the hour, it is creating the issue, it works fine once we are changing the minutes as well.

### Reproducer

https://stackblitz.com/edit/u49nwk?file=src%2FApp.jsx

### PrimeReact version

10.5.1

### React version

17.x

### Language

ES6

### Build / Runtime

Create React App (CRA)

### Browser(s)

Edge 121

1. Go to the reproducer (stackblitz link above) which is basically forked from prime react official documentation.
2. Click on the input field and then click on the up or down arrow once for changing hour in the time picker overlay.
3. You can see the minute being reflected in the input field and overlay is different and it's also not in the steps of 30 but looks like it is taking the mins from current local time until or unless you are changing minutes.

### Expected behavior

The minutes shown in the input field should be same and in sync with the time picker overlay if we are changing hour only and minutes should be changed according to whatever the stepMinute user has provided, which is 30 in this case.

shubhsaur:
Fix #5998: fixed minute values updated by hour changes with stepMinute

### DIFF
--- a/components/lib/calendar/Calendar.js
+++ b/components/lib/calendar/Calendar.js
@@ -576,6 +576,14 @@ export const Calendar = React.memo(
             }
         };
 
+        const roundMinutesToStep = (minutes) => {
+            if (props.stepMinute) {
+                return Math.round(minutes / props.stepMinute) * props.stepMinute;
+            }
+
+            return minutes;
+        };
+
         const incrementHour = (event) => {
             const currentTime = getCurrentDateTime();
             const currentHour = currentTime.getHours();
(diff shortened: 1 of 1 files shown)


mertsincan:

Issue #6121, opened 2024-03-08
MultiSelect - SelectionLimit


Nandovga:
### Describe the bug

The SelectionLimit property set to 1 allows you to select more than one element.

```
<MultiSelect optionLabel="name"
             options={data}
             optionValue="id"
             selectionLimit={1}
             value={select}
             onChange={e => setSelect(e.value)}/>
```

### PrimeReact version

10.5.1

### React version

18.x

### Language

TypeScript

### Build / Runtime

Vite

### Browser(s)

Chrome

PR #6218, opened 2024-03-24
Fix/#6216 MultiSelect - SelectionLimit issue

### LINKED ISSUE #6391: MultiSelect: selectionLimit does not limit selections
### Describe the bug

When using MultiSelect with a selectionLimit (for example a selection limit of 2) the component still allows me to select an unlimited amount of options. 

### Reproducer

https://stackblitz.com/edit/vitejs-vite-qu7tmj?file=src%2FApp.tsx

### PrimeReact version

10.5.1

### React version

18.x

### Language

TypeScript

### Browser(s)

ALL

1. use multi select
2.  give 3+ options 
3.  add selectionLimit of 2 and select more than two options 

### Expected behavior

This should only allow you to select whatever the limit is set to, but it does not

charithAmila:
### Defect Fixes
Fix: #6121 MultiSelect - SelectionLimit issue

Fix #6391


components/lib/multiselect/MultiSelect.js
```diff
@@ -653,6 +657,10 @@ export const MultiSelect = React.memo(
                 return ObjectUtils.isFunction(props.optionDisabled) ? props.optionDisabled(option) : ObjectUtils.resolveFieldData(option, props.optionDisabled);
             }
 
+            if (props.selectionLimit && props.value && props.value.length >= props.selectionLimit && !isSelected(option)) {
```
mertsincan, line 660: I don't understand why this check is needed. If the user exceeds the selected limit, other options do not need to be disabled.

charithAmila: Hi @mertsincan 
So how should `selectionLimit` work? Can users select more than the selection limit?

XYIAN:
this is an issue for my team as well - when trying to add any selection limit - the component still allows us to select as many as we want. If I have a multi select of 50 and want a limit of 2, adding selectionLimit does not limit users from selecting more than 2 

https://stackblitz.com/edit/vitejs-vite-qu7tmj?file=src%2FApp.tsx

components/lib/multiselect/MultiSelect.js
XYIAN: according to prime react JSDoc selection limit should limit the amount of selections. 

# This is obviously a broken prop

melloware: OK I made some tweaks and thoroughly tested this.  It is in line with how PrimeNG and PrimeVue handle this scenario.   @mertsincan the reason we need to check if the current option is already selected is to allow them to unselect it.  So in the scenario of `selectionLimit={1}` they need to be able to unselect the selected one

image

melloware (approved):

PR #6219, opened 2024-03-24
New components IconField & InputIcon

nitrogenous:
https://github.com/primefaces/primereact/issues/6229


components/lib/iconfield/IconField.js
```diff
@@ -0,0 +1,37 @@
...
+            {
+                className: cx('root', { iconPosition: props.iconPosition })
+            },
+            IconFieldBase.getOtherProps(props),
+            ptm('root')
+        );
+
+        return (
+            <div {...rootProps} ref={elementRef}>
+                <>{props.children}</>
```
melloware, line 31: Do you need the empty react element wrapping the children?  It is already wrapped in a DIV so it should be fine?

melloware (changes requested):

PR #6224, opened 2024-03-25
Fix #6223: RadioButton onChange called twice

### LINKED ISSUE #6223: RadioButton: onChange is being triggered twice.
### Describe the bug

OnChange on RadioButton component event is being triggered twice, and I see that logic in RadioButton.js here https://github.com/primefaces/primereact/blob/master/components/lib/radiobutton/RadioButton.js#L60 and https://github.com/primefaces/primereact/blob/master/components/lib/radiobutton/RadioButton.js#L68 it's being called twice

We should only trigger onChange only once.

### Reproducer

https://stackblitz.com/edit/ugdbqe?file=src%2FApp.jsx

### PrimeReact version

10.5.3

### React version

17.x

### Language

TypeScript

### Build / Runtime

Create React App (CRA)

### Browser(s)

Any

Click on the checkbox above, you should see TWO alerts on change.

### Expected behavior

It should only show alert once on change of radio button.

melloware:
Fix #6223: RadioButton onChange called twice

### DIFF
--- a/components/lib/checkbox/Checkbox.js
+++ b/components/lib/checkbox/Checkbox.js
@@ -56,7 +56,7 @@ export const Checkbox = React.memo(
                     }
                 };
 
-                props.onChange && props.onChange(eventData);
+                props?.onChange?.(eventData);
 
                 // do not continue if the user defined click wants to prevent
                 if (event.defaultPrevented) {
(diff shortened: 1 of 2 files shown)


nitrogenous (approved):
looking fine

PR #6228, opened 2024-03-26
Enhancement focusOnHover Prop

nitrogenous:
Added new 'focusOnHover' prop to components below

1. PickList 
2. OrderList
3. MultiSelect
4. Dropdown


components/lib/picklist/PickListSubList.js
```diff
@@ -78,6 +84,7 @@ export const PickListSubList = React.memo(
                             onMouseDown={(event) => props.onOptionMouseDown({ ...event, index, type: props.type })}
                             ptm={ptm}
                             cx={cx}
+                            onMouseMove={() => changeFocusedItemOnHover(index)}
```
melloware, line 87: This is a total nitpick but I like when the `on` events grouped together so move this 2 lines up under `onMouseDown`  If you don't want to I understand

melloware (changes requested):

components/lib/picklist/PickListSubList.js
nitrogenous: Done

melloware (approved):

PR #6244, opened 2024-03-28
New component stepper stepperpanel

gucal:
New component - Stepper

### DIFF
--- a/components/doc/stepper/pt/wireframe.js
+++ b/components/doc/stepper/pt/wireframe.js
@@ -1,0 +1,12 @@
+import { DocSectionText } from '@/components/doc/common/docsectiontext';
+
+export const Wireframe = (props) => {
+    return (
+        <>
+            <DocSectionText {...props} />
+            <div>
+                <img className="w-full" src="https://primefaces.org/cdn/primevue/images/pt/wireframe-placeholder.jpg" alt="pt_image" />
+            </div>
+        </>
+    );
+};
--- a/components/lib/stepper/Stepper.js
+++ b/components/lib/stepper/Stepper.js
@@ -10,7 +10,7 @@ import { StepperHeader } from './StepperHeader';
 import { StepperSeparator } from './StepperSeparator';
 
 export const Stepper = React.memo(
-    React.forwardRef((inProps) => {
+    React.forwardRef((inProps, ref) => {
         const mergeProps = useMergeProps();
         const context = React.useContext(PrimeReactContext);
         const props = StepperBase.getProps(inProps, context);
@@ -156,6 +156,12 @@ export const Stepper = React.memo(
                 );
             });
         };
+
+        React.useImperativeHandle(ref, () => ({
+            getElement: () => navRef.current,
+            nextCallback: (e) => nextCallback(e, activeStepState),
+            prevCallback: (e) => prevCallback(e, activeStepState)
+        }));
 
         const createPanelContent = () => {
             return stepperPanels().map((step, index) => {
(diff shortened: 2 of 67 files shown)


components/lib/stepper/Stepper.js
```diff
@@ -0,0 +1,334 @@
...
+        };
+
+        const isItemDisabled = (index) => {
+            return props.linear && !isStepActive(index);
+        };
+
+        const updateActiveStep = (event, index) => {
+            setActiveStepState(index);
+
+            props.onChangeStep && props.onChangeStep({ originalEvent: event, index });
```
melloware, line 55: My only thought here is what I have done in other components is give the user a change to say they don't want to change the step because some validation they have or some check.  So someting like this...

```js
const updateActiveStep = (event, index) => {
            props.onChangeStep && props.onChangeStep({ originalEvent: event, index });
            
            // allow user to mark event.preventDefault() to stop this step from advancing
            if (event.defeaultPrevented()) {
                return;
                }
                
               // set the new step
               setActiveStepState(index);
```

melloware (changes requested):

components/lib/stepper/Stepper.js
```diff
@@ -0,0 +1,334 @@
...
+        useHandleStyle(StepperBase.css.styles, isUnstyled, { name: 'stepper' });
+
+        useMountEffect(() => {
+            if (!idState) {
+                setIdState(UniqueComponentId());
+            }
+        });
+
+        const getStepProp = (step, name) => {
+            return step.props ? step.props[name] : undefined;
```
nitrogenous, line 33: return steps?.props?.[name];

components/lib/stepper/Stepper.js
```diff
@@ -0,0 +1,334 @@
...
+                <>
+                    <ul {...navProps}>{items}</ul>
+                    <div {...panelContainerProps}>{createPanelContent()}</div>
+                </>
+            );
+        };
+
+        const createVertical = () => {
+            return stepperPanels().map((step, index) => {
+                const contentRef = React.useRef();
```
nitrogenous, line 226: react hook cannot be called inside a callback

components/lib/stepper/StepperBase.js
```diff
@@ -0,0 +1,116 @@
...
+            })
+    },
+    panelContainer: 'p-stepper-panels',
+    panel: ({ props, isStepActive, index }) =>
+        classNames('p-stepper-panel', {
+            'p-stepper-panel-active': props.orientation === 'vertical' && isStepActive(index)
+        })
+};
+
+const styles = `
```
nitrogenous, line 36: i believe we can move this styles to react sass theme

components/lib/stepper/Stepper.js
```diff
@@ -0,0 +1,334 @@
+import * as React from 'react';
```
nitrogenous, line 1: i think we can use synthetic imports

import React from 'react';

components/doc/stepper/pt/wireframe.js
```diff
@@ -0,0 +1,12 @@
+import { DocSectionText } from '@/components/doc/common/docsectiontext';
```
nitrogenous, line 1: import React from 'react';

components/lib/stepper/StepperBase.js
gucal: I will do it during the theme update, thanks.

components/lib/stepper/Stepper.js
gucal: ok, thanks

components/lib/stepper/Stepper.js
gucal: I made a note to add this later. Thanks for the feedback.

Issue #6265, opened 2024-03-30
Dropdown filter : Filter Field Resets to Empty String on Empty Filter Result when uing onFilterEvent to fetch options dynamically


mohamadbdeir:
### Describe the bug

When utilizing the Dropdown component with the onFilterEvent callback to fetch options dynamically from a backend API, an issue occurs where the filter field automatically resets to an empty string if the filtering operation results in an empty array. 

### PrimeReact version

10.6.2

### React version

18.x

### Language

ALL

### Build / Runtime

Create React App (CRA)

### Steps to reproduce the behavior

1. Implement a call to remote API to fetch Dropdown options by a field value
2. Initialize a state variable that will hold the filter value
3. Use onFilter event in Dropdown to set the filter variable to event.filter
4. use useEffect to react to filter value change to fetch options
5. type a filter value that returns no result and filter value will be reset to empty string

### Expected behavior

The filter value should not be reset

dingzhanjun:
I am on 10.7.0 and the reset still happens

melloware:
@dingzhanjun can you create a StackBlitz reproducer and open a new ticket if you can reproduce it?

dingzhanjun:
thanks for your response! it is really weird, I created a StackBlitz and it works, but it just does not work in my project, what could be the issue on earth?

melloware:
hmmm its possible RHF is messing it up?  I wouldn't think so though unless the onChange or onBlur of RHF is messing it up?

PR #6267, opened 2024-03-30
Fixed dropdown filter reset when options array is empty

mohamadbdeir:
### Defect Fixes
Issue #6265 

Fixed Dropdown filter being reset to empty string when options array is empty after fetching data dynamically


nitrogenous:
Is this error still occurring? @mohamadbdeir

nitrogenous:
It seems like @melloware already merged a solution for this

melloware:
Yep I will close this.  @mohamadbdeir let us know if it is still broken after the next release

Issue #6274, opened 2024-04-01
InputText: It doesn't take classNames when unstyled is true


mevlutcantuna:
### Describe the bug

InputText doesn't take given classNames when unstyled prop is true

### Reproducer

https://stackblitz.com/edit/kuem11?file=src%2FApp.jsx,package.json

### PrimeReact version

10.6.2

### React version

18.x

### Language

TypeScript

### Build / Runtime

Create React App (CRA)

### Browser(s)

Chrome

### Steps to reproduce the behavior

1. Pass unstyled prop to InputText
2. Give it a className
3. Check it has or not className

### Expected behavior

It must take given classNames, but remove initial classNames that come from prime

melloware:
i thought in unstyled mode you were supposed to use PT.

See: https://stackblitz.com/edit/kuem11-ajcwbr?file=src%2FApp.jsx

mevlutcantuna:
@melloware okay.   I am using unstyled mode to remove initial class names like other component

melloware:
i will leave this open as I am not sure but at least you have a workaround!

melloware:
Not sure i will let Primetek chime in. I assigned this to them.

peconomou929:
This is also a problem I experienced. 

Imagine you want to build a reusable component like a button, call it `MyButton` (see the code snippet below). To do this you use an `unstyled` PrimeReact `Button`, and give it your custom pass through options. In order to abstract away these modifications, you wrap it in a component which "forwards" its `props` to PrimeReact `Button`. 

```
import { Button, ButtonProps } from 'primereact/button'

export function MyButton(props: ButtonProps) {
    return <Button {...props} unstyled pt={myButtonPTOptions} />
}
```

Now, say I want to create a `MyButton` with a left margin. If am using Tailwind, I might try something like `<MyButton className="ml-2"/>`, but, for the reasons mentioned in this issue, this will not work, as the class name gets thrown away in `unstyled` mode.

There are of course work-arounds, such as
- using the `style` property instead of the `className` property
- adding custom login inside the pass-through options object which "reads" the `className` attribute, and appends those class names to the root element class name list. 

But...it would be nice if PrimeReact just took care of this properly. The expected behaviour would be that the `unstyled` tag simply remove PrimeReact class names, such as `p-component`, `p-button`, etc. and doesn't remove any other class names.

melloware:
Agreed this is unexpected behavior.

peconomou929:
[...]

Issue #6160, opened 2024-03-15
TextInput: className property is written two times into the class attribute in DOM


cfpfeiff:
### Describe the bug

```
<InputText className="class1 class2" />
```
leads to 
```
<input class="class1 class2 p-inputtext p-component class1 class2" data-pc-name="inputtext" data-pc-section="root">
```
containing two times the string `class1 class2`

### Reproducer

https://stackblitz.com/edit/z8ngam?file=src%2FApp.jsx

### PrimeReact version

10.5.1

### React version

18.x

### Language

ALL

### Build / Runtime

Vite

### Steps to reproduce the behavior

1. Add value to `className` property of InputText.

### Expected behavior

The `class` attribute of the `input` tag should be containing the value of the `className` property only once.

melloware:
Thanks for reporting. PR Submitted.

PR #6323, opened 2024-04-08
#6274 - InputText: It doesnt take classNames when unstyled is true

gucal:
#6274 - InputText: It doesnt take classNames when unstyled is true


melloware (changes requested):
See my comment

gucal:
> @gucal i think we need to remove prop.className from InputBase.js or else this will happen again: #6160

Good catch. Thank you!

melloware (approved):

Issue #6336, opened 2024-04-08
InputText: Typescript useRef types with new const instead of class components


dvpritzbuer:
### Describe the bug

`const ref = useRef<Stepper>(null);` --> "'Stepper' refers to a value, but is being used as a type here. Did you mean 'typeof Stepper'?"

`const ref = useRef<Panel>(null);` --> works fine

### Reproducer

https://stackblitz.com/edit/vitejs-vite-osrp8p?file=src%2FApp.tsx,package.json

### PrimeReact version

10.6.2

### React version

18.x

### Language

TypeScript

### Build / Runtime

Vite

melloware:
@dvpritzbuer can you wait until the next release I am pretty sure all of your issues you are reporting are already fixed.

dvpritzbuer:
I mean my first issue today was not already fixed and why shouldn't you take my free feedback even if you just tag it with fixed in 10.6.3? 

And this issue here is valid for all new components that use const instead of class (also tested it for IconField) and not only the Stepper!

But ok.. at least in short notice my other bugs as comment:
- IconField export/import wrong ("primereact/iconfield/iconfield")
- Stepper documentation example and stackblitz use wrong imports (@/components/lib... instead of primereact/) 
- InputIcon documentation example uses vue code instead of react

melloware:
IconField already fixed: https://github.com/primefaces/primereact/pull/6271

What I was saying was in `main` many things have already been fixed related to Stepper, InputIcon, IconField.

dvpritzbuer:
Yeah that's fine, but it's a lot easier for me to just post an issue report instead of searching the code for possible fixes.
Maybe a rough estimate when 10.6.3 will release?

melloware:
i think this week sometime.  I don't work for PrimeTek...

gucal:
Hi,

With release 10.6.3 the issue should be fixed. If you notice that the issue persists, please ping me.

dvpritzbuer:
This error is not fixed - updated the linked stackblitz.

Also I recognized the original ticket description is not correct since Stepper doesn't have a ref and I just tested it initially with this component since this is a new const type definition. 
The stackblitz is now based on InputText.

I have no perfect solution here for you since it would either require the user to write more code or to import another type.

But it would work with this line:
`const inputRef2 = useRef<React.ElementRef<typeof InputText>>(null);`

[...]

PR #6356, opened 2024-04-09
Fix #6334: typescript definition for Stepper and StepperPanel

### LINKED ISSUE #6334: StepperPanel: New const instead of class components are no valid JSX.Element's
### Describe the bug

"JSX element type 'Stepper' does not have any construct or call signatures."

"ERROR(TypeScript)  'StepperPanel' cannot be used as a JSX component.
  Its type 'Component<StepperPanelProps, any, any>' is not a valid JSX element type."

### Reproducer

https://stackblitz.com/edit/vitejs-vite-9mym3z?file=src%2FApp.tsx,package.json

### PrimeReact version

10.6.2

### React version

18.x

### Language

TypeScript

### Build / Runtime

Vite

### Browser(s)

_No response_

_No response_

### Expected behavior

_No response_

dvpritzbuer:
Fix #6334: typescript definition for Stepper and StepperPanel

### DIFF
--- a/components/lib/stepper/stepper.d.ts
+++ b/components/lib/stepper/stepper.d.ts
@@ -87,8 +87,8 @@ export interface StepperChangeEvent {
  */
 export interface StepperRefAttributes {
     getElement: () => HTMLDivElement;
-    nextCallback: (e: React.SyntheticEvent) => void;
-    prevCallback: (e: React.SyntheticEvent) => void;
+    nextCallback: (e?: React.SyntheticEvent) => void;
+    prevCallback: (e?: React.SyntheticEvent) => void;
 }
 
 /**
--- a/components/lib/stepper/stepper.d.ts
+++ b/components/lib/stepper/stepper.d.ts
@@ -134,4 +134,4 @@ export interface StepperProps {
  *
  * @group Component
  */
-export declare const Stepper: React.Component<StepperProps, any>;
+export declare const Stepper: React.FC<React.PropsWithChildren<StepperProps>>;
--- a/components/lib/stepper/stepper.d.ts
+++ b/components/lib/stepper/stepper.d.ts
@@ -83,6 +83,15 @@ export interface StepperChangeEvent {
 }
 
 /**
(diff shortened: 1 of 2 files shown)


melloware (approved):

melloware:
@gucal @nitrogenous @mertsincan did we not use `React.PropsWithChildren` because its React 18 only and we wanted to be backward compatible with React 17?  https://stackoverflow.com/questions/71788254/react-18-typescript-children-fc

Is that why we declare children in most components like this:

```ts
    /**
     * Used to get the child elements of the component.
     * @readonly
     */
    children?: React.ReactNode | undefined;
```

??

dvpritzbuer:
@melloware 
seems to work fine with react 17.0.2 - here is a stackblitz example: https://stackblitz.com/edit/vitejs-vite-9iwtma?file=src%2FApp.tsx

melloware:
nice!

dvpritzbuer:
I've recognized the Stepper does have a ref property and was missing the definition for that.

Updated the branch with a proposed solution. I'm not sure about naming and the documentation part, but at least you have a general example of how the typing could work with the ref and custom methods.

melloware (approved):

sancelot:
Hi, 

I have had this error, I think it is related, but I am unable to tell it is "normal" or a "mistake" somewhere in typings ?

error TS2749: 'Stepper' refers to a value, but is being used as a type here. Did you mean 'typeof Stepper'?
const stepperRef = useRef<Stepper|null>(null);

melloware:
No @sancelot the typing of this compoennt looks wrong and weird to me. I can reproduce it right here @dvpritzbuer see: https://stackblitz.com/edit/gwed3brm?file=src%2FApp.tsx

dvpritzbuer:
@melloware yeah this bug was already known as this PR here was created, I created this issue for that: https://github.com/primefaces/primereact/issues/6336

melloware:
Thank you. I see the issue these are declared differently than all other components I am not quite sure why.

PR #6360, opened 2024-04-09
Fix #6357: typescript definition for ButtonGroup

dvpritzbuer:
Fix #6357: typescript definition for ButtonGroup (missing ref definition)

PR #6388, opened 2024-04-12
Add missing children prop in FloatLabel.d.ts

### LINKED ISSUE #6387: FlotaLabel: Type definitions are missing a children prop
### Describe the bug

Even though the component clearly accepts children, the props in the accompanying .d.ts file don't contain an entry `children`.

### Reproducer

_No response_

### PrimeReact version

10.6.3

### React version

17.x

### Language

TypeScript

### Build / Runtime

Vite

### Browser(s)

_No response_

The example from the docs:

```tsx
<FloatLabel>
    <InputText id="username" value={value} onChange={(e) => setValue(e.target.value)} />
    <label for="username">Username</label>
</FloatLabel>
```

does not compile. The error is

```
error TS2559: Type '{ children: Element; }' has no properties in common with type 'IntrinsicAttributes & FloatLabelProps & RefAttributes<HTMLSpanElement>'
```

### Expected behavior

The code should compile, as it is the intended usage.

NiclasvanEyk:
### Defect Fixes

Fixes #6387

I used the same description format as in the other components. I validated the change works by adding it to the local .d.ts file from my node_modules and it makes the example mentioned in the issue and my code compile again.

### DIFF
--- a/components/lib/floatlabel/FloatLabel.d.ts
+++ b/components/lib/floatlabel/FloatLabel.d.ts
@@ -50,6 +50,11 @@ export interface FloatLabelPassThroughOptions {
  */
 export interface FloatLabelProps {
     /**
+     * Used to get the child elements of the component.
+     * @readonly
+     */
+    children?: React.ReactNode | undefined;
+    /**
      * Used to pass attributes to DOM elements inside the component.
      * @type {FloatLabelPassThroughOptions}
      */


melloware:
@NiclasvanEyk can you check out this PR: https://github.com/primefaces/primereact/pull/6360  @dvpritzbuer thinks this is a better way to define Children?

NiclasvanEyk:
@melloware does not really matter. If you look [at the react type definitions](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/0cebe5d5bcaad4a8f18e525b1cd098583b252585/types/react/index.d.ts#L1586), it boils down to the same thing ([also in React 17](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react/v17/index.d.ts#L824)).

I personally prefer the explicit way, but either way works.

melloware:
OK leave it the explicit way!

PR #6407, opened 2024-04-17
Fix#6404: invalid boolean fix

### LINKED ISSUE #6404: Password: Invalid PT props
### Describe the bug

In the Features Tab you can see the invalid prop can call directly but not, when I read the API Tab you can call the "invalid" inside pt prop. That's why I got an error everytime I call "invalid" directly.

IMG_20240417_113144

IMG_20240417_113452

### Reproducer

_No response_

### PrimeReact version

10.6.3

### React version

18.x

### Language

TypeScript

### Build / Runtime

Vite

### Browser(s)

_No response_

_No response_

### Expected behavior

_No response_

Rekl0w:
Fix #6404: invalid boolean fix

### DIFF
--- a/components/doc/password/invaliddoc.js
+++ b/components/doc/password/invaliddoc.js
@@ -5,7 +5,7 @@ import { Password } from '@/components/lib/password/Password';
 export function InvalidDoc(props) {
     const code = {
         basic: `
-<Password invalid />
+<Password invalid="true" />
         `,
         javascript: `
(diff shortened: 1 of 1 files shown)


melloware:
I submitted the correct fix: https://github.com/primefaces/primereact/pull/6409

Rekl0w:
#6404 according to him invalid prop is not working with direct call so i added this

PR #6425, opened 2024-04-19
Fix #6422: Calendar updateViewDate fix

### LINKED ISSUE #6422: Calendar: updateViewDate method brings the error with null value
### Describe the bug

I have this example code, where I need to reset calendar's date manually via its API (https://primereact.org/calendar/#api.Calendar.methods).
image

In this example I'm trying to call method updateViewDate, passing null inside it.

But in fact, I'm getting the error while pressing the button:

image

If I try to check the source code from the link in call stack, I will see that the method updateViewDate calls setCurrentMonth(value.getMonth()) without checking for truthy of value (the same thing for year property).

image

Note: the API of this method is documented, that it can accept Nullable (value: Nullable<Date | Date[]>)

My suggestion is to check value for truthy before trying to set month/year in this method.

### Reproducer

https://codesandbox.io/p/devbox/objective-lena-ypxjyy?file=%2Fsrc%2FApp.tsx%3A22%2C16

### PrimeReact version

10.6.3

### React version

18.x

### Language

TypeScript

### Build / Runtime

Vite

### Browser(s)

_No response_

1. Try to call method `updateViewDate` and pass `null` to reset date.

### Expected behavior

_No response_

Rekl0w:
Fix #6422: Calendar updateViewDate fix

It needs null check for value

### DIFF
--- a/components/lib/calendar/Calendar.js
+++ b/components/lib/calendar/Calendar.js
@@ -1055,7 +1055,7 @@ export const Calendar = React.memo(
                 setViewDateState(value);
             }
 
-            if (value !== null || true) {
+            if (value !== null) {
                 setCurrentMonth(value.getMonth());
                 setCurrentYear(value.getFullYear());
             } else {
--- a/components/lib/calendar/Calendar.js
+++ b/components/lib/calendar/Calendar.js
@@ -1055,8 +1055,13 @@ export const Calendar = React.memo(
                 setViewDateState(value);
             }
 
-            setCurrentMonth(value.getMonth());
-            setCurrentYear(value.getFullYear());
+            if (value !== null || true) {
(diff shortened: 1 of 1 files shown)


components/lib/calendar/Calendar.js
```diff
@@ -1055,8 +1055,13 @@ export const Calendar = React.memo(
                 setViewDateState(value);
             }
 
-            setCurrentMonth(value.getMonth());
-            setCurrentYear(value.getFullYear());
+            if (value !== null || true) {
```
melloware, line 1058:
```suggestion
            if (value !== null) {
```

melloware (changes requested):

melloware (approved):

PR #6430, opened 2024-04-20
Fix #6124 : Dropdown navigation like PrimeVue

### LINKED ISSUE #6124: Dropdown: keyboard navigation doesn't work on dropdown with groups
### Describe the bug

Keyboard navigation doesn't work on Dropdown with groups because `findNextOptionIndex` and `findPrevOptionIndex` uses top level of `visibleOptions` array to find next / prev position but in case of groups that array consists of group items and actual items are inside `items` property.

### Reproducer

https://primereact.org/dropdown/#group

### PrimeReact version

10.5.1

### React version

17.x

### Language

TypeScript

### Build / Runtime

Create React App (CRA)

### Browser(s)

_No response_

1. Open Dropdown with groups
2. Try to keyboard navigate with arrows

### Expected behavior

`findNextOptionIndex` and `findPrevOptionIndex` should look into `items` property for actual items.

### LINKED ISSUE #6417: Dropdown: All items at the same index in a group are marked focused
### Describe the bug

When using [grouped options](https://primereact.org/dropdown/#group), when hovering over an option, all options at the same index index in the group have the `p-focus` class and the `data-p-focused="true"` attribute set. See screenshot from the PrimeReact site:
Screenshot 2024-04-18 at 10 06 07 PM

### Reproducer

_No response_

### PrimeReact version

10.6.3

### React version

18.x

### Language

TypeScript

### Build / Runtime

Vite

### Browser(s)

_No response_

_No response_

### Expected behavior

_No response_

KirilCycle:
Fix #6124
Fix #6417 
Each Group item now have unique index

### DIFF
--- a/components/lib/dropdown/Dropdown.js
+++ b/components/lib/dropdown/Dropdown.js
@@ -58,6 +58,7 @@ export const Dropdown = React.memo(
         });
 
         const flattenedOptions = props.optionGroupLabel ? props.options.flatMap((item) => item.items) : null
+        const appropriateOptions = () => flattenedOptions ?? visibleOptions
 
         const getVisibleOptions = () => {
             if (hasFilter && !isLazy) {
@@ -354,8 +355,6 @@ export const Dropdown = React.memo(
 
             return selectedIndex < 0 ? findLastOptionIndex() : selectedIndex;
         };
-
-        const appropriateOptions = () => flattenedOptions ?? visibleOptions
 
         const findFirstOptionIndex = () => {
             return appropriateOptions().findIndex((option) => isValidOption(option));
--- a/components/lib/dropdown/Dropdown.js
+++ b/components/lib/dropdown/Dropdown.js
@@ -57,7 +57,7 @@ export const Dropdown = React.memo(
             when: overlayVisibleState
         });
 
-         function flatOptions(options) {
+         const flatOptions = (options) => {
             return (options || []).reduce((result, option, index) => {
                 result.push({ optionGroup: option, group: true, index, code: option.code, label: option.label });
 
--- a/components/lib/dropdown/Dropdown.js
+++ b/components/lib/dropdown/Dropdown.js
@@ -80,7 +80,7 @@ export const Dropdown = React.memo(
                 if (props.optionGroupLabel) {
                     let filteredGroups = [];
 
-                    for (let optgroup of options) {
+                    for (let optgroup of props.options) {
                         let filteredSubOptions = FilterService.filter(getOptionGroupChildren(optgroup), searchFields, filterValue, props.filterMatchMode, props.filterLocale);
 
                         if (filteredSubOptions && filteredSubOptions.length) {
@@ -88,7 +88,7 @@ export const Dropdown = React.memo(
                         }
                     }
 
-                    return filteredGroups;
+                    return flatOptions(filteredGroups);
                 }
 
                 return FilterService.filter(options, searchFields, filterValue, props.filterMatchMode, props.filterLocale);
--- a/components/lib/dropdown/Dropdown.js
+++ b/components/lib/dropdown/Dropdown.js
@@ -57,6 +57,8 @@ export const Dropdown = React.memo(
             when: overlayVisibleState
         });
 
+        const flatedGroupOptions = props.optionGroupLabel ? props.options.flatMap((item) => item.items) : null
+
         const getVisibleOptions = () => {
             if (hasFilter && !isLazy) {
                 const filterValue = filterState.trim().toLocaleLowerCase(props.filterLocale);
@@ -298,7 +300,8 @@ export const Dropdown = React.memo(
         };
 
         const findSelectedOptionIndex = () => {
-            return hasSelectedOption ? visibleOptions.findIndex((option) => isValidSelectedOption(option)) : -1;
+            const appropriateOptions = flatedGroupOptions ?? visibleOptions
+            return hasSelectedOption ? appropriateOptions.findIndex((option) => isValidSelectedOption(option)) : -1;
         };
 
         const findFirstFocusedOptionIndex = () => {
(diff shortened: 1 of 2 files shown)


components/lib/dropdown/Dropdown.js
```diff
@@ -352,22 +355,26 @@ export const Dropdown = React.memo(
             return selectedIndex < 0 ? findLastOptionIndex() : selectedIndex;
         };
 
+
+        const appropriateOptions = () => flatedGroupOptions ?? visibleOptions
```
KirilCycle, line 359: maybe its better to use a simple variable

KirilCycle:
@melloware Hello! Its done

components/lib/dropdown/Dropdown.js
melloware: Yes move this up to where you get `flattenedOptions`  and shouldn't `visibleOptions = flattenedOptions` and then all this code will just work using `visibleOptions`?

components/lib/dropdown/Dropdown.js
```diff
@@ -57,6 +57,8 @@ export const Dropdown = React.memo(
             when: overlayVisibleState
         });
 
+        const flatedGroupOptions = props.optionGroupLabel ? props.options.flatMap((item) => item.items) : null
```
melloware, line 60: This should be `flattenedOptions`

components/lib/dropdown/DropdownPanel.js
```diff
@@ -76,7 +76,7 @@ export const DropdownPanel = React.memo(
             }
         };
 
-        const createGroupChildren = (optionGroup, style) => {
+        const createGroupChildren = (optionGroup, style, passedOptionsCount) => {
```
melloware, line 79:
```suggestion
        const createGroupChildren = (optionGroup, style, index) => {
```

Call `passedOptionsCount` just `index`

melloware (changes requested):

components/lib/dropdown/Dropdown.js
KirilCycle: visibleOptions need to be the same as it was, this format will be used at DropdownPanel component to make groups

melloware:
@KirilCycle i think this may also fix this issue correct?  https://github.com/primefaces/primereact/issues/6417

KirilCycle:
> @KirilCycle i think this may also fix this issue correct? #6417

Exactly!

components/lib/dropdown/Dropdown.js
KirilCycle: I cant  access visibleOptions  before initialization, should i leave it as method or move up their initialization ?

melloware:
Let me take a deeper look at this. I like where you are going though.

KirilCycle:
> Let me take a deeper look at this. I like where you are going though.

I also second approach, we can use simple number for key at simple list, but for group list we can use key like `${groupIndex}_${index}`

melloware:
I think we need to study PrimeVue they actually do this: https://github.com/primefaces/primevue/blob/344ec64cd461aad1493073dd3ff10bc8a7338654/components/lib/dropdown/Dropdown.vue#L879-L888

And the Groups are actually a different type of Option that has `result.push({ optionGroup: option, group: true, index });` so they index everything properly in getVisibleOptions

https://github.com/primefaces/primevue/blob/344ec64cd461aad1493073dd3ff10bc8a7338654/components/lib/dropdown/Dropdown.vue#L902-L926

KirilCycle:
> I think we need to study PrimeVue they actually do this: https://github.com/primefaces/primevue/blob/344ec64cd461aad1493073dd3ff10bc8a7338654/components/lib/dropdown/Dropdown.vue#L879-L888
> 
> And the Groups are actually a different type of Option that has `result.push({ optionGroup: option, group: true, index });` so they index everything properly in getVisibleOptions
> 
> https://github.com/primefaces/primevue/blob/344ec64cd461aad1493073dd3ff10bc8a7338654/components/lib/dropdown/Dropdown.vue#L902-L926

Should I try to use it the way it's done in vue?

melloware:
Yes we should see if we can modify things the Vue way.   If you try the Vue showcase the Dropdown with Groups is working fine: https://primevue.org/dropdown/#group

KirilCycle:
> Yes we should see if we can modify things the Vue way. If you try the Vue showcase the Dropdown with Groups is working fine: https://primevue.org/dropdown/#group

Okey ill take a look

KirilCycle:
@melloware Vue Dropdown will flat all this to single options list first, and then all logic will work with this flatted list,And React Dropdown component use same format as it was passed and functions are not able to work with this type of data, and DropdownPanel is designed to handle groups the way they are, however still has an issue with indexses. Should i keep current solution to this problem or apply Vue behaviour ?

melloware:
Yes I saw that. I feel like Vue is doing it a cleaner way and we just need to tweak the React code to do the same thing. I think DropDownPanel will need some changes but I think in the long run this will be better.

components/lib/dropdown/Dropdown.js
```diff
@@ -57,15 +57,30 @@ export const Dropdown = React.memo(
...
+
             if (hasFilter && !isLazy) {
                 const filterValue = filterState.trim().toLocaleLowerCase(props.filterLocale);
                 const searchFields = props.filterBy ? props.filterBy.split(',') : [props.optionLabel || 'label'];
 
                 if (props.optionGroupLabel) {
                     let filteredGroups = [];
 
-                    for (let optgroup of props.options) {
+                    for (let optgroup of options) {
```
KirilCycle, line 83: I did not check if filter functionality works fine yet

components/lib/dropdown/DropdownPanel.js
```diff
@@ -144,17 +113,14 @@ export const DropdownPanel = React.memo(
...
-                        {groupChildrenContent}
-                    </React.Fragment>
-                );
+                    </li>
+                )
+                
             }
-
-            const optionLabel = props.getOptionLabel(option);
             const optionKey = index + '_' + props.getOptionRenderKey(option);
```
KirilCycle, line 122: Group Item key not the same as it was, i do not use here group index to create render key, is this okey ?

components/lib/dropdown/Dropdown.js
```diff
@@ -57,6 +57,9 @@ export const Dropdown = React.memo(
             when: overlayVisibleState
         });
 
+        const flattenedOptions = props.optionGroupLabel ? props.options.flatMap((item) => item.items) : null
```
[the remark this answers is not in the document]
KirilCycle: Sorry, what you mean ? Filtering works fine with last commit

components/lib/dropdown/DropdownPanel.js
KirilCycle: Same thing, but the index is at the back

melloware:
Wow this is looking so much cleaner!

melloware:
Oh run `npm run format` on the code.

melloware (approved):

nitrogenous (approved):
LGTM. Amazing job thx!

PR #6435, opened 2024-04-21
Fix #6095: Dialog refactor using FocusTrap

### LINKED ISSUE #6095: Dialog: (accessibility) "display: none" inside dialog breaks focus tabbing
### Describe the bug

When a Dialog contains a section with "display: none" and a <input/> as child of section, user cannot reach each focusable input/buttons/... using TAB key.

### Reproducer

https://stackblitz.com/edit/qejj7w?file=src%2FApp.jsx

### PrimeReact version

10.5.1

### React version

17.x

### Language

TypeScript

### Build / Runtime

Create React App (CRA)

### Browser(s)

_No response_

1. open stackBlitz app  (spawn from Dialog basic showcase)
2. use tab to move set focus on all input & buttons:  (this works)
3. open dialog
4. use tab to move focus around dialog's buttons and inputs:  Focus is stopped on input due to "display: none" section.

50 cents: Issue comes from DomHandler.getFocusableElements() ?

Note that issue is not the same as  https://github.com/primefaces/primereact/issues/4035

### Expected behavior

All focusable elements inside a dialog box should be reached using Tab (Shift+Tab) key.

melloware:
Fix #6095: Dialog refactor using FocusTrap

### DIFF
--- a/components/lib/dialog/Dialog.js
+++ b/components/lib/dialog/Dialog.js
@@ -2,6 +2,7 @@ import * as React from 'react';
 import PrimeReact, { PrimeReactContext, localeOption } from '../api/Api';
 import { useHandleStyle } from '../componentbase/ComponentBase';
 import { CSSTransition } from '../csstransition/CSSTransition';
+import FocusTrap from '../focustrap/FocusTrap';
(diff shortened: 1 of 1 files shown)


nitrogenous (approved):
LGTM

PR #6488, opened 2024-04-27
Fix #6489: add menu context to menu

### LINKED ISSUE #6489: Feature Request: include "context" in the PrimeReact Menu component
### Discussed in https://github.com/orgs/primefaces/discussions/1728

<div type='discussions-op-text'>

<sup>Originally posted by **peconomou929** April 21, 2024</sup>
All PrimeReact components have the so-called "Pass Through Method Options" which include a subset of the following three objects: `props`, `state`, and `context`. 

Let's take a look at the `TabMenu` for example. It has all three objects. The `context`, which is of type `TabMenuContext`, has a property `item` and a property `index`. This allows you to query each tab item individually in the pass through options, and as a result, give each tab its own styles. For example, we can make the selected tab look different from the rest. 

This is not the case with the PrimeReact Menu. There is no `context` in the Pass Through Method Options. Hence, there is no way to apply different styles to the different menu items (without creating your own custom JavaScript solution which targets certain elements at run time). This would be useful if I wanted to show a particular menu item as being "selected" (much like a selected tab in a TabMenu). 

Could we therefore add a `context` to the PrimeReact Menu component?</div>

peconomou929:
As discussed [here](https://github.com/orgs/primefaces/discussions/1728), adds a MenuContext to the Menu so that individual menu items can be customized via the Pass Through options, mirroring the behavior of the Tab Menu. 

Can the maintainers take a look and open an issue for this feature?'

Fix #6489

### DIFF
--- a/components/doc/common/apidoc/index.json
+++ b/components/doc/common/apidoc/index.json
@@ -34421,7 +34421,7 @@
                             "name": "item",
                             "optional": false,
                             "readonly": false,
-                            "type": "any",
+                            "type": "MenuItem",
                             "description": "Current menuitem"
                         },
                         {
--- a/components/doc/common/apidoc/index.json
+++ b/components/doc/common/apidoc/index.json
@@ -34403,6 +34403,40 @@
                             "optional": false,
                             "readonly": false,
                             "type": "MenuState"
+                        },
+                        {
+                            "name": "context",
+                            "optional": false,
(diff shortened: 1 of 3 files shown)


melloware:
I will take a look

melloware (approved):

nitrogenous (approved):
LGTM. Great PR

PR #6519, opened 2024-04-30
Fix #6518: fix optionLabel

### LINKED ISSUE #6518: Dropdown: Missing `optionLabel` property causes error
### Describe the bug

When creating a Dropdown comp without the `optionLabel` property, opening the dropdown causes the page to error. When in the past it would properly default to `"label"`.

This is especially a problem with `RowsPerPageDropdown` in `Paginator` where you cannot simply add that prop.

### Reproducer

https://stackblitz.com/edit/vitejs-vite-sswemw?file=src%2FApp.tsx

### PrimeReact version

10.6.4

### React version

18.x

### Language

TypeScript

### Build / Runtime

Vite

### Browser(s)

_No response_

1. Create a `Dropdown` 
2. Don't set the `optionLabel` prop
3. Click dropdown

### Expected behavior

Not setting the `optionLabel` property should default to using `"label"`.

Rekl0w:
Fix #6518: fix `optionLabel`

### DIFF
--- a/components/lib/dropdown/Dropdown.js
+++ b/components/lib/dropdown/Dropdown.js
@@ -887,7 +887,7 @@ export const Dropdown = React.memo(
         };
 
         const getOptionValue = (option) => {
-            return props.optionValue ? ObjectUtils.resolveFieldData(option, props.optionValue) : option;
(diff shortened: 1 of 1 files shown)


Rekl0w:
I added "label" for `optionLabel` and "value" for `optionValue`

PR #6560, opened 2024-05-03
Fix: 6555 Missing PT Option for IconField in Password 

### LINKED ISSUE #6555: Missing PT Option for IconField in Password
### Describe the bug

Relates to #6349 - fixed by @melloware in #6493.
The point here is that I'm using `<Password>` within `display: flex`, and I'm no longer able to let the input grow. The reason is that there is no PT option within `<Password>` to give any `className` down to `IconField`.

[The relevant code can be found in password](https://github.com/primefaces/primereact/blob/7312fe466e189581c53c3fe43bf3381db4b59455/components/lib/password/Password.js#L483)

### Reproducer

https://primereact.org/password/#togglemask

### PrimeReact version

10.6.4

### React version

18.x

### Language

ALL

### Build / Runtime

Vite

### Browser(s)

_No response_

_No response_

### Expected behavior

_No response_

KirilCycle:
### Defect Fixes
Fix #6555 
IconField can be modified from Password component

### DIFF
--- a/components/lib/password/Password.js
+++ b/components/lib/password/Password.js
@@ -481,13 +481,16 @@ export const Password = React.memo(
         let input = <InputText {...inputTextProps} />;
 
         if (icon) {
-
-            const iconFieldProps = mergeProps(
+            
+            const iconField = props.iconField
+            const iconFieldProps = iconField ? mergeProps(
                 {
-                    className: classNames(props.iconFieldClassName)
+                    className: classNames(iconField.className)
                 },
-                ptm('iconFieldClassName')
-            );
+                ptm('iconField')
+            ) : {} 
+
+            console.log(iconFieldProps)
 
             input = (
                 <IconField {...iconFieldProps} >
--- a/components/lib/password/Password.js
+++ b/components/lib/password/Password.js
@@ -481,17 +481,8 @@ export const Password = React.memo(
         let input = <InputText {...inputTextProps} />;
 
         if (icon) {
-            const iconFieldProps = props.iconField
-                ? mergeProps(
-                      {
-                          className: classNames(props.iconField.className)
-                      },
-                      ptm('iconField')
-                  )
-                : {};
-
             input = (
-                <IconField {...iconFieldProps}>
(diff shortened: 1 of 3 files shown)


components/lib/password/PasswordBase.js
```diff
@@ -85,6 +85,7 @@ export const PasswordBase = ComponentBase.extend({
         header: null,
         content: null,
         footer: null,
+        iconField: null,
```
melloware, line 88: I don't think you need this we aren't adding a new property just a new passthrough.

components/lib/password/Password.js
```diff
@@ -481,8 +481,18 @@ export const Password = React.memo(
         let input = <InputText {...inputTextProps} />;
 
         if (icon) {
+            const iconField = props.iconField;
```
melloware, line 484: I don't think you need the `iconField` part?

melloware (changes requested):

components/lib/password/PasswordBase.js
KirilCycle: Without this it wont work 'React does not recognize the `iconField` prop on a DOM'

components/lib/password/Password.js
KirilCycle: Well, I thought that in the future it would be easier to access the needed field value

components/lib/password/PasswordBase.js
melloware: I just updated your PR can you do a git pull and test again with my changes?

components/lib/password/Password.js
melloware: You can already change the icons with `props.hideIcon` etc.

```js
            if (unmaskedState) {
                icon = props.hideIcon || <EyeSlashIcon {...hideIconProps} />;
            } else {
                icon = props.showIcon || <EyeIcon {...showIconProps} />;
            }
```

components/lib/password/PasswordBase.js
KirilCycle: Okey

melloware:
I think these latest changes are good @nitrogenous !

melloware (approved):

PR #6566, opened 2024-05-04
Fix #6522: Minus sign after all selection - InputNumber

### LINKED ISSUE #6522: InputNumber: Selected negative value can't be overwritten with a minus ('-') symbol
### Describe the bug

When there is already a '-' entered into an InputNumber component it correctly does not allow adding another '-'.
However this currently also applies when you want to replace a selected value.
It is rather easy to go around this issue with pressing  the 'backspace' key first to clear the value but it adds some inconvenience for the user when replacing negative values.

### Reproducer

https://primereact.org/inputnumber/

### PrimeReact version

10.6.4

### React version

18.x

### Language

TypeScript

### Build / Runtime

Create React App (CRA)

### Browser(s)

_No response_

1. Go to https://primereact.org/inputnumber/
2. Enter "-100" in the first rendered InputNumber (with text "Integer only")
3. Select the "-100"
4. Press the '-' key again
5. Nothing happens

image

### Expected behavior

When selecting the whole value together with the '-' and then pressing the key '-', I would expect everything to be replaced with a single '-'.

ujshaikh:
Fix #6522: 
Actual: Not allowed to enter minus sign after all number selection if it exists- InputNumber Component

### DIFF
--- a/components/lib/inputnumber/InputNumber.js
+++ b/components/lib/inputnumber/InputNumber.js
@@ -650,7 +650,7 @@ export const InputNumber = React.memo(
             }
 
             _currency.current.lastIndex = 0;
-
+            
             return { decimalCharIndex, minusCharIndex, suffixCharIndex, currencyCharIndex };
         };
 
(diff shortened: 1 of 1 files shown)


melloware:
Please run `npm run format`

melloware (approved):

nitrogenous (approved):

PR #6587, opened 2024-05-07
Fix #6083 TreeTable onSort

### LINKED ISSUE #6083: TreeTable: Sort toggle when using custom onSort function
### Describe the bug

Hi, I'm using onSort function to make an API call for sorting, but when onSort function is used, original icon highlight and toggling between ASC and DESC is lost.

Expected:

image

- Highlight icon showing up
- Click on arrows toggling the state between ASC and DESC (0 and 1)

Actual:

- No indicator of sort being enabled
- sortOrder in event is always 1

image

image

 

### Reproducer

https://stackblitz.com/edit/vqw5gi-vnizjy?file=src%2FApp.jsx

### PrimeReact version

9.5.0

### React version

17.x

### Language

TypeScript

### Build / Runtime

Create React App (CRA)

KirilCycle:
### Defect Fixes
Fix #6083

Apply sortField and sortOrder properties only during the first rendering

### DIFF
--- a/components/lib/treetable/TreeTable.js
+++ b/components/lib/treetable/TreeTable.js
@@ -161,12 +161,11 @@ export const TreeTable = React.forwardRef((inProps, ref) => {
 
         if (props.onSort) {
             props.onSort(eventMeta);
-        } else {
-            setFirstState(0);
-            setSortFieldState(eventMeta.sortField);
-            setSortOrderState(eventMeta.sortOrder);
-            setMultiSortMetaState(eventMeta.multiSortMeta);
         }
+        setFirstState(0);
+        setSortFieldState(eventMeta.sortField);
+        setSortOrderState(eventMeta.sortOrder);
+        setMultiSortMetaState(eventMeta.multiSortMeta);
 
         if (props.onValueChange) {
             props.onValueChange(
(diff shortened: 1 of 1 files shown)


melloware:
Hmm I am not sure this is correct.  Did you look at DataTable?  I feel like this was done on purpose if the user is using `props.onSort` they are controlling the sort so `props.sort` is applied else the internal `sortState` is used if PrimeReact is controlling the sort not the user.

KirilCycle:
> Hmm I am not sure this is correct. Did you look at DataTable? I feel like this was done on purpose if the user is using `props.onSort` they are controlling the sort so `props.sort` is applied else the internal `sortState` is used if PrimeReact is controlling the sort not the user.

Yes i understand what you mean, let me take a look at DataTable

KirilCycle:
Yes, you are right, DataTable behaves the same way as TreeTable, you need to manually control "sortField" and "sortOrder" to make it work. This approach is more flexible, but not as clear. I saw that primevue works similarly to this PR, even with onScroll it will work, but there is no way to control these parameters externally, does it make sense to control them externally?

melloware:
Its the way its always been designed for right or wrong so I would leave it.

PR #6588, opened 2024-05-08
Fix #5147: Check meta key for MacOS

### LINKED ISSUE #5147: KeyFilter: Paste under MacOS does not work
### Describe the bug

Under MacOS (both Safari and Chrome) pasting from the clipboard via key combination is not possible when an InputText has a KeyFilter (pint in my case) configured. Pasting via the context menu works.

The issue can be reproduced on the Component Showcase: https://primereact.org/keyfilter/

### Reproducer

_No response_

### PrimeReact version

10.0.2

### React version

17.x

### Language

TypeScript

### Build / Runtime

Vite

### Browser(s)

MacOS, Safari, Chrome

1. Go to https://primereact.org/keyfilter/
2. Paste a number with key combination in the different fields

### Expected behavior

Copied content should be pasted.

axlerk:
### Defect Fixes

Fix #5147: We should check e.metaKey for MacOS to unlock copy/cut/paste/select all by pressing combinations with command (like `⌘+C`, `⌘+X`, `⌘+V`, `⌘+A` etc).

MDN: https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/metaKey

### DIFF
--- a/components/lib/keyfilter/KeyFilter.js
+++ b/components/lib/keyfilter/KeyFilter.js
@@ -34,7 +34,7 @@ export const KeyFilter = {
             return;
         }
 
-        if (e.ctrlKey || e.altKey) {
+        if (e.ctrlKey || e.altKey || e.metaKey) {
             return;
         }
 


melloware (approved):

melloware:
Thanks @axlerk !

Issue #6569, opened 2024-05-04
DataTable: Autocomplete with Dialog on cell editing is not working


didix16:
### Describe the bug

Hello everyone,

Firstly, let me explain what I'm trying to achieve: I have a DataTable with a column from which you can choose a list of objects (for example, Agents) that have the properties id and name. This list is displayed with an Autocomplete cell editor. I've added functionality to this Autocomplete so that if the current value doesn't match the name of one of the objects in the list, it displays a Dialog with an error message. This forces the Autocomplete to remain focused until a correct value is entered or the Autocomplete is empty (null). To accomplish this, I handle both onBlur and onHide events, etc. You can see the implementation of the component in the EntityAutoComplete.tsx file in the provided reproducer.

The component works correctly if it's not within a BodyCell. The issue arises when you edit a cell. Given this code snippet:

https://github.com/primefaces/primereact/blob/7312fe466e189581c53c3fe43bf3381db4b59455/components/lib/datatable/BodyCell.js#L61-L75

When you click outside the cell, the document's listener is triggered, causing the table cell to return to view mode. Additionally, the parameter ```options: true``` is instructing the document's addEventListener method to enable ```useCapture```. As a result, it executes before any EventTarget and renders attempts to stop propagation to the document ineffective. This potentially prevents the cell from remaining in edit mode. It might be interesting to define options as false and have control over choosing to cancel the event, so that it doesn't reach the document and the cell doesn't return to view mode.

Furthermore, I've noticed that in version 10.6.3, the autoFocus in the button added in the Dialog's footer worked. Now, in version 10.6.4, the focus always seems to be on the Dialog's close button and the autoFocus of the button in the footer is ignored.

I have also observed that if I add the ```appendTo="self"``` property (line 188 in the EntityAutoComplete.tsx file) to the Dialog, since the Dialog is within the cell in this case, clicking the **Accept** button or the close button correctly returns focus to autocomplete and displays the suggested list. However, other cases still don't work, such as pressing Escape or pressing Enter focusing on the **Accept** button or the close button.

One last observation: if I don't add the ```appendTo="self"``` property to the Dialog, pressing Enter with focus on the **Accept** button or the close button triggers the onCellEditComplete callback twice in succession, once with the KeyDown event and the other with the PointerEvent:

![image](https://github.com/primefaces/primereact/assets/10291092/ca6e841d-847d-4f3d-bfef-08c3bf521f71)

If I add the property, only the KeyDown event is triggered.
[...]

PR #6611, opened 2024-05-11
Added new capabilities and properties to Column and Dialog

didix16:
This fixes some problmes with autocomplete component with dialog
Issue #6569


components/lib/dialog/Dialog.js
```diff
@@ -449,7 +449,8 @@ export const Dialog = React.forwardRef((inProps, ref) => {
                     type: 'button',
                     className: cx('closeButton'),
                     'aria-label': ariaLabel,
-                    onClick: onClose
+                    onClick: onClose,
+                    onKeyDown: (ev) => props.stopPropagationOnKeyDown && ev.stopPropagation()
```
melloware, line 453: this doesn't need to be a property right? Shouldn't stopPropagation just be the default?

melloware (changes requested):

components/lib/column/column.d.ts
```diff
@@ -963,6 +963,12 @@ export interface ColumnProps {
      * @defaultValue false
      */
     frozen?: boolean | undefined;
+    /**
+     * When enabled and cellEditorValidator is set, force to call cellEditorValidator
+     * before cell editor is closed. If cellEditorValidator returns false, editor stays open.
+     * @defaultValue false
+     */
+    forceHasValidValue?: boolean | undefined;
```
melloware, line 971: Is this property really needed as the check for `cellEditValidator` already exists so if you have a cellEditValidator defined it should check right?  Or is this a case where you have a `cellEditValidator` defined but DONT want to check it?

melloware: I guess I don't understand why you would declare a `cellEditValidator` and then want to skip using it? Just don't delcare a `cellEditValidator` right?

components/lib/dialog/Dialog.js
didix16: Yeah it could be the default. I just added a property because I thought that maybe someone needs that the event propagates for some reason, this way the code adapts for different needs but maybe we can try to stop the propagation by default.

components/lib/column/column.d.ts
didix16: Well this is just for the case that submit is true and you want to force that the cell contains a valid value. Since submit is false only when cancelling edit cell on Escape key and cellValidator only fires in that case, I just needed to add some logic to force call cellValidator even if the cell value was submited.

https://github.com/primefaces/primereact/blob/133e2d02559e9eccfe96b1e55a5489071ea90c10/components/lib/datatable/BodyCell.js#L181-L182

In my case, I have the autocomplete component which must have a valid value (i.e, the value must be an object picked from the list) and when the switchCellToView method is called I need to force check that the value is correct to properly avoid close the cell but without that logic I cannot do that because as I said, cellValidator only is being executed when submit value is false which inly happens on cell edit cancel.

melloware:
Ok let me think about a better name for the prop then!

didix16:
Perfect, thanks! By the way, should I set onKeyDown event prevention tonbe done always and remove the Dialog property?

melloware:
Yes just default that on key down

didix16:
I just removed the property added to Dialog and now the onKeyDown propagation event is stoped by default. Also I restored back the Dialog api.

components/lib/column/column.d.ts
melloware: How about `cellEditValidateOnClose` ?

melloware (changes requested):

components/lib/dialog/DialogBase.js
```diff
@@ -295,6 +295,7 @@ export const DialogBase = ComponentBase.extend({
         resizable: true,
         rtl: false,
         showHeader: true,
+        stopPropagationOnKeyDown: false,
```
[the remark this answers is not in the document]
didix16: Oh I forgot to revert it. I gonna remove it now

components/lib/column/column.d.ts
didix16: Sounds good to me. I'll change now

melloware (approved):

melloware:
Assigned to PrimeTek for review!

nitrogenous:
Looks good! Superb

nitrogenous (approved):

PR #6627, opened 2024-05-15
Feat #6589 return option when optionValue is empty

### LINKED ISSUE #6589: Dropdown: Init-Value not working if options have label,value properties
### Describe the bug

If Dropdown is filled with `{label: string, value: any}[]` the inital value isn't selected.

### Reproducer

https://stackblitz.com/edit/vitejs-vite-2prkbq?file=src%2FApp.tsx

### PrimeReact version

10.6.5

### React version

18.x

### Language

ALL

### Build / Runtime

Vite

### Browser(s)

_No response_

_No response_

### Expected behavior

_No response_

KirilCycle:
### Defect Fixes
Fix #6589

Similar to primevue, returns the entire option object if the desired optionValue is missing

### DIFF
--- a/components/lib/dropdown/Dropdown.js
+++ b/components/lib/dropdown/Dropdown.js
@@ -890,7 +890,7 @@ export const Dropdown = React.memo(
         };
 
         const getOptionValue = (option) => {
-            return props.optionValue ? ObjectUtils.resolveFieldData(option, props.optionValue) : ObjectUtils.resolveFieldData(option, 'value') || option;
+            return props.optionValue ? ObjectUtils.resolveFieldData(option, props.optionValue) : option;
         };
 
         const getOptionRenderKey = (option) => {


melloware:
cc @Rekl0w i don't think this is the correct fix.

Rekl0w:
No it's not. @melloware

sja-cslab:
I'm confused now.
#6518 is about `optionLabel` as your fix in #6520 
but this fix here is about `optionValue`.

melloware:
@sja-cslab you can see label and value do the same things..

```js
        const getOptionLabel = (option) => {
            return props.optionLabel ? ObjectUtils.resolveFieldData(option, props.optionLabel) : ObjectUtils.resolveFieldData(option, 'label') || option;
        };

        const getOptionValue = (option) => {
            return props.optionValue ? ObjectUtils.resolveFieldData(option, props.optionValue) : ObjectUtils.resolveFieldData(option, 'value') || option;
        };
```

sja-cslab:
@melloware sure I noticed that but thought @Rekl0w tried to roll that back - just asking because that would fix the issue #6589 I guess.

PR #6638, opened 2024-05-17
Fix #6632 InputOtp keyboard navigation

### LINKED ISSUE #6632: OTPInput Keyboard Support in Forms
Using the OtpInput component, the following cases are being observed.

1° When all fields are filled with a value and then attempting to use the Tab key to move between other components or even between the values, the action does not occur. Another case is when there is a form and we try to use Enter for submission, but the action also does not occur. However, if we leave the value incomplete in the component, both Tab and Enter work.
Example on the Prime website: [primereact.org/inputotp/#basic](https://primereact.org/inputotp/#basic)

2° When using the integerOnly prop, tab navigation does not work from the beginning of the component, thus also not functioning with Enter if in a form.
Example on the Prime website: [primereact.org/inputotp/#integeronly](https://primereact.org/inputotp/#integeronly)

https://github.com/orgs/primefaces/discussions/1845

KirilCycle:
### Defect Fixes
Fix #6632 ignore Tab and Enter pressing

### DIFF
--- a/components/lib/inputotp/InputOtp.js
+++ b/components/lib/inputotp/InputOtp.js
@@ -164,9 +164,11 @@ export const InputOtp = React.memo(
                 }
 
                 default: {
+                    console.log('DEF')
                     // Prevent non-numeric characters from being entered if integerOnly is true or if the length of the input is greater than the specified length
-                    if ((props?.integerOnly && !((event.code.startsWith('Digit') || event.code.startsWith('Numpad')) && Number(event.key) >= 0 && Number(event.key) <= 9)) || (tokens.join('').length >= props.length && event.code !== 'Delete')) {
-                        event.preventDefault();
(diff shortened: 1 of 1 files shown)


KirilCycle:
I can only reproduce the second bug in Firefox

Latest Firefox:

https://github.com/primefaces/primereact/assets/118115736/6c1d784a-7e42-4e79-8550-0616a6595bd0

Latest version of Chrome: 

https://github.com/primefaces/primereact/assets/118115736/0870cd12-c9ff-445b-840e-4f06d6db4a0e

I have checked other UI libraries like MUI, Chackra, ant-desing and they all behave the same, I also noticed that this error appears in some specific tag combinations, for example the ant-desing example works fine until I remove the h5 title inside container [input-otp](https://ant.design/components/input#components-input-demo-otp)

melloware:
Why not just this...

``js
                case 'Tab':
                case 'Enter': {
                    // allow these keys
                    break;
                }
```

basically if Enter or Tab just allow the key.

KirilCycle:
> Why not just this...
> 
> ```js
>                 case 'Tab':
>                 case 'Enter': {
>                     // allow these keys
>                     break;
>                 }
> ```
> 
> basically if Enter or Tab just allow the key.

Yes, i noticed that

melloware (approved):

PR #6643, opened 2024-05-20
Fix #6639: VirtualScroller improve useUpdate comparison

### LINKED ISSUE #6639: VirtualScroller: does not update items if only rows length changes
### Describe the bug

VirtualScroller with both directions does not update items if the number of columns changes but not the number of rows.
The error is in useUpdateEffect that compares prev items length with current items length.
This will return false if only the number of columns has changed.

Note that itemSize needs to be set to a constant for the issue to occur.

### Reproducer

[Demo on stackblitz](https://stackblitz.com/edit/vitejs-vite-fw2vsw?file=src%2FApp.tsx)

### PrimeReact version

9.6.0+

### React version

18.x

### Language

TypeScript

### Build / Runtime

Vite

### Browser(s)

_No response_

_No response_

### Expected behavior

_No response_

melloware:
Fix #6639: VirtualScroller compare using deepEquals

### DIFF
--- a/components/lib/virtualscroller/VirtualScroller.js
+++ b/components/lib/virtualscroller/VirtualScroller.js
@@ -568,7 +568,29 @@ export const VirtualScroller = React.memo(
         }, [numToleratedItemsState]);
 
         useUpdateEffect(() => {
-            if (!prevProps.items || prevProps.items.length !== (props.items || []).length) {
+            // Check if the previous/current rows array exists
+            const prevRowsExist = prevProps.items !== undefined && prevProps.items !== null;
+            const currentRowsExist = props.items !== undefined && props.items !== null;
+
+            // Get the length of the previous/current rows array, or 0 if it doesn't exist
+            const prevRowsLength = prevRowsExist ? prevProps.items.length : 0;
+            const currentRowsLength = currentRowsExist ? props.items.length : 0;
+
+            // Check if the length of the rows arrays has changed
+            let valuesChanged = prevRowsLength !== currentRowsLength;
+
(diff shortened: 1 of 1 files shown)


mamsellem:
(copied from comment on the ticket #6639).

I don't think using deepequals is a good idea.
We are only interesteed in updating the bounds of the virtual scroller, so only the dimensions of the items matter, not the items themselves.

This was my fix for this issue:

line 570 of VirtualScroll.js
useUpdateEffect(() => { if (!prevProps.items || prevProps.items.length !== (props.items || []).length || (both && prevProps.items[0].length !== (props.items ? props.items[0].length : 0))) { init(); } 

What do you think ?

melloware:
Yes but your fix presupposes `item[0]` is an array of things etc.  To me that seems riskier than a deep equals where IF the collection changes at all it must re-render?

mamsellem:
My understanding is that when virtualScroller "both" property is set to true, then items should be an array of arrays, of the same length, so items[0] must be an array.

There is another occurence of items[0] at line 250 of virtualScroller.js:
```
 const getLast = (last = 0, isCols) => {
            return props.items ? Math.min(isCols ? (props.columns || props.items[0])?.length || 0 : (props.items || []).length, last) : 0;
        };
```

melloware:
understood let me look deeper

melloware:
OK updated, I made it a little more readable and understandable for the next person!

mamsellem:
Thank you for taking the time to rewrite the code and make it more readable.
A small optimization would be make the calculations only if both is true, but it's really minor.

PR #6685, opened 2024-05-27
Fix onEditorCallback row data not being updated

### LINKED ISSUE #6683: DataTable: Tabulation is not working propertly. Also  editorCallback does not reflect the new value in some cases.
### Describe the bug

Hello there.

After updating to the last version 10.6.6, I'm facing new issues with DataTable... Now tabbing between cells seems broken. What I'm experimeting is that now, changing to next cell is not inmediate, I mean, current cell lost focus but on tab, the next cell is not opened nor focused (I don't know where the tab focus goes). You have to enter Tab again to change to the next cell... 

You can experiment the behaviour in the reproducer.

Also going backwards (Shift + Tab) It seems don't work. It stucks on the cell. However if you keep pressing Shift + Tab, eventually it jumps backwards.

And finally, if I go to a cell where I have an Autocomplete component, I select a value from it then when I press Tab does nothing. It does not change to next cell.

This is because when executing cellEditValidator, the event is returning a value which is not the desired value. If I type "a" on autocomplete component on a cell table, it updates the the row value to that value. But if I select the autocomplete value using the keyboard (arrows + enter), which the resulting value is an object, the row value seems not to be updated. However the editorCallback is executed but the internal rowData is not updated. @kl-nevermore I think your change in #6641  needs to be fixed for that case, when updating the rowData using the editorCallback.

What I've noticed is that `...editingRowDataStateRef.current` on `switchCellToViewMode` does not contains the updated value from editorCallback. So The useEffect() is not being executed after calling `setEditingRowDataState(editingRowData);` on editorCallback and thus, when switchToCellViewMode ends, sets the same old value that row has, avoiding the new value.

Hope the errors are understanded.

### Reproducer

https://stackblitz.com/edit/vitejs-vite-9p4ifk?file=src%2FExample.tsx

### PrimeReact version

10.6.6

### React version

18.x

### Language

TypeScript

### Build / Runtime

Vite

### Browser(s)

Any

To reproduce the TAB error.

1. Go to a editable cell on table
2. Press TAB. See that the next cell is not being selected
3. Press TAB again. Next cell is selected
4. Try to go backward using Shift + TAB. If the cell is not an autocomplete, the cell closes and does nothing else.
5. Press again shift + TAB on the same cell. The same cell is focused again, instead of previous.

To reproduce the Autocomplete error:

1. Go to a cell with autocomplete edior.
2. Type "a" then press TAB.
3. The autocomplete value has "Agent007" but the value returned in validator is false and thus the cell is not closing.

You can see the console.logs to track the value changes.

### Expected behavior

For the TAB.

That works like before, on TAB, change to next cell inmediatly and also Shift + TAB returns to previous cell and inmediatly.

didix16:
Fix editorCallback not updating row value on
Fix #6683


didix16:
For some reason, the bot does not detect the Issue Link in the description 😕

melloware:
No worries I linked it.

melloware (approved):

Issue #6680, opened 2024-05-27
Tooltip: Escape key not dismissing tooltip (WCAG 2.1 1.4.13)


kishore5242:
### Describe the bug

The tooltip displayed may obscure adjacent contents and user may want to dismiss it by pressing Esc key without moving the mouse.

![image](https://github.com/primefaces/primereact/assets/15242152/c3d72e49-97ce-4361-8f2d-c5f621d41624)

![image](https://github.com/primefaces/primereact/assets/15242152/daeb3ab0-ac56-457c-8185-2d91fc238edf)

![image](https://github.com/primefaces/primereact/assets/15242152/2bce0538-ab47-45c7-bfcb-16bf5a36c383)

**Prime React document also says that we support `escape` key.**
![image](https://github.com/primefaces/primereact/assets/15242152/042210af-836e-4841-81a5-aefbc498afdd)

**Also as per WCAG 1.4.13 -**
![image](https://github.com/primefaces/primereact/assets/15242152/2d7533a2-9425-426e-b772-17c2fa6b1519)

### Reproducer

https://stackblitz.com/edit/vitejs-vite-sihhuh?file=src%2FApp.tsx

### PrimeReact version

10.5.3

### React version

17.x

### Language

TypeScript

### Build / Runtime

Create React App (CRA)

### Steps to reproduce the behavior

1. Hover over a component that displays a Tooltip
2. Try to dismiss the tooltip by pressing `escape` key

### Expected behavior

`escape` key closes the Tooltip

melloware:
Thanks for reporting.  Definitely a missing feature.

akshayaqburst:
Fixed the issue in my local. Will send the Pull Request soon.

PR #6687, opened 2024-05-27
fix: #6680, Tooltip: Escape key not dismissing tooltip (WCAG 2.1 1.4.13)

akshayaqburst:
### Defect Fixes
fix: #6680, Tooltip: Escape key not dismissing tooltip (WCAG 2.1 1.4.13)


akshayaqburst:
Anyway, we have to specify closeOnEscape in the tooltipOptions prop, which is already specified in the documentation.

                <InputText value={value} onChange={(e) => setValue(e.target.value)}  tooltip="tooltip content" tooltipOptions={{closeOnEscape: true}} />

components/lib/hooks/useGlobalOnEscapeKey.js
```diff
@@ -5,6 +5,7 @@ import { DomHandler } from '../utils/Utils';
  * Priorities of different components (bigger number handled first)
  */
 export const ESC_KEY_HANDLING_PRIORITIES = {
+    TOOLTIP: 10,
```
melloware, line 8: I wonder if tooltip should be 1200?  I would think tooltips should always be on top of everything else?

components/lib/tooltip/Tooltip.js
```diff
@@ -32,6 +32,13 @@ export const Tooltip = React.memo(
         const { ptm, cx, sx, isUnstyled } = TooltipBase.setMetaData(metaData);
 
         useHandleStyle(TooltipBase.css.styles, isUnstyled, { name: 'tooltip' });
+        useGlobalOnEscapeKey({
+            callback: () => {
+                hide();
+            },
+            when: props.closeOnEscape,
+            priority: [ESC_KEY_HANDLING_PRIORITIES.TOOLTIP, 0]
```
melloware, line 40: I think closeOnEscpae might be missing from the tooltip options in the typescript definition

melloware (changes requested):

components/lib/tooltip/Tooltip.js
akshayaqburst: Added. @melloware

melloware (approved):

PR #6692, opened 2024-05-28
Fix #6691: fix default editMode value

### LINKED ISSUE #6691: Doc say DataTable editMode default is "cell" but its Null
Doc:
https://primereact.org/datatable/#api.DataTable.props.editMode

Code:
https://github.com/primefaces/primereact/blob/063c100193ebdd8518b143fa34f962ed21e69ba8/components/lib/datatable/DataTableBase.js#L444C9-L444C24

Rekl0w:
Fix #6691: fix default editMode value

### DIFF
--- a/components/doc/common/apidoc/index.json
+++ b/components/doc/common/apidoc/index.json
@@ -16584,7 +16584,7 @@
                             "optional": true,
                             "readonly": false,
                             "type": "string",
-                            "default": "cell",
+                            "default": "null",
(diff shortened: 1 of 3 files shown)


melloware:
This would make all tables now edit cell by default?  I think the TS def should update to default NULL.

melloware (changes requested):
see comments

Rekl0w:
You're right about it. I'll change the docs.

PR #6698, opened 2024-05-28
Fix: Datatable radiobuttons and types

### LINKED ISSUE #6697: Datatable: Unstyled mode doesn't work with radiobutton
### Describe the bug

I see there are some types wrong defined in Datatable props:
- filterOperatorDropdown
- filterMatchModeDropdown
- filterRemoveButton
- filterAddRuleButton
- filterClearButton
- filterApplyButton

Other types are missing:

- footerCell
- rowReorderIcon

And prop **radiobuttonwrapper** is not used nowhere.

Also, when I was trying to style radiobuttons in unstyled mode is not working. Classnames setted in PT are not applied to the column radiobutton (this is the example I'm talking about https://primereact.org/datatable/#radiobutton_row_selection).

I leave an example below with 2 datatables:
- One with unstyled mode not being applied (actual bug)
- Other prepared to check a pull request I will do later

[Open in StackBlitz](https://stackblitz.com/github/ivanpajon/datatable-radiobutton)

### Reproducer

https://stackblitz.com/github/ivanpajon/datatable-radiobutton

### PrimeReact version

10.6.6

### React version

18.x

### Language

TypeScript

### Build / Runtime

Next.js

### Browser(s)

_No response_

1. Open Stackblitz example
2. Inspect radiobuttons in "Deafult unstyled values" datatable
3. Only the div with the attribute **data-pc-name="radiobutton"** is working
4. Not styles applied to **data-pc-section="input"**, **data-pc-section="box"** nor **data-pc-section="icon"** like I show in the image below

image

### Expected behavior

Can customize radiobuttons properly through PT

ivanpajon:
### Defect Fixes
Fix #6697 

I fix the behavior of radiobuttons in datatable so they can be styled through PT, and although I don't even know what _ptm_, _ptmo_, _cx_, _sx_ functions mean and do... I figured out what they did and managed to make it work. So please, if there is a cleaner way or a 'Prime React' way you usually use, feel free to make the necessary changes.

Changes made:
- Types: Fix wrong defined types
- Types: Add missing types
- Types: Remvove types not used
- Datatable: Propagate unstyled prop inside Datatable internal components
- Radiobutton: Change prop input to radiobuttoninput
- Radiobutton: Change prop box to radiobuttonbox
- Radiobutton: Change prop icon to radiobuttonicon
- Update Tailwind PT
- Update Datatable doc unstyled example

Using the same example used to reproduce the bug you can see the result:

- Download stackblitz example to local
- Clone PrimeReact with changes from this pull request
- Build PrimeReact in local
- Link PrimeReact local library in Stackblitz local example
- Serve
- You will see 2 Datatables
- Styles in "Fixed unstyled values" Datatable should be correct.

[Open in StackBlitz](https://stackblitz.com/github/ivanpajon/datatable-radiobutton)

Table "Deafult unstyled values" not working:
default_unstyled_values

Fixed table "Fixed unstyled values":
fixed_unstyled_values

Example of PrimeReact [documentation](https://primereact.org/datatable/#radiobutton_row_selection):
radiobutton_primereact_doc

### ONLY RADIOBUTTONS, DON'T COMPARE REST OF STYLES

### DIFF
--- a/components/doc/common/apidoc/index.json
+++ b/components/doc/common/apidoc/index.json
@@ -13155,7 +13155,7 @@
                             "name": "filterOperatorDropdown",
                             "optional": true,
                             "readonly": false,
-                            "type": "DropdownPassThroughOptions",
+                            "type": "ColumnPassThroughType<HTMLAttributes<HTMLDivElement>>",
                             "description": "Uses to pass attributes to the Dropdown component."
                         },
                         {
@@ -13176,7 +13176,7 @@
                             "name": "filterMatchModeDropdown",
                             "optional": true,
                             "readonly": false,
-                            "type": "DropdownPassThroughOptions",
+                            "type": "ColumnPassThroughType<HTMLAttributes<HTMLDivElement>>",
                             "description": "Uses to pass attributes to the Dropdown component."
                         },
                         {
@@ -13190,7 +13190,7 @@
                             "name": "filterRemoveButton",
                             "optional": true,
                             "readonly": false,
-                            "type": "DropdownPassThroughOptions",
+                            "type": "ColumnPassThroughType<HTMLAttributes<HTMLButtonElement>>",
                             "description": "Uses to pass attributes to the Button component."
                         },
                         {
@@ -13204,7 +13204,7 @@
                             "name": "filterAddRuleButton",
                             "optional": true,
                             "readonly": false,
-                            "type": "DropdownPassThroughOptions",
+                            "type": "ColumnPassThroughType<HTMLAttributes<HTMLButtonElement>>",
                             "description": "Uses to pass attributes to the Button component."
                         },
                         {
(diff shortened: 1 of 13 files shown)


melloware (changes requested):
You don't need to update index.json you should update the .d.ts file and it gets autogenerated from the TypeScript files.

melloware:
See "npm run build:api" script in package.json

ivanpajon:
In fact I didn't manually change any the index.json, but after running **npm run dev:link:windows** and making the changes that file was auto generated.
If it is not needed when doing a commit maybe .gitignore should be updated...

melloware:
Hmm OK i didn't see you had updated `.d.ts` that is what triggers the updating of index.json.

melloware:
@nitrogenous this looks great to me but wanted you to review also.

nitrogenous:
Seems fine but we are currently working on some structural updates let me also check is there any conflicts with our current tasks. Appreciate it @melloware

PR #6699, opened 2024-05-28
Fix: metergroup typos

### LINKED ISSUE #6696: MeterGroup: CustomRenderProps don't work
### Describe the bug

In the MeterGroup component, there are ‘CustomRenderProps’ property. In the documentation, they are called :
image

There is no problem with ‘values’ but the others are not correct and do not work.

Here is an example of a console.log() in a ‘start’ method which takes an object of type CustomRenderProps as a parameter. The console.log() of this type return this (see image below).

image

You can see the ‘totalPercent’ attribute does not take an ‘L’ (l upper) and the second attribute is written as ‘precentages’ instead of ‘percentages’.

image
image

It is not possible to access the ‘totalPercent’ and ‘precentages’ attributes because they have different names to those in the documentation and the attributes 'totaLPercent' and 'percentages' return nothing.

### Reproducer

_No response_

### PrimeReact version

10.6.6

### React version

18.x

### Language

TypeScript

### Build / Runtime

Create React App (CRA)

### Browser(s)

Edge

_No response_

### Expected behavior

You would have to change the name of the ‘customRenderProps’ attributes to be able to access them.

ivanpajon:
### Defect Fixes
Fix #6696


melloware:
Don't we need to update the metergroup.d.ts as well?

melloware (changes requested):
See comment

ivanpajon (commented):
I don't see any comment...

One of the changes I made was updating the metergroup.d.ts

melloware (approved):

PR #6703, opened 2024-05-29
Fix #6058 Dropdown: Select value with Enter when optionValue is set

### LINKED ISSUE #6058: Dropdown: Loses value if value changed via ENTER key
### Describe the bug

Dropdown is empty as onChange event value property gets undefined if Dropdown has optionValue set.

Note:
In the stackblitz I created two dropdowns. One with optionValue and one without. The one without is working as expected.

### Reproducer

https://stackblitz.com/edit/vitejs-vite-tawz4r?file=src%2FApp.tsx

### PrimeReact version

10.5.1

### React version

18.x

### Language

TypeScript

### Build / Runtime

Vite

### Browser(s)

Chrome 121

1) Open dropdown (via mouse or tabbing)
2) Select a value with cursor keys
3) Press ENTER

### Expected behavior

After pressing ENTER value change should be reflected.

GabriellDatacamp:
Fix #6058 

When optionValue was set, and try to select the value with `Enter` key, the value is not selected.

In my correction, I saw that the `selectItem` function already has the `getOptionValue` function, making it unnecessary to call it within the `onOptionSelect` function.


melloware:
I think this reverts a fix that was specifically made for another scenario.

melloware (approved):

matthijsgroen:
This is quite a breaking change, when setting optionValue, normally you get a prop of an item object back in the onChange, now you always get the entire item. also setting the value using this optionValue is no longer working

melloware:
@matthijsgroen i think its already fixed for 10.7.1: https://github.com/primefaces/primereact/issues?q=is%3Aclosed+milestone%3A10.7.1

pradipjs (commented):
This created another issue when "optionValue" is provided. In this case, it started returning whole object instead of just a value of specific field provided in optionValue.

dotnet22:
> This created another issue when "optionValue" is provided. In this case, it started returning whole object instead of just a value of specific field provided in optionValue.

@melloware Yes, this is true. It created another bug. My entire app broke. Please help, as Dropdown is a very important component.

melloware:
OK all I reverted it here: https://github.com/primefaces/primereact/pull/6885

afjblt:
I don't understand what I should do so that enter in my dropdown now works together with the inputValue property
Can anyone help me with this example?
import React, { useState } from "react";
import { Dropdown } from 'primereact/dropdown';

export default function BasicDemo() {
    const [selectedCity, setSelectedCity] = useState(null);
    const cities = [
        { name: 'New York', code: 'NY' },
        { name: 'Rome', code: 'RM' },
        { name: 'London', code: 'LDN' },
        { name: 'Istanbul', code: 'IST' },
        { name: 'Paris', code: 'PRS' }
    ];

    return (
        <div className="card flex justify-content-center">
            <Dropdown value={selectedCity} onChange={(e) => setSelectedCity(e.value)} options={cities} optionLabel="name" 
                optionValue="code" placeholder="Select a City" className="w-full md:w-14rem" />
        </div>
    )
}

PR #6759, opened 2024-06-14
fix: highlight range of month in datePicker

### LINKED ISSUE #6653: Calendar: Incorrect styling when selecting month/year range
### Describe the bug

There is an issue with the Calendar component when using the properties view='month', and selectionMode='range'. When selecting months in the following sequence - selecting January and then a later month like May - the styling appears correctly. The problem arises when I first select a later month and then an earlier one, such as selecting May and then February. Although the range is successfully built, the styling is not applied correctly, and it appears as if the user has selected only one month instead of two. 
Screenshot 2024-05-17 at 17 01 20
Screenshot 2024-05-17 at 17 01 09

### Reproducer

https://stackblitz.com/edit/hyhoxe?file=src%2FApp.jsx

### PrimeReact version

10.5.0

### React version

18.x

### Language

TypeScript

### Build / Runtime

Vite

### Browser(s)

_No response_

1. Open calendar
2. Select May 2024
3. Select  February 2024

Only February is highlighted and May is not focused. The range is correctly filled.

### Expected behavior

Second month (May 2024) should be focused as well.

KumJungMin:
### Defect Fixes
- fix #6653
- To resolve this, the condition for highlighting months has been modified.

### DIFF
--- a/components/lib/calendar/Calendar.js
+++ b/components/lib/calendar/Calendar.js
@@ -2328,17 +2328,29 @@ export const Calendar = React.memo(
         };
 
         const isMonthSelected = (month) => {
-            if (isComparable()) {
-                let value = isRangeSelection() ? props.value[0] : props.value;
+            if (!isComparable()) return false;
 
-                if (isMultipleSelection()) {
-                    return value.some((currentValue) => currentValue.getMonth() === month && currentValue.getFullYear() === currentYear);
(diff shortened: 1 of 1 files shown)


melloware (approved):

melloware:
Thanks for the PR @KumJungMin looks great!

PR #6770, opened 2024-06-19
feat #6772 : defaultVisible prop added

### LINKED ISSUE #6772: Dialog: Add defaultVisible property
### Use Case

Imagine a scenario where a dialog needs to be immediately visible upon page load, such as for displaying an important announcement or a welcoming message. Here, the parent component doesn't require ongoing control over the visibility state post-initial render.

The `defaultVisible` prop is a boolean that, when set to `true`, ensures that the Dialog is displayed immediately upon mounting. Conversely, setting this prop to `false` or omitting it entirely will maintain the default behavior, where the Dialog is hidden initially and requires a trigger to become visible.

#### Changes

- Added `defaultVisible` prop to the Dialog component's API.
- Updated the Dialog component's internal state management to account for the `defaultVisible` prop.
- Ensured backward compatibility by defaulting the `defaultVisible` prop to `false`.

#### Usage

```jsx
<Dialog defaultVisible={true}>
    <!-- Dialog content goes here -->
</Dialog>
```

jpranays:
####  `defaultVisible` Prop to Dialog Component

Fix #6772

This pull request introduces a new `defaultVisible` prop to the Dialog component in our UI library. The `defaultVisible` prop allows developers to control the initial visibility of the Dialog when the component mounts.

#### Description

The `defaultVisible` prop is a boolean that, when set to `true`, ensures that the Dialog is displayed immediately upon mounting. Conversely, setting this prop to `false` or omitting it entirely will maintain the default behavior, where the Dialog is hidden initially and requires a trigger to become visible.

#### Changes

- Added `defaultVisible` prop to the Dialog component's API.
- Updated the Dialog component's internal state management to account for the `defaultVisible` prop.
- Ensured backward compatibility by defaulting the `defaultVisible` prop to `false`.

#### Usage

```jsx
<Dialog defaultVisible={true}>
    <!-- Dialog content goes here -->
</Dialog>
```

### DIFF
--- a/components/lib/dialog/dialog.d.ts
+++ b/components/lib/dialog/dialog.d.ts
@@ -354,6 +354,11 @@ export interface DialogProps {
      */
     transitionOptions?: CSSTransitionProps | undefined;
     /**
+     * Specifies the default visibility of the dialog.
+     * @defaultValue false
+     */
+    defaultVisible?: boolean | undefined;
+    /**
      * Specifies the visibility of the dialog.
      * @defaultValue false
      */
@@ -396,7 +401,7 @@ export interface DialogProps {
     /**
      * Callback to invoke when dialog is hidden (Required).
      */
-    onHide(): void;
+    onHide?(): void;
     /**
      * Callback to invoke when the mask is clicked.
      * @param {React.MouseEvent<HTMLElement>} event - Browser event.
--- a/components/lib/dialog/Dialog.js
+++ b/components/lib/dialog/Dialog.js
@@ -67,7 +67,15 @@ export const Dialog = React.forwardRef((inProps, ref) => {
     const [bindDocumentDragEndListener, unbindDocumentDragEndListener] = useEventListener({ type: 'mouseup', target: () => window.document, listener: (event) => onDragEnd(event) });
 
     const onClose = (event) => {
-        props.onHide();
+        if (props.onHide) {
+            props.onHide();
+        }
+
+        if (props.defaultVisible && isCloseOnEscape) {
+            setVisibleState(false);
+            setMaskVisibleState(false);
+        }
+
         event.preventDefault();
     };
 
@@ -405,6 +413,13 @@ export const Dialog = React.forwardRef((inProps, ref) => {
             setVisibleState(true);
         }
     }, [maskVisibleState]);
+
+    useUpdateEffect(() => {
+        if (props.defaultVisible) {
+            setVisibleState(true);
+            setMaskVisibleState(true);
+        }
+    }, [props.defaultVisible]);
 
     useUpdateEffect(() => {
         updateGlobalDialogsRegistry(true);
--- a/components/lib/dialog/DialogBase.js
+++ b/components/lib/dialog/DialogBase.js
@@ -297,6 +297,7 @@ export const DialogBase = ComponentBase.extend({
         showHeader: true,
         style: null,
         transitionOptions: null,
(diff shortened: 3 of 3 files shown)


components/lib/dialog/dialog.d.ts
```diff
@@ -353,6 +353,11 @@ export interface DialogProps {
      * The properties of CSSTransition can be customized, except for "nodeRef" and "in" properties.
      */
     transitionOptions?: CSSTransitionProps | undefined;
+    /**
+     * Specifies the default visibility of the dialog.
+     * @defaultValue false
+     */
+    defaultVisible?: boolean | undefined;
```
melloware, line 360: None of this is necessary just use the visible property if its true the dialog will be displayed even on inital page load.

```js
 /**
     * Specifies the visibility of the dialog.
     * @defaultValue false
     */
    visible?: boolean | undefined;
```

melloware (changes requested):

components/lib/dialog/dialog.d.ts
jpranays: You're right, but there's an additional aspect to consider. When it comes to closing the dialog, the user would need to manage the dialog's visibility state and an `onHide` function. By introducing the `defaultVisible` prop, could eliminate the need for the parent component to handle state hooks and `onHide` logic.

Since the Dialog component is already managing its own state, there's no need for the parent component to duplicate this effort just to handle initial visibility. Using `defaultVisible` simplifies the parent component and keeps the state management encapsulated within the Dialog component, making it easier to use.

melloware: @jpranays i would need to see a use case where this makes any sense to add this property.  The showcase examples all show pretty clear usage and require you to control the state with the `visible` and `onHide` properties.  THis new property just adds confusion.

jpranays: > @jpranays i would need to see a use case where this makes any sense to add this property. The showcase examples all show pretty clear usage and require you to control the state with the `visible` and `onHide` properties. THis new property just adds confusion.

Consider a scenario where a dialog should be displayed immediately upon page load, such as an important announcement or a welcome message. In this case, the parent component doesn't need to control the visibility state after the initial render.
While it's true that visible and onHide properties are essential for controlled components, the defaultVisible prop is intended for scenarios where the dialog's initial state is the only concern. It doesn't override or conflict with the existing visible and onHide properties but rather offers an alternative for specific use cases.

I hope this clarifies the value and use case for the defaultVisible prop. I believe it will enhance the flexibility and usability of the Dialog component without adding unnecessary confusion.

melloware: @jpranays can you open a ticket first with your use case and details so this can be associated to this PR.

jpranays: Ya sure

jpranays:
Any update on this ?

melloware:
> Any update on this ?

It is assigned to PrimeTek for their review.

nitrogenous:
Thank you for bringing this issue to our attention and for your valuable contribution. We truly appreciate the effort you put into this pull request.

However, after careful consideration, we have decided to reject the pull request. Our decision is based on the need to maintain synchronization across multiple UI libraries that we are currently developing. Additionally, we believe that this feature might lead to confusion among other users.

We hope you understand our decision, and we look forward to any future contributions you may have.

@jpranays

Issue #6666, opened 2024-05-24
Calendar: onMonthChange callback works incorrectly 


galimovkirill:
### Describe the bug

I expect that `onMonthChange` callback will be executed by clicking on arrow buttons for selecting next/prev month.
In fact, it executes only by changing month in specific dropdown (by clicking on button with current month).

### Reproducer

https://codesandbox.io/p/devbox/objective-lena-ypxjyy?file=%2Fsrc%2FApp.tsx%3A13%2C16

### PrimeReact version

10.6.5

### React version

18.x

### Language

TypeScript

### Build / Runtime

Create React App (CRA)

PR #6799, opened 2024-06-26
Fix: #6666 onMonthChange prop not being called when user increments / decrements month using arrow buttons.

PuppetFoam:
navForward / navBackward now calls onMonthChange (if defined).

Fix [#6666](https://github.com/primefaces/primereact/issues/6666)

**Description**

navForward / navBackward methods (called when user interacts with the increment / decrement month buttons) were not calling onMonthChange. 

**Changes**

- New variable to store incrementYear() / decrementYear() to avoid extra call.
- Check if onMonthChange prop is defined, and call with the month and year.

### DIFF
--- a/components/lib/calendar/Calendar.js
+++ b/components/lib/calendar/Calendar.js
@@ -361,11 +361,15 @@ export const Calendar = React.memo(
 
             if (currentView === 'date') {
                 if (newViewDate.getMonth() === 0) {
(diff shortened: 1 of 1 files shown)


melloware (changes requested):
run `npm run format` on the code its failing the prettier check

PuppetFoam (commented):
ran npm run format

Issue #6027, opened 2024-02-21
`Dropdown` component fails to expand the dropdown box after adding the `filter` and `editable` properties. 


Traveller23:
### Describe the bug

![chrome_8kzdDvKBNG](https://github.com/primefaces/primereact/assets/117014547/0c0983a1-932f-46f5-ac1c-83144b2efef6)

As shown in the image, clicking the button does not expand the dropdown box, nor does entering text.
As far as I remember, previous versions didn't have this problem.

### Reproducer

https://stackblitz.com/edit/vitejs-vite-4yjebs?file=src%2FApp.tsx

### PrimeReact version

10.5.1

### React version

18.x

### Language

TypeScript

### Build / Runtime

Vite

melloware:
Interesting the showcase doesn't have the same behavior? https://primereact.org/dropdown/#filter

Traveller23:
> Interesting the showcase doesn't have the same behavior? https://primereact.org/dropdown/#filter

Both `filter` and `editable` must be set at the same time for the problem to occur.

melloware:
Thanks!

PR #6810, opened 2024-07-01
Fix #6809: Make sure editable dropdown opens when user types

### LINKED ISSUE #6809: Dropdown: does not open OnEditableInputChange
### Describe the bug

If you have an editable dropdown in focus, and you start typing, the dropdown should open automatically. Currently, this does not happen. It will not open if, for example, you type "a", although it will open if you hit the `backspace` key. 

### Reproducer

_No response_

### PrimeReact version

10.6.6.

### React version

17.x

### Language

TypeScript

### Build / Runtime

Create React App (CRA)

### Browser(s)

_No response_

1. Go to the [editable primereact dropdown](https://primereact.org/dropdown/#editable)
2. Click inside the input field
3. Start typing "paris"
4. Observe as the dropdown does not open, and therefore does not automatically put the option "paris" in focus.
5. Hit the `backspace` key
6. Observe as the dropdown opens.

### Expected behavior

The editable dropdown should open when the user types.

peconomou929:
Fix #6809. 

This bug did not always exist. It was first introduced in #6032.

### DIFF
--- a/DEVELOPMENT.md
+++ b/DEVELOPMENT.md
@@ -23,8 +23,9 @@ C:\primereact> npm run dev:link:windows
 
 This will alter the bundler to only emit non-minified esm modules. The aliasing plugin has also been disabled for components. Once everything has been bundled (this can take a few minutes) you should keep this command running. It will allow for incremental builds as you develop in the `primereact/` directory.
 
-> [!NOTE]
-> It will be finished when the terminal displays: `[20xx-xx-xx 00:00:00] waiting for changes...`.
+> Note: if you get "Error: JavaScript heap out of memory", it may help to set the following node variable: `export NODE_OPTIONS=--max-old-space-size=8192`.
+
+> Note: The build will be finished when the terminal displays: `[20xx-xx-xx 00:00:00] waiting for changes...`.
 
 You will now `cd` into the `primereact/dist` directory and run:
 
--- a/components/doc/common/apidoc/index.json
+++ b/components/doc/common/apidoc/index.json
@@ -562,9 +562,8 @@
                             "type": "PrimeReactProviderProps"
                         },
                         {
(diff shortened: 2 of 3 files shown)


melloware:
@peconomou929 can you retest the issue in https://github.com/primefaces/primereact/issues/6027 that caused me to make the original change please?

melloware (changes requested):
I must have had a good reason to make this change

peconomou929:
@melloware I am testing my changes with a dropdown that has `filter=true` and `editable=true` and the issue in #6027 is not present.

melloware:
Nice!  From looking at this code again it has me wondering.  Why don't show() and hide() just check the overlayvisible state instead of everywhere in the code?

Like for example:

```js
const show = () => {
            if (overlayVisibleState) return;

            setFocusedOptionIndex(focusedOptionIndex !== -1 ? focusedOptionIndex : props.autoOptionFocus ? findFirstFocusedOptionIndex() : props.editable ? -1 : findSelectedOptionIndex());
            setOverlayVisibleState(true);
        };
```


I know its not related to your PR but doesn't it make sense that show() and hide() would just check the overlayVisibleState?

peconomou929:
> Nice! From looking at this code again it has me wondering. Why don't show() and hide() just check the overlayvisible state instead of everywhere in the code?

Hmm, I see your point. It might make sense to do that, but I'm not entirely sure it's a good idea (for now at least). For example, the `hide` function is

```
const hide = () => {
    setOverlayVisibleState(false);
    clickedRef.current = false;
};
```

In addition to setting the visible state, it also sets the `clickedRef`. What if somebody calls the `hide` function (perhaps one of the current usages of `hide`) and expects the `clickedRef` to always be set to false?

melloware:
ok i will take a look!

Issue #6561, opened 2024-05-03
Dropdown, Dialog: Blank dropdown when using an array of primary values


alefduarte:
#### Dropdown Display Issue:

When using an array of primary values, the dropdown is rendering blank. This issue was first observed during the upgrade from version 10.6.3 to 10.6.4. Even after upgrading to version 10.6.5, the problem persists.
[Dropdown Issue](https://stackblitz.com/edit/5fjpsq?file=src%2FApp.jsx,src%2Fmain.jsx)

![image](https://github.com/primefaces/primereact/assets/29279051/ed889e55-c208-4329-b80a-5f3e667add70)

~~#### Dialog onHide Called Twice:
The onHide event for the dialog component is being triggered twice. This issue can be observed in 
[Dialog Issue](https://stackblitz.com/edit/4re1hf?file=src%2FApp.jsx)~~

#### Datatable Paginator Issue:
In version 10.6.4, the datatable paginator was broken, throwing an error "Objects are not valid as a React child". This issue has been addressed in version 10.6.5.

Note: It's worth mentioning that upgrading from version 10.6.3 to 10.6.4 should ideally involve only minor changes. However, several breaking changes were observed, causing unexpected behavior and additional effort in testing.

---

The observed breaking changes between minor versions raise concerns about the level of professionalism in the library's release management process. Minor version updates are typically expected to introduce incremental improvements and bug fixes without major disruptions. However, the recent update from 10.6.3 to 10.6.4 resulted in significant changes, leading to unexpected issues. This lack of predictability in updates makes it challenging to plan and execute upgrades with confidence.

We kindly request the PrimeReact team to exercise more caution and thorough testing during the development and release of new versions. Alternatively, reconsidering the version naming convention to accurately reflect the magnitude of changes might help manage expectations and mitigate surprises for users.

We appreciate the efforts of the PrimeReact team and apologize if our feedback comes across as harsh. However, it's essential to highlight these issues to improve and enhance the reliability of the library.

### Reproducer

https://stackblitz.com/edit/5fjpsq?file=src%2FApp.jsx,src%2Fmain.jsx

### PrimeReact version

10.6.5

### React version

18.x

### Language

TypeScript

### Build / Runtime

Vite

### Browser(s)

Chrome

### Steps to reproduce the behavior

[Dropdown Issue](https://stackblitz.com/edit/5fjpsq?file=src%2FApp.jsx,src%2Fmain.jsx)
[...]

Issue #6617, opened 2024-05-13
Option with value 0 has different behavior than any other number in dropdown


espentveit:
### Describe the bug

Hi,

Seems like the behavior has been changed by commit 7312fe466e189581c53c3fe43bf3381db4b59455.

return props.optionValue ? ObjectUtils.resolveFieldData(option, props.optionValue) : ObjectUtils.resolveFieldData(option, 'value') || option;

The || here passes option through if the value is 0.

0 || "Test" == "Test"
true

### PrimeReact version

10.6.5

### React version

18.x

### Language

TypeScript

### Build / Runtime

Create React App (CRA)

melloware:
cc @Rekl0w

melloware:
it probably needs to be `ObjectUtils.isNotEmpty(props.optionValue)`

Rekl0w:
No, it is weird that i tried almost everthing but not works for this. I also tried to use old version of this function but still the same behavior. I guess this is another bug. @melloware

espentveit:
The case where I tested was for the one with the 'value' key set. Maybe the isNotEmpty should check the output of resolveFieldData?

```javascript
const getOptionValue = (option) => {
  if (props.optionValue) {
    return ObjectUtils.resolveFieldData(option, props.optionValue);
  } else {
    const optionValue = ObjectUtils.resolveFieldData(option, "value");
    return ObjectUtils.isNotEmpty(optionValue) ? optionValue : option;
  }
};
```

Rekl0w:
Both codes causes the blank dropdown options.

![image](https://github.com/primefaces/primereact/assets/91488737/4f43c3a6-ee8a-4ec1-8cdc-5b5bc347ce13)

espentveit:
Ok. I'll test some more tomorrow

melloware:
Fixed with: https://github.com/primefaces/primereact/pull/7083

PR #6872, opened 2024-07-11
Re-fix primefaces#6561: Return scalar values directly as a label for Dropdown

### LINKED ISSUE #6858: Dropdown: Object not valid when you use an Empty String   ""
### Describe the bug

In my system i have an object with 2 dropdown, we can see the page in the image below:
Captura de Tela 2024-07-09 às 12 13 46

Each dropdown is an object. The first named categoriaProduto and second named fornecedor. Below the objects interface:
CategoriaProduto
export interface ICategoriaProduto {
    id: number
    descricao: string
    status: boolean
}

export const initCategoriaProduto = (): ICategoriaProduto => ({
    id: 0,
    descricao: '',
    status: false
})

Fornecedor:
export interface IFornecedor {
    id: number
    descricao: string
    status: boolean
}

export const initFornecedor = (): IFornecedor => ({
    id: 0,
    descricao: '',
    status: true
})

The 2 objects have a method to return a valid instance of the object. This instance have a simple values id (like a code in english) equals 0 (zero), descricao (description in english) equals '' and status equals true or false.

I have a principal object named ProdutoAtendimento. ProdutoAtendimento have some properties and 2 of this properties are categoriaProduto and fornecedor. Below we can see the object ProdutoAtendimento:
Captura de Tela 2024-07-09 às 12 29 49

In component, i just have a dropdown using ProdutoAtendimento object and the properties categoriaProduto and fornecedor:
Captura de Tela 2024-07-09 às 12 14 44

So, in the new version of primereact, 10.7.0, when i create an instance of categoriaProduto or fornecedor with the description equals a empty string like '' i got the error below:
Captura de Tela 2024-07-09 às 12 15 46

When i chance the properties description to any string value like 'test', the component works fine.

Thats the problem, '' is a valid string to init my component and its works fine in 10.6.6 versions and all versions. 

So, to fix my application i need to go to categoriaProduto and fornecedor objects and change de instance method to use a fake value in descricao properties to my app works. 

export interface ICategoriaProduto {
    id: number
    descricao: string
    status: boolean
}

export const initCategoriaProduto = (): ICategoriaProduto => ({
    id: 0,
    descricao: 'test', //i put this fake value here and "fix" the problem
    status: false
})

export interface IFornecedor {
    id: number
    descricao: string
    status: boolean
}

export const initFornecedor = (): IFornecedor => ({
    id: 0,
    descricao: 'test', //i put this fake value here and "fix" the problem
    status: true
})

Even though my paleative solution works, it still seems like a bug in the Dropdown component.
and it always worked in previous versions

### Reproducer

_No response_

### PrimeReact version

10.7.0

### React version

18.x

### Language

TypeScript

### Build / Runtime

Next.js

### Browser(s)

_No response_

When you use a default value to an object with de description propertie with the default value equals an empty string like "", the dropdown component don't work. and it always worked in previous versions.

i put a fake value to "fix" my problem:
{
    id: 0,
    descricao: 'test', // fake value here to fix my problem
    status: true
}

### Expected behavior

The dropdown component needs to work, with a default value like the object below:
{
    id: 0,
    descricao: '',
    status: true
}

VytautasLozickas:
Fix #6858 

While fixing #6858 (#6871) I apparently broke #6561 (Support for an array of simple values as options for Dropdown). This should fix it for good :)

### DIFF
--- a/components/lib/dropdown/Dropdown.js
+++ b/components/lib/dropdown/Dropdown.js
@@ -886,7 +886,11 @@ export const Dropdown = React.memo(
         };
 
         const getOptionLabel = (option) => {
-            const optionLabel = props.optionLabel ? ObjectUtils.resolveFieldData(option, props.optionLabel) : option ? option['label'] : ObjectUtils.resolveFieldData(option, 'label');
+            if (ObjectUtils.isScalar(option)) {
+                return `${option}`;
+            }
+
+            const optionLabel = props.optionLabel ? ObjectUtils.resolveFieldData(option, props.optionLabel) : option['label'];
 
             return `${optionLabel}`;
         };
--- a/components/lib/utils/ObjectUtils.js
+++ b/components/lib/utils/ObjectUtils.js
@@ -386,6 +386,15 @@ export default class ObjectUtils {
         return /^[a-zA-Z\u00C0-\u017F]$/.test(char);
     }
 
+    static isScalar(value) {
+        return value != null && (
+            typeof value === 'string' ||
+            typeof value === 'number' ||
+            typeof value === 'bigint' ||
+            typeof value === 'boolean'
+        );
+    }
+
     /**
      * Firefox-v103 does not currently support the "findLast" method. It is stated that this method will be supported with Firefox-v104.
      * https://caniuse.com/mdn-javascript_builtins_array_findlast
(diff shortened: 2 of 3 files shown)


melloware (approved):

melloware:
@VytautasLozickas does this fix this one also: https://github.com/primefaces/primereact/issues/6617

VytautasLozickas:
@melloware Yes it does :)

melloware:
got it i see its the value not the label. thanks!

PR #6933, opened 2024-07-27
Fix #6927: Removing chips in multiselect cause open/close multiselect

### LINKED ISSUE #6927: Removing chips in multiselect cause open/close multiselect
### Describe the bug

We think it's not good user experience. If you agree could you please fix it. Thanks
https://primereact.org/multiselect/#chips

### Reproducer

https://primereact.org/multiselect/#chips

### PrimeReact version

10

### React version

18.x

### Language

TypeScript

### Build / Runtime

Vite

### Browser(s)

_No response_

_No response_

### Expected behavior

Multiselect stay closed when removing chips

meddvedev:
Fix #6927: Removing chips in multiselect cause open/close multiselect

### DIFF
--- a/components/lib/multiselect/MultiSelect.js
+++ b/components/lib/multiselect/MultiSelect.js
@@ -158,7 +158,7 @@ export const MultiSelect = React.memo(
         };
 
         const onClick = (event) => {
-            if (!props.inline && !props.disabled && !props.loading && !isPanelClicked(event) && DomHandler.getAttribute(event.target, 'data-pc-section') !== 'removetokenicon' && !isClearClicked(event)) {
+            if (!props.inline && !props.disabled && !props.loading && !isPanelClicked(event) && !isClearClicked(event)) {
                 overlayVisibleState ? hide() : show();
                 DomHandler.focus(inputRef.current);
                 event.preventDefault();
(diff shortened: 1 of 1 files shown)


components/lib/multiselect/MultiSelect.js
```diff
@@ -158,7 +158,7 @@ export const MultiSelect = React.memo(
         };
 
         const onClick = (event) => {
-            if (!props.inline && !props.disabled && !props.loading && !isPanelClicked(event) && DomHandler.getAttribute(event.target, 'data-pc-section') !== 'removetokenicon' && !isClearClicked(event)) {
+            if (!props.inline && !props.disabled && !props.loading && !isPanelClicked(event) && !isClearClicked(event)) {
```
melloware, line 161: Are you sure this change is necessary this looks like its there for unstyled mode?  I think the change below is all you need?

melloware (changes requested):

components/lib/multiselect/MultiSelect.js
meddvedev: Yep, since we are calling `event.stopPropagation`, the code handling this event won't be executed at all, regardless of whether we pass `unstyled` or not.

melloware (approved):

PR #6982, opened 2024-08-03
Fix #6949: Stepper: onChangeStep function not initially set

### LINKED ISSUE #6949: Stepper: onChangeStep function not initially set
### Describe the bug

The Stepper component does not initially have the onChangeStep function set which results in a warning message "Unknown event handler property `onChangeStep`. It will be ignored."

Once the component has rendered and changes are performed, the component works as normal.

### Reproducer

https://stackblitz.com/edit/8ur8c3?file=src%2FApp.jsx

### System Information

  System:
    OS: Windows 11 10.0.22631
    CPU: (20) x64 13th Gen Intel(R) Core(TM) i7-1370P
    Memory: 5.65 GB / 31.66 GB
  Binaries:
    Node: 20.11.0 - C:\Program Files\nodejs\node.EXE
    npm: 10.8.2 - C:\Program Files\nodejs\npm.CMD
  Browsers:
  npmPackages:
    primereact: ^10.8.0 => 10.8.0
    react: ^18 => 18.2.0

1. Add the Stepper Component
2. Set the prop onChangeSet={(step) => console.log(step)}

### Expected behavior

_No response_

meddvedev:
Fix #6949: Stepper: onChangeStep function not initially set

### DIFF
--- a/components/lib/stepper/StepperBase.js
+++ b/components/lib/stepper/StepperBase.js
@@ -107,7 +107,8 @@ export const StepperBase = ComponentBase.extend({
         __TYPE: 'Stepper',
         activeStep: 0,
         orientation: 'horizontal',
-        linear: false
+        linear: false,
+        onChangeStep: undefined
     },
     css: {
         classes,
--- a/components/lib/stepper/StepperBase.js
+++ b/components/lib/stepper/StepperBase.js
@@ -108,7 +108,7 @@ export const StepperBase = ComponentBase.extend({
         activeStep: 0,
         orientation: 'horizontal',
         linear: false,
-        onChangeStep: undefined
+        onChangeStep: null
     },
     css: {
         classes,


components/lib/stepper/StepperBase.js
```diff
@@ -107,7 +107,8 @@ export const StepperBase = ComponentBase.extend({
         __TYPE: 'Stepper',
         activeStep: 0,
         orientation: 'horizontal',
-        linear: false
+        linear: false,
+        onChangeStep: undefined
```
melloware, line 111:
```suggestion
        onChangeStep: null
```

We typically use NULL

melloware (changes requested):

components/lib/stepper/StepperBase.js
meddvedev: Thank you for your comment, I'll keep it in mind

melloware: Thanks for the PR.

Issue #6380, opened 2024-04-11
Calendar: unable typing time value


augustosnk12:
### Describe the bug

When using `timeOnly` with `mask` props of the Calendar component, and try to type the value, I'm unable to type the last value. To better see the problem please access the sandbox provided.

It's related to #6099 and #4928 

### Reproducer

https://codesandbox.io/p/sandbox/primereact-calendar-mask-cursor-error-10-6-3-349xfz?file=%2Fsrc%2FApp.js%3A23%2C13

### PrimeReact version

10.6.3

### React version

18.x

### Language

ES6

### Build / Runtime

Create React App (CRA)

### Steps to reproduce the behavior

Steps are provided in the sandbox 

### Expected behavior

Expect to be able to type the value when use the `mask` props

Issue #6099, opened 2024-03-06
Calendar: manual input value is wrongly corrected when typing


idzmag:
### Describe the bug

Using dateFormat="yy.mm.dd" type "2024.03." and next start typing "18" or something like that. After you type "1" the date is set to 2024.03.01 and makes typing "18" here impossible. 

### Reproducer

https://stackblitz.com/edit/vitejs-vite-uz69pt?file=src%2FApp.tsx

### PrimeReact version

10.5.1

### React version

18.x

### Language

ES6

### Build / Runtime

Vite

### Steps to reproduce the behavior

1. Start typing in Calendar 2024.03.
2. Try to type 18 next
3. NOT OK - input is changed to 2024.03.01

### Expected behavior

Parsing the date should be done only if provided input matches the format, so if dd is at the end .1 should not become .01 automatically

KirilCycle:
@melloware Hello! I want to try this, can u verify my solution please 
## Accept next digit, and move previous to start, when input is filled
For example you want to type 2024.03.18, after typing '1' to 2024.03
this will update input to  2024.03.01 and after entering '8', this will shift first dig '0' in cur cel, and then move last entered digits to achieve 2024.03.18

melloware:
@KirilCycle this one is complex but you are welcome to submit a PR so we can test it?

KirilCycle:
vueprime works fine at this case, i should take a look at their code base

melloware:
Good idea to review PrimeVue and compare.

Issue #5986, opened 2024-02-16
Calendar: text entry change


mwwhited:
### Describe the bug

If you try to change the entry value by typing after entering a date in the calendar control the input automatically changes and moves the cursor to the end of the input field.  

this can be duplicated on your demo page
https://primereact.org/calendar/

### PrimeReact version

10.5.1

### React version

17.x

### Language

TypeScript

### Build / Runtime

Create React App (CRA)

### Steps to reproduce the behavior

1) enter a value either by selection or typing
2) try changing part of the value such as the day just by keyboard

### Expected behavior

You should be able to edit values by keyboard alone

Issue #3728, opened 2022-12-01
Component: Calendar


Asderel:
### Describe the bug

Strange behaviour when writing a time on inputText of the Calendar component

The behaviour occur only when the component has a value and the props timeOnly, mask="99:99" and  showOnFocus={false}. It does not happen when showOnFocus={true}

After focusing the input and tart typing, the input doesn't show anything until two characters have been typed and then the cursor appears at the beginning of the input therefore, the first character you typed is pushed to the end of whatever your typing

### Reproducer

https://codesandbox.io/s/primereact-test-forked-uz7go2

### PrimeReact version

8.6.1

### React version

17.x

### Language

TypeScript

### Build / Runtime

Create React App (CRA)

### Browser(s)

Chrome

### Steps to reproduce the behavior

1. Set a value to the component
2. Click on inputtext
3. The contents of the inputtext gets focused
4. Type just one character
5. Nothing appears
6. After typing the second character the first one should appear at the end

### Expected behavior

1. Set a value to the component
2. Click on inputtext
3. The contents of the inputtext gets focused
4. Type anything
5. It show the value correctly while typing

PR #6986, opened 2024-08-05
fix #6984: calendar does not displays up-to-date value if date is manually entered

### LINKED ISSUE #4134: Calendar: Allow typing dates manually

### LINKED ISSUE #6984: Calendar: does not displays up-to-date value if date is manually entered
### Describe the bug

If date is **manually** changed in Calendar (manually type date in date input), and then it got changed by some external code, e.g. "reset" button, then calendar still displays old value (the one that was entered manually).

### Reproducer

https://codesandbox.io/p/sandbox/hopeful-moon-6q66wq

### System Information

 System:
    OS: Windows 11 10.0.22621
    CPU: (12) x64 12th Gen Intel(R) Core(TM) i5-1235U
    Memory: 1.28 GB / 15.71 GB
  Binaries:
    Node: 18.20.3 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.22.19 - ~\AppData\Roaming\npm\yarn.CMD
    npm: 10.7.0 - C:\Program Files\nodejs\npm.CMD
  Browsers:
  npmPackages:
    primereact: 10.7.0 => 10.7.0
    react: 18.3.1 => 18.3.1

In the demo above:

1. Click on "calendar" icon and pick some date in date picker - new date will appear in date input
2. Click on "reset" button - pay attention that date is empty now (this is expected)
3. Now enter the same date as in step 1 but manually (type it inside input)
4. Click on "reset" button - pay attention that date in date input will not change

Expected: date input should become empty, the same way as in step 2.

### Expected behavior

Calendar should correctly sync displayed date in date input with the date passed in props.

avasuro:
Fix #6984 
Fix #4134

### DIFF
--- a/components/lib/calendar/Calendar.js
+++ b/components/lib/calendar/Calendar.js
@@ -148,7 +148,6 @@ export const Calendar = React.memo(
                 const value = parseValueFromString(props.timeOnly ? rawValue.replace('_', '') : rawValue);
 
                 if (isValidSelection(value)) {
-                    isTypingRef.current = true;
                     updateModel(event, value);
                     updateViewDate(event, value.length ? value[0] : value);
                 }
@@ -3074,11 +3073,12 @@ export const Calendar = React.memo(
             const newDate = props.value;
 
             if (previousValue !== newDate) {
-                if (!isTypingRef.current) {
+                const isInputFocused = document.activeElement === inputRef.current;
+
+                // Do not update value in input if user types something in it:
+                if (!isInputFocused) {
                     updateInputfield(newDate);
                 }
-
-                isTypingRef.current = false;
 
                 // #3516 view date not updated when value set programatically
                 if (!newDate) return;
--- a/components/lib/calendar/Calendar.js
+++ b/components/lib/calendar/Calendar.js
@@ -24,7 +24,6 @@ export const Calendar = React.memo(
         const [overlayVisibleState, setOverlayVisibleState] = React.useState(false);
         const [viewDateState, setViewDateState] = React.useState(null);
         const [idState, setIdState] = React.useState(props.id);
-        const isTypingRef = React.useRef(false);
 
         const metaData = {
             props,


components/lib/calendar/Calendar.js
```diff
@@ -148,7 +148,6 @@ export const Calendar = React.memo(
                 const value = parseValueFromString(props.timeOnly ? rawValue.replace('_', '') : rawValue);
 
                 if (isValidSelection(value)) {
-                    isTypingRef.current = true;
```
melloware, line 151: Because isTypingRef is not longer used you should remove the `isTypingRef` declaration at the top of the file

melloware (changes requested):

avasuro:
@melloware 

As you written in your comment to the ticket, tested the following, works fine for me:
- #4134 
- #6099
- #3728
- #5986
- #6380

melloware (approved):

Issue #7014, opened 2024-08-09
Inserting a mention in between text results in text to be duplicated


chaowss:
### Describe the bug

Whenever inputing a mention between text (for instance if someone were to replace a mention after they typed the continuation of their message), the beginning of the text gets repeated and there is also the original text there.

https://github.com/user-attachments/assets/d27d0e4b-665d-41af-885e-c637d742e387

### System Information

This is the demo on https://primereact.org/mention/

### Steps to reproduce the behavior

input a mention between two strings, observe that the mention is appended to the old string with all of the characters preceding it repeated.

### Expected behavior

It should just input the mention correctly without repeating text.

chaowss:
Created the above PR ^ to fix

PR #7017, opened 2024-08-10
fixes issue primefaces#7014: inserting mentions results in duplicated text

chaowss:
Fixes: #7014 

The issue was that `event.target` is not the form element but actually a list item element, which doesn't have the selectionStart attribute.

I've also added a change where inserting the mention in front of a space does not result in a double space.

Note: this is my first time contributing to an open source project, let me know if there's anything I can do better!

### DIFF
--- a/components/lib/mention/Mention.js
+++ b/components/lib/mention/Mention.js
@@ -208,8 +208,10 @@ export const Mention = React.memo(
         };
 
         const selectItem = (event, suggestion) => {
-            const value = inputRef.current.value;
-            const selectionStart = event.target.selectionStart;
+            const input = inputRef.current;
+            const value = input.value;
+            const selectionStart = input.selectionStart;
+
             const spaceIndex = value.indexOf(' ', triggerState.index);
             const currentText = value.substring(triggerState.index, spaceIndex > -1 ? spaceIndex : selectionStart);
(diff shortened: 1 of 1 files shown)


chaowss:
I'm not sure how to re-trigger the linked issue in PRs link, is there a way to re-run the workflow?

components/lib/mention/Mention.js
```diff
@@ -218,7 +220,8 @@ export const Mention = React.memo(
                 const prevText = value.substring(0, triggerState.index);
                 const nextText = value.substring(spaceIndex > -1 ? selectionStart : triggerState.index + currentText.length);
 
-                inputRef.current.value = `${prevText}${selectedText} ${nextText}`;
+                inputRef.current.value = spaceIndex > -1 ? `${prevText}${selectedText}${nextText}` : `${prevText}${selectedText} ${nextText}`;
```
melloware, line 223: This ternary operator looks like both cases print out the exact same string?

melloware (changes requested):

components/lib/mention/Mention.js
chaowss: one of them contains a space.

So if the mention is inserted in front of a space it wont append another one. Not sure if thats too much of a behaviour change...

melloware (approved):

Issue #3819, opened 2022-12-23
DataTable: Global search won't reset when the field is empty on react 17


alefduarte:
### Describe the bug

When using the globalFilter parameter of the DataTable, it won't reset the searched items when clearing the search input. Let's say you have 100 items and search for the letter 'a', now you have 20 items, if you clear the input or erase the word, you still have 20 items, it won't reset when the input is empty.
You can see this behavior in the examples: https://www.primefaces.org/primereact/datatable/crud.
You start out with 30 items, if you search for the letter 'z' you will have 4 items, if you erase it,  you still have 4 items.

### Reproducer

https://codesandbox.io/s/soyt1t

### PrimeReact version

8.0.0

### React version

18.x

### Language

TypeScript

### Build / Runtime

Create React App (CRA)

### Browser(s)

Chrome

### Steps to reproduce the behavior

1. Search for any word on the search input.
2. Clear the input.
3. The items won't reset.

### Expected behavior

When clearing the input, all the existing items should be displayed.

PR #7104, opened 2024-09-02
fix(DataTable): Don't mix controlled and uncontrolled filter

### LINKED ISSUE #7103: DataTable: empyting global filter reset column's filter for uncontrolled filter
### Describe the bug

Discoverd while responding to https://github.com/primefaces/primereact/issues/6663#issuecomment-2263580845 and follow-up of https://github.com/primefaces/primereact/issues/3819

### Reproducer

https://stackblitz.com/edit/vzt525-tv5ff7?file=package.json

### System Information

```Shell
N/A
```

1. Open Stackblitz
2. Click on filter for Name
3. Type Bl (for start with): 3 rows
4. Click apply
5. in global filter type "a": 2 rows
6. erase the global filter: 5 rows

### Expected behavior

In step 6 I expected 3 rows (like step 3)

Et7f3:
In a components where `props.filters` is not set because user rely on generated filter object: if a column filter is applied, a global filter is applied and then the global filter set to "" the `props.filter` is used (null) instead of the generated filter (object with only one column defined).

The generated filter object before global filtering enter in action has no `global` key: so we revert this modifications.

Fix #7103

### DIFF
--- a/components/lib/datatable/DataTable.js
+++ b/components/lib/datatable/DataTable.js
@@ -1503,7 +1503,12 @@ export const DataTable = React.forwardRef((inProps, ref) => {
             filter(props.globalFilter, 'global', props.globalFilterMatchMode);
         } else {
             // #3819 was filtering but now reset filter state
-            setFiltersState(props.filters);
+            if (d_filtersState["global"]) {
+                let filters = { ...d_filtersState };
+                delete filters["global"];
+                setD_filtersState(filters);
+                onFilterApply(filters);
+            }
         }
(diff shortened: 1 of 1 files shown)


melloware:
Did you retest #3819 to make sure that scenario still works?

Et7f3:
> Did you retest #3819 to make sure that scenario still works?

Yes because my bug follow the same reproduction steps. The fix do reset too much (columns filter).

melloware (approved):

PR #7108, opened 2024-09-04
Fix #6784 - Toast onRemove message

### LINKED ISSUE #6784: Toast and message onRemove argument is not a ToastMessage type when it has _pId
### Describe the bug

Hi, 

I've seen some changes in toast/message code and I think these changes may have broken the onRemove call parameters:

image

In the first call to onRemove it may not have the ._pId attribute, but in the second call it has it, and instead of sending the ToastMessage in onRemove function it returns the _pId.

I think the onRemove call should be called with `messageInfo.message || messageInfo` instead of `removeMessage`.

Regards
Pablo García

### Reproducer

_No response_

### PrimeReact version

10.6.6

### React version

17.x

### Language

ES6

### Build / Runtime

Create React App (CRA)

### Browser(s)

_No response_

- Create a Toast or Message 
- Show the message
- Click on onRemove button (it contains a message as first parameter)
- Show the message again
- Click on onRemove button (it contains a number as first parameter)

### Expected behavior

- Create a Toast or Message 
- Show the message
- Click on onRemove button (it contains a message as first parameter)
- Show the message again
- Click on onRemove button (it contains a message as first parameter)

lukelalo:
### Defect Fixes
This PR changes the return parameters on onRemove method to pass the original message instead of changing to return _pId when existing, if the original message is there, because removeMessage.message doesn't exist when the _pId is defined. This PR fix #6784

### DIFF
--- a/components/lib/messages/Messages.js
+++ b/components/lib/messages/Messages.js
@@ -77,7 +77,7 @@ export const Messages = React.memo(
 
             setMessagesState((prev) => prev.filter((msg) => msg._pId !== messageInfo._pId && !ObjectUtils.deepEquals(msg.message, removeMessage)));
 
-            props.onRemove && props.onRemove(removeMessage.message || removeMessage);
+            props.onRemove && props.onRemove(messageInfo.message || removeMessage);
         };
 
         const onClose = (messageInfo) => {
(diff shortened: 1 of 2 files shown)


melloware:
Can you do the same for `Messages.js` it should be the same code.

lukelalo:
Ok

melloware (approved):

PR #7144, opened 2024-09-09
Fix #7134:Dropdown can now has options with keys named value #7141

### LINKED ISSUE #7134: Dropdown: When the options object has a field named value, the selected value automatically takes the value of that field instead of the entire object

mohamadbdeir:
Fix #7134

### DIFF
--- a/components/lib/dropdown/Dropdown.js
+++ b/components/lib/dropdown/Dropdown.js
@@ -825,7 +825,7 @@ export const Dropdown = React.memo(
         };
 
         const getOptionValue = (option) => {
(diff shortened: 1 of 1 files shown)


melloware:
fixed by https://github.com/primefaces/primereact/pull/7141

PR #7149, opened 2024-09-10
feat: make AdditionalProps argument in IconOptions type optional

alperari:
### Description
When switching from PrimeReact `v9.6.3` to `v10.8.2`, i had to do some migrations due to the feature introduced by https://github.com/primefaces/primereact/pull/4642.

Also, i noticed the inconsistency among the code in `utils.d.ts`, i.e. `AdditionalProps` argument is defined to be optional in some places. Yet, I believe, `IconOptions` was missed out.

This PR makes the `AdditionalProps` argument for the type `IconOptions` optional, so that you don't get the following TypeScript error:
screenshot

P.S. This relates to a closed issue https://github.com/primefaces/primereact/issues/4639 and I'm not sure whether to create new one.


melloware (approved):

melloware:
Thanks!

PR #7167, opened 2024-09-12
Fix #7166 - Datatable : fix ColumnFilter aria-controls

### LINKED ISSUE #7166: DataTable : fix ColumnFilter aria-controls
### Describe the bug

I think that the `aria-controls` attribute of the column filter button (in Datatable component) should only be set when the popup menu is visible in the DOM, otherwise a screen reader will give you a hint that you can move to the controlled element but it will have no effect because the controlled element doesn't exist.

### Reproducer

_No response_

### System Information

```Shell
"primereact": "10.8.2"
"react": "^18.2.0"
```

_No response_

### Expected behavior

_No response_

kyybo:
Set `aria-controls` for ColumnFilter only when the menu filter is visible in DOM.

fix #7166

### DIFF
--- a/components/lib/datatable/ColumnFilter.js
+++ b/components/lib/datatable/ColumnFilter.js
@@ -559,7 +559,7 @@ export const ColumnFilter = React.memo((props) => {
                 'aria-haspopup': true,
                 'aria-expanded': overlayVisibleState,
                 'aria-label': label,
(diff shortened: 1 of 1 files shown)


kyybo:
I don't understand why it says "PR does not seem to be linked to any issues". Am I missing something?

melloware (approved):

Issue #7087, opened 2024-08-26
Tree DragDrop doesn't allow drag and drop between nodes


michalmova:
### Describe the bug

PrimeReact - latest version 10.8.2 - Tree DragDrop doesn't allow drag and drop between nodes.

working version (7.1.0) sandbox: https://codesandbox.io/p/sandbox/amazing-gould-qy4m6r?from-embed=&workspaceId=4aecfb1c-06ef-4a24-a86f-2cc9c686780e

NOT working version sandbox (10.8.2): https://codesandbox.io/p/sandbox/ecstatic-heyrovsky-5ntj5z?from-embed=&workspaceId=4aecfb1c-06ef-4a24-a86f-2cc9c686780e

### System Information

    "primereact": "^10.8.2",
    "react": "18.2.0",

melloware:
Duplicate of https://github.com/primefaces/primereact/issues/7058

PR #7181, opened 2024-09-13
fix: Tree missing drag drop functionality

### LINKED ISSUE #7058: Tree: Missing Drag and Drop functionality
### Describe the bug

The Tree Component is missing some old functionality in the newest version. In older versions it was possible to change the order of the Items within the same hierarchy. If you dragged an item between two other items a line showed up and you could just drop the item there to place it between these other items. This functionality is totally missing in the newest version. Another one is quite similar. You are not able to drag and drop items to the top level (hierarchy). If you accidentally moved a top level item somewhere else, it is not possible to undo this action and put the node back to the top level.
I tried to visualize it with some screenshots. There is the current version and an older one (9.6.x)

Screenshot 2024-08-16 at 09 28 38
Screenshot 2024-08-16 at 09 28 52

Screenshot 2024-08-16 at 09 30 29
Screenshot 2024-08-16 at 09 30 36

### Reproducer

_No response_

### System Information

primereact: 10.8.2
react: 18

_No response_

### Expected behavior

The same functionality like in the previous primereact versions.
E.g. Drag and drop nodes/items to the top/root level of the tree and change the order of nodes within the same hierarchy.

### LINKED ISSUE #7156: `Tree` items can no longer be reordered
### Describe the bug

In previous versions, the items of the `Tree` component could be dragged to change their order, but now they can only be dragged to another item to make them a child of another item.

This issue can be reproduced directly on the official site.

### Reproducer

https://primereact.org/tree/#dragdrop

### System Information

```Shell
System:
    OS: Windows 10 10.0.19045
    CPU: (32) x64 13th Gen Intel(R) Core(TM) i9-13900
    Memory: 54.75 GB / 63.77 GB
  Binaries:
    Node: 21.7.1 - C:\Program Files\nodejs\node.EXE
    npm: 10.5.0 - C:\Program Files\nodejs\npm.CMD
    pnpm: 9.4.0 - C:\Program Files\nodejs\pnpm.CMD
  Browsers:
  npmPackages:
    primereact: ^10.8.2 => 10.8.2
    react: ^18.3.1 => 18.3.1
    tailwindcss: ^3.4.10 => 3.4.10
```

_No response_

### Expected behavior

_No response_

pascalbaetz:
Fix #7058
Fix https://github.com/primefaces/primereact/issues/7087
Fix #7156

Fixes the missing drag and drop functionality described in #7058
It was caused due to an incorrect comparison of the *droppable* prop which is `undefined` by default and not `true`

### DIFF
--- a/components/lib/tree/UITreeNode.js
+++ b/components/lib/tree/UITreeNode.js
@@ -847,7 +847,15 @@ export const UITreeNode = React.memo((props) => {
         const contentProps = mergeProps(
             {
                 ref: contentRef,
-                className: classNames(props.node.className, cx('content', { checked, selected, nodeProps: props, isCheckboxSelectionMode })),
+                className: classNames(
+                    props.node.className,
+                    cx('content', {
+                        checked,
+                        selected,
+                        nodeProps: props,
+                        isCheckboxSelectionMode
+                    })
+                ),
                 style: props.node.style,
                 onClick: onClick,
                 onDoubleClick: onDoubleClick,
@@ -961,7 +969,6 @@ export const UITreeNode = React.memo((props) => {
                 'aria-setsize': props.node.children ? props.node.children.length : 0,
                 'aria-posinset': props.index + 1,
                 onKeyDown: onKeyDown,
-                'aria-expanded': expanded,
                 'aria-selected': checked || selected
             },
             getPTOptions('node')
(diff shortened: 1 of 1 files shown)


components/lib/tree/UITreeNode.js
```diff
@@ -961,7 +969,6 @@ export const UITreeNode = React.memo((props) => {
                 'aria-setsize': props.node.children ? props.node.children.length : 0,
                 'aria-posinset': props.index + 1,
                 onKeyDown: onKeyDown,
-                'aria-expanded': expanded,
```
melloware, line 964: Are you sure you need to remove this aria-expanded?

melloware (changes requested):

components/lib/tree/UITreeNode.js
pascalbaetz: it was a duplicated entry so i removed it too due to an eslint error

melloware (approved):

melloware:
Thanks!

PR #7185, opened 2024-09-14
Fix #7184 : Usestate integration so that EnterKey can take custom CSS.

### LINKED ISSUE #7184: InputNumber: Pressing Enter Key should also take CSS just like Spacebar Key does.
### Describe the bug

While Pressing on increment or decrement button in InputNumber using Keys like Enter should also take CSS just like Spacebar.
CSS should look uniform if user uses Spacebar/Enter Key on Keyboard based navigation.

{A158FCF7-5D36-42B9-A4E0-06D0F06D86CA}

### Reproducer

https://stackblitz.com/edit/vitejs-vite-pcimsn?file=src%2FApp.tsx

### System Information

```Shell
PrimeReact: 10.8.2
React: 18.2.0
Browser: Chrome
```

1. Open the Stackblitz link.
2. Try to increment using mouse so that button gets focus.
3. Move the cursor away from buttons so that its not in hover state.
4. Try to increment/decrement using Enter Key and later try with Spacebar. 

### Expected behavior

Same CSS styles should take place on using both Keys.

Sumeet-2023:
### Defect Fixes
Fixes #7184 

I had to create custom CSS for Enter Key,
Spacebar: Activates the `:active` ( pseudo-class ) state because it mimics the press-and-hold behavior of a mouse click.
Enter key: Simply triggers the button's action (click event) and does not hold the `:active` state for any visible duration.

### DIFF
--- a/components/lib/inputnumber/InputNumber.js
+++ b/components/lib/inputnumber/InputNumber.js
@@ -16,11 +16,13 @@ export const InputNumber = React.memo(
         const context = React.useContext(PrimeReactContext);
         const props = InputNumberBase.getProps(inProps, context);
         const [focusedState, setFocusedState] = React.useState(false);
+        const [keyPressed, setKeyPressed] = React.useState(0);
         const metaData = {
             props,
             ...props.__parentMetadata,
             state: {
-                focused: focusedState
+                focused: focusedState,
+                pressed: keyPressed
             }
         };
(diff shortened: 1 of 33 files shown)


melloware:
I think there might be an easier way to do this.

Sumeet-2023:
Should i remove changes that i made in CSS files and wait for the review? 
If there is better approach to reach then i/you should close this pull request :)

PR #7188, opened 2024-09-15
Fix #7187: Provide readable input value based on selected option labels

### LINKED ISSUE #7187: MultiSelect renders input value as a stringified array including the entire item objects, but it should display a readable value based on the selected option labels
### Describe the bug

There are two main issues with the current approach:
- As a user with disabilities, I want to see the selected items in the combobox value, instead of a programmatic string.
- As a developer, I want to test selected options in a normal way:
```js
describe('MultiSelect', () => {
    it('should allow selecting multiple options', async () => {
        const user = userEvent();
        render(<MultiSelect aria-label="Countries" options={countries} optionLabel="name" />);

        await user.click(screen.getByRole('combobox', { name: 'Countries' }));
        await screen.findByRole('listbox', { name: 'Countries' });

        await user.click(screen.getByRole('option', { name: 'Rome' }));
        await user.click(screen.getByRole('option', { name: 'London' }));

        expect(screen.getByRole('combobox', { name: 'Countries' })).toHaveValue('Rome, London');
    });
});
```

In my view, the input value string should consist of the option labels - the values the user sees on the screen. The following logic should be enough to fix this issue:

```js
const getInputValue = (value = []) => {
    if (props.optionLabel && Array.isArray(value)) {
        return value.map((val) => getLabelByValue(val)).join(', ');
    }

    return value;
};

const inputProps = mergeProps(
    {
        ...otherProps,
        value: getInputValue(props.value),
    },
    ptm('input')
);
```

### Reproducer

_No response_

### System Information

```Shell
primereact: latest
react: 18.3.1
```

1. Open the basic example at [PrimeReact MultiSelect Doc](https://primereact.org/multiselect/) and choose a few options in the dropdown.
2. Open the browser devtools and check the input value.

Screenshot example:

Screenshot 2024-09-15 at 13 07 16

### Expected behavior

Screenshot 2024-09-15 at 13 13 52

iamkyrylo:
Fix #7187

### DIFF
--- a/components/lib/multiselect/MultiSelect.js
+++ b/components/lib/multiselect/MultiSelect.js
@@ -1021,6 +1021,14 @@ export const MultiSelect = React.memo(
             );
         };
 
+        const getInputValue = (value = []) => {
+            if (props.optionLabel && Array.isArray(value)) {
+                return value.map((val) => getLabelByValue(val)).join(', ');
+            }
+            
+            return value;
+        }
+
         const visibleOptions = getVisibleOptions();
 
         const hasTooltip = ObjectUtils.isNotEmpty(props.tooltip);
@@ -1082,7 +1090,7 @@ export const MultiSelect = React.memo(
                 'aria-expanded': overlayVisibleState,
                 disabled: props.disabled,
                 tabIndex: !props.disabled ? props.tabIndex : -1,
-                value: JSON.stringify(props.value),
+                value: getInputValue(props.value),
                 ...ariaProps
             },
             ptm('input')


melloware:
@iamkyrylo thank you I assigned to PrimeTek for review.

nitrogenous (approved):
lgtm

ivanpajon:
Removing line with JSON.stringify breaks this fix #6897

iamkyrylo:
> Removing line with JSON.stringify breaks this fix #6897

I'm not sure I understand the issue. This doesn’t break anything. The input value shouldn’t consist of serialized objects with all their attributes. Instead, the input value should be a simple string made up of the option labels, as this is what users see on the screen and select from the dropdown.

ivanpajon:
The issue is that not serializing the value to the input causes this:
image

And if you are using the component inside a form tag for example you will submit `[object Object], [object Object]` instead of the correct value of the input:
image

iamkyrylo:
@ivanpajon Please take a closer look at the changes in this PR, specifically [this line](https://github.com/primefaces/primereact/pull/7188/files#diff-1e4d4da32c316984372a1ab7058316c67c367e2713a830ddb6e8b3a144e64909R1026). With these changes, based on your example, the input value will be a simple string like `input="Module 1, Module 2"` instead of `input="[object Object], [object Object]"`.

ivanpajon:
@iamkyrylo yes, you are right, I have don't check it, sorry.

In my case I need to use a **custom object** as the value so I think now I should use **useOptionAsValue** prop. I have check it with useOptionAsValue prop and in this case input value is `[object Object], [object Object]`. You can see it in this [reproducer](https://stackblitz.com/edit/t8zhwn?file=src%2FApp.tsx) I made.

iamkyrylo:
Oh, I see what you mean and understand the problem. It’s not related to the `useOptionAsValue` prop. It comes from checking `props.optionLabel` in the condition. 

I was misled by the component types, which indicated that `optionLabel` has a default value of `label`. But, since the component is implemented in JS, it would need to have `defaultProps` defined to align with the type declaration file, but it doesn’t - there are no `propTypes` or `defaultProps`🤷‍♂️

Anyway, I can make a fix.

ivanpajon:
With optionLabel it works but not as intended.
I mean, if I set useOptionAsValue I want the whole object to be the value, which in the reproducer is doing it well but without serialize it. That is why input has `[object Object], [object Object]` as value instead of `[{"id":"fd9jlbv3s","name":"city_one","label":"City One"},{"id":"sdvb56n5f","name":"city_two","label":"City Two"}]`.

iamkyrylo:
This approach is technically incorrect. Exposing the entire value in the dom is not good for accessibility and security. You should keep in mind that **the input value in the dom is for users, not developers**. If you need to handle the option object, store it in the component's state rather than using a JSON string in the dom.

I’m not sure why the `useOptionAsValue` prop was added, but if it’s for the reason you mentioned, I'd suggest that the core team remove it to avoid confusion and maintain better accessibility practices. The `MultiSelect` component already allows you to manage the selected option directly using the option object when `optionValue` is not provided.

I’d like to get the core team's thoughts on this. @melloware @nitrogenous

melloware:
totally agree with @iamkyrylo JSON in a readable element is wrong and a code smell.

iamkyrylo:
I've submitted a follow-up fix - https://github.com/primefaces/primereact/pull/7216

ivanpajon:
I also don't like the approach with serialization.
I only want to mention that without that this component would never work with NextJS form action submissions. At least without extra boilerplate.

iamkyrylo:
@ivanpajon I'm going to point this out again so you don't confuse or mislead yourself or others — it's not recommended to serialize objects for the input value, regardless of which form implementation you're using (NextJS, Remix, native, etc.). Since `MultiSelect` is a controlled component, you need to manage its state anyway. Also, `MultiSelect` allows you to handle objects directly to define the selected options. So, there's no issue using this state for the request instead of relying solely on the `formData` object.

Check [NextJS documentation](https://nextjs.org/docs/app/building-your-application/data-fetching/server-actions-and-mutations) again, and you’ll find a convenient solution that suits your case.

Issue #7200, opened 2024-09-17
DataTable: 10.8.3 - Drag selection fails due to `TypeError: event.originalEvent is undefined` (Next.JS)


SolidAnonDev:
### Describe the bug

Drag selection fails and is in general having problems across version 10.8.2 and 10.8.3, but appears to be entirely broken in 10.8.3 in Next.JS

See #7192 for previous bug report regarding 10.8.2

### Reproducer

https://stackblitz.com/edit/stackblitz-starters-zx3txq?file=package.json

### System Information

```Shell
System:
    OS: Linux 5.0 undefined
    CPU: (4) x64 Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
    Memory: 0 Bytes / 0 Bytes
    Shell: 1.0 - /bin/jsh
  Binaries:
    Node: 18.20.3 - /usr/local/bin/node
    Yarn: 1.22.19 - /usr/local/bin/yarn
    npm: 10.2.3 - /usr/local/bin/npm
    pnpm: 8.15.6 - /usr/local/bin/pnpm
  npmPackages:
    primereact: ^10.8.3 => 10.8.3 
    react: ^18.3.1 => 18.3.1 
    tailwindcss: ^3.4.11 => 3.4.11 


"dependencies": {
    "@types/node": "^22.5.5",
    "@types/react": "^18.3.6",
    "@types/react-dom": "^18.3.0",
    "autoprefixer": "^10.4.20",
    "eslint": "^8.57.0",
    "eslint-config-next": "^14.2.11",
    "next": "^14.2.11",
    "postcss": "^8.4.47",
    "primereact": "^10.8.3",
    "react": "^18.3.1",
    "react-dom": "^18.3.1",
    "tailwindcss": "^3.4.11",
    "typescript": "5.2.2"
  }
```

### Steps to reproduce the behavior

1. In reproduction link, attempt to make a drag selection across datatable rows
2. An error will be thrown:

```
Unhandled Runtime Error

TypeError: event.originalEvent is undefined
Call Stack
event
node_modules/primereact/datatable/datatable.esm.js (2630:61)
allowDrag
node_modules/primereact/datatable/datatable.esm.js (2633:37)
event
node_modules/primereact/datatable/datatable.esm.js (3130:22)
onRowDragStart
node_modules/primereact/datatable/datatable.esm.js (2190:11)
_onDragStart
node_modules/primereact/datatable/datatable.esm.js (2419:14)
```

### Expected behavior

[...]

PR #7217, opened 2024-09-19
fix: #7200 Fix Drag Selection for Datatable

gcko:
### Defect Fixes
fix: #7200 Fix Drag Selection for Datatable

The existing implementation of `allowRowDrag` takes an "event" and passes it through to `allowDrag`, however `allowDrag` expects the `event` parameter to have `originalEvent` as a property. This PR fixes that issue.

### DIFF
--- a/components/lib/datatable/TableBody.js
+++ b/components/lib/datatable/TableBody.js
@@ -151,7 +151,7 @@ export const TableBody = React.memo(
         };
 
         const allowRowDrag = (event) => {
-            return (!allowCellSelection() && allowDrag(event)) || props.reorderableRows;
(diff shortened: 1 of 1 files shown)


melloware (approved):

konstantin-tarasov-volantetech:
I still experience the problem on **10.8.4**. Can't select cell's data. The cell is dragged instead.

Also would be good to update official documentation on [primereact.org](https://primereact.org) to **10.8.4**. The problem can be reproducing on the datatable's page.

AntonBiruk:
the bug is still present on 10.8.4

PR #7228, opened 2024-09-21
fix #7220: resolve global esc key listener conflict in Calendar component

### LINKED ISSUE #7220: Calendar: Global esc key listener with priority [600, 0] already exists
### Describe the bug

I encountered an issue when using two Calendar components on the same page as time pickers. The error appears after selecting time in first calendar and move the focus to second calendars.
This issue happens when the Calendar component is used as a time picker, not only as a date picker.

_Error: Unexpected: global esc key listener with priority [600, 0] already exists._
I believe the error is related to the `useGlobalOnEscapeKey` hook, which may be misconfigured or have a bug affecting the Calendar component's behavior.

### Reproducer

https://stackblitz.com/edit/stackblitz-starters-tuzydi?file=app%2Fpage.tsx

### System Information

```Shell
System:
    OS: macOS 14.6.1
    CPU: (8) x64 Intel(R) Core(TM) i5-8279U CPU @ 2.40GHz
    Memory: 593.41 MB / 8.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 21.2.0 - /usr/local/bin/node
    Yarn: 1.22.22 - /usr/local/bin/yarn
    npm: 10.2.4 - /usr/local/bin/npm
    Watchman: 2023.11.20.00 - /usr/local/bin/watchman
  Browsers:
    Chrome: 128.0.6613.138
    Safari: 17.6
  npmPackages:
    primereact: ^10.8.3 => 10.8.3 
    react: ^18 => 18.2.0 
    tailwindcss: ^3.3.0 => 3.4.0
```

1. Select a start time in the first Calendar component.
2. Without pressing any keys or clicking elsewhere, select an end time in the second Calendar component.  An error occurs at this point.

### Expected behavior

The calendars should function independently as time pickers without triggering any errors.

ozantekin:
Fix issue: #7220 

This also resolves the following issues:
- Fixed the same error in the Time section on the Calendar page.
Screen Shot 2024-09-21 at 23 09 10 PM
- Corrected the overlay behavior that did not open and close properly when the target was not correctly identified.

### DIFF
--- a/components/lib/calendar/Calendar.js
+++ b/components/lib/calendar/Calendar.js
@@ -2,7 +2,7 @@ import * as React from 'react';
 import PrimeReact, { PrimeReactContext, localeOption, localeOptions } from '../api/Api';
 import { Button } from '../button/Button';
 import { useHandleStyle } from '../componentbase/ComponentBase';
-import { useMergeProps, useMountEffect, useOverlayListener, usePrevious, useUnmountEffect, useUpdateEffect, useGlobalOnEscapeKey, ESC_KEY_HANDLING_PRIORITIES } from '../hooks/Hooks';
+import { useMergeProps, useMountEffect, useOverlayListener, usePrevious, useUnmountEffect, useUpdateEffect, useGlobalOnEscapeKey, ESC_KEY_HANDLING_PRIORITIES, useDisplayOrder } from '../hooks/Hooks';
 import { CalendarIcon } from '../icons/calendar';
 import { ChevronDownIcon } from '../icons/chevrondown';
 import { ChevronLeftIcon } from '../icons/chevronleft';
@@ -24,6 +24,8 @@ export const Calendar = React.memo(
         const [overlayVisibleState, setOverlayVisibleState] = React.useState(false);
         const [viewDateState, setViewDateState] = React.useState(null);
         const [idState, setIdState] = React.useState(props.id);
+        const isCloseOnEscape = overlayVisibleState && props.closeOnEscape;
+        const overlayDisplayOrder = useDisplayOrder('overlay-panel', isCloseOnEscape);
 
         const metaData = {
             props,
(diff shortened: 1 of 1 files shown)


components/lib/calendar/Calendar.js
```diff
@@ -1852,7 +1854,7 @@ export const Calendar = React.memo(
                 setOverlayVisibleState(true);
 
                 overlayEventListener.current = (e) => {
-                    if (!isOutsideClicked(e)) {
+                    if (!isOutsideClicked(e.target)) {
```
melloware, line 1857: This line looks incorrect as `isOutsideClicked` takes an `event` and gets its `event.target`?

melloware (changes requested):

components/lib/calendar/Calendar.js
ozantekin: When target is not included, the panel of the first calendar input stays open after making changes in the first input and shifting focus to the second calendar input.

Screen Shot 2024-09-22 at 15 01 01 PM

melloware: OK but if you look at the code 

```js
const isOutsideClicked = (event) => {
            return elementRef.current && !(elementRef.current.isSameNode(event.target) || isNavIconClicked(event.target) || elementRef.current.contains(event.target) || (overlayRef.current && overlayRef.current.contains(event.target)));
        };
```

That means if you pass `e.target` then inside isOutSideClicked they will be calling `target.target` does that make sense?  the Target is a DOM element and it won't have a target property?

ozantekin: Thanks for the clarification, I missed that detail. I'll remove the target and push the update. Can I address the other issue in a different PR?

ozantekin: noticed missing condition, added a fix to resolve the issue.

melloware (approved):

ozantekin:
> I am using `primereact` version `10.8.3` and the error persists `Uncaught Error: Unexpected: global esc key listener with priority [600, 0] already exists.`. It can be reproduced also in your demo. Steps to reproduce:
> 
> 1. Go to https://primereact.org/calendar/#time
> 2. Open "Time Only" picker and change **hour** to any other value. Just click up or down to change it. Specifically hours! Changing minutes seems to works fine.
> 3. Click on the sibling datepicker/timepicker. The application crashes.
>    When clicking `Esc` or non-datepicker element - it works though.
> 
> Maybe better to create a new task, right?

I've fixed this issue, but it hasn't been pushed to production yet. You can track further details here. #7220 @mrpsiho

Issue #7239, opened 2024-09-24
DataTable incorrectly indicates drop functionality for unrelated draggable elements


kttmv:
### Describe the bug

The DataTable component incorrectly indicates drop functionality when any draggable element is hovered over its rows, even if drop actions are not supported. The cursor changes to indicate that a drop is possible, even though dropping on a row doesn't actually perform any action unless it's specifically for reordering rows within the same DataTable.

### System Information

```Shell
System:
    OS: Windows 10 10.0.19045
    CPU: (20) x64 13th Gen Intel(R) Core(TM) i7-13800H
    Memory: 15.57 GB / 31.66 GB
  Binaries:
    Node: 20.15.1 - C:\Program Files\nodejs\node.EXE
    npm: 10.8.3 - C:\Program Files\nodejs\npm.CMD
    pnpm: 9.6.0 - ~\AppData\Local\pnpm\pnpm.EXE
  Browsers:
    Edge: Chromium (127.0.2651.74)
    Internet Explorer: 11.0.19041.3636
  npmPackages:
    react: 18.3.1 => 18.3.1
```

### Steps to reproduce the behavior

1. Navigate to the [DataTable page](https://primereact.org/datatable/)
2. Find any draggable element on the page. For example this can be:
   * A link (e.g., the PrimeReact logo or the Github link in the top-right corner of the page)
   * An image
   * Selected text
   * A row from another DataTable
   * A row from the same DataTable when reordering is disabled for that table
3. Click and drag the chosen element.
4. Hover the dragged element over any row in the DataTable.

Result: The cursor changes to indicate a possible drop action (a cursor that indicates "copy" appears).

https://github.com/user-attachments/assets/687bcf54-a630-47ab-a128-95b5992f629a

### Expected behavior

The DataTable should only indicate drop functionality in specific scenarios:
1. For DataTables with row reordering enabled: 
   * The cursor should only change when dragging a row from the same DataTable.
   * It should not react to other draggable elements from outside the DataTable.
2. For DataTables without row reordering functionality:
   * The cursor should not change at all when any element (including its own rows) is dragged over it.

melloware:
I think its related to this: https://github.com/primefaces/primereact/issues/7200

PR #7240, opened 2024-09-24
fix: #7239: Incorrect behaviour of dragging over DataTable rows

kttmv:
### Defect Fixes
https://github.com/primefaces/primereact/issues/7239

### Changes
Fixed onRowDragOver listener to prevent default action only when row dragging is active. Added an early return if rowDragging is inactive, avoiding unnecessary preventDefault() calls.

https://github.com/user-attachments/assets/27c8984d-83eb-4728-a192-e9e300b4da91

https://github.com/user-attachments/assets/24224481-76bb-47e6-8e50-9cf78fac4268


melloware:
@kttmv did this not fix your issue?

Issue #6682, opened 2024-05-27
Scrollpanel: pt barX is gets overwritten when mouse enters scrollpanel 


PrebenTjemsland:
### Describe the bug

The Pass Through for barX on Scrollpanel is overwritten when you hover on the panel. 

Initial:
element.style {
    opacity: 0;
    width: 40px;
} 

after hover: 
element.style {
    width: 97.2366%;
    left: 0%;
    bottom: -416px;
}

### Reproducer

https://stackblitz.com/edit/vitejs-vite-jcmten?file=src%2FApp.tsx

### PrimeReact version

10.6.3

### React version

18.x

### Language

TypeScript

### Build / Runtime

Create React App (CRA)

### Browser(s)

Chrome 125

### Steps to reproduce the behavior

Make a scrollpanel with content styled with style={{ margin: 8, width: '100%' }}
Add pt for barX
hover on panel

### Expected behavior

pt stays on mouse enter

gcko:
Hey @PrebenTjemsland I created a pr to address the overwritten style issue - note the override issue also affected the y scrollbar as well.

One thing to note: for barX and barY there are a few styles that are there for UI consistency, thus in my fix any overrides for those will be ignored. The good folks at primefaces will have to weigh in if they feel that can be adjusted cc @melloware 

The following styles for "barX" can not be overridden:
 - width
 - left
 - bottom

 The following styles for "barY" can not be overridden:
 - height
 - top
 - right

PR #7247, opened 2024-09-25
fix: #6682

gcko:
### Defect Fixes
fix: #6682
- Ensure element styles set via pt merge with required `xBarRef` and `yBarRef` styles

The following styles for "barX" can not be overridden:
 - width
 - left
 - bottom

 The following styles for "barY" can not be overridden:
 - height
 - top
 - right

### DIFF
--- a/components/lib/utils/DomHandler.js
+++ b/components/lib/utils/DomHandler.js
@@ -1055,9 +1055,9 @@ export default class DomHandler {
 
     static applyStyle(element, style) {
         if (typeof style === 'string') {
-            element.style.cssText = this.style;
+            element.style.cssText = style;
         } else {
-            for (let prop in this.style) {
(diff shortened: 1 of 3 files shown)


components/lib/utils/DomHandler.js
```diff
@@ -1055,9 +1055,9 @@ export default class DomHandler {
 
     static applyStyle(element, style) {
         if (typeof style === 'string') {
-            element.style.cssText = this.style;
+            element.style.cssText = style;
```
gcko, line 1058: @melloware I updated `DomHandler.applyStyle` to correctly "apply" the passed in `style` parameter.

melloware (approved):

melloware (approved):

PR #7271, opened 2024-09-28
fix: remove a div attribute type from TreeTable's prop type

### LINKED ISSUE #7269: TreeTable: size prop not working, even in official StackBlitz example
I'm trying to use the size prop in PrimeReact's TreeTable component, but it's not having any effect. The TypeScript type indicates that the size prop should accept a number, but when I pass a value like size={10}, there's no noticeable difference in the table.

I even tried this in the official [Prime React TreeTable Basic example][1] example on StackBlitz, using the basic usage demo, but still no effect. Here's my code based on that example:
```
import { TreeTable } from 'primereact/treetable';
import { Column } from 'primereact/column';
export default function BasicDemo() {
  const [nodes, setNodes] = useState([]);

  useEffect(() => {
    NodeService.getTreeTableNodes().then((data) => setNodes(data));
  }, []);

  return (
    <div className="card">
      <TreeTable size={10} value={nodes} tableStyle={{ minWidth: '100rem' }}>
        <Column field="name" header="Name" expander></Column>
        <Column field="size" header="Size"></Column>
        <Column field="type" header="Type"></Column>
      </TreeTable>
    </div>
  );
}

```

  [1]: https://primereact.org/treetable/

blan19:
### Defect Fixes
fix #7269 and reflects the div attributes not being applied

### Why fix?

In the type declaration, it was stated that div attrs were allow, but these were not actually applied in the TreeTable component.

```typescript
/**
 * Defines valid properties in TreeTable component. In addition to these, all properties of HTMLDivElement can be used in this component.
 * @group Properties
 */
export interface TreeTableProps extends Omit<React.DetailedHTMLProps<React.InputHTMLAttributes<HTMLDivElement>, HTMLDivElement>, 'onContextMenu' | 'onSelect' | 'ref' | 'value'> { ... }
```

in the rootProps, div attrs are excluded.
For example, attrs like 'hidden', 'size', etc.

```typescript
const rootProps = mergeProps(
  {
    role: 'table',
    id: props.id,
    className: classNames(props.className, ptCallbacks.cx('root', { isRowSelectionMode })),
    style: props.style,
    'data-scrollselectors': '.p-treetable-wrapper'
  },
  ObjectUtils.findDiffKeys(props, TreeTable.defaultProps),
  ptCallbacks.ptm('root')
);
```

i have fix the type declaration for TreeTable's Props to only expose the props allowed in the official documentation, similar to the Props type declaration for Tree.

### DIFF
--- a/components/lib/treetable/treetable.d.ts
+++ b/components/lib/treetable/treetable.d.ts
@@ -577,10 +577,10 @@ interface TreeTableColReorderEvent {
 }
 
 /**
- * Defines valid properties in TreeTable component. In addition to these, all properties of HTMLDivElement can be used in this component.
+ * Defines valid properties in TreeTable component.
  * @group Properties
  */
-export interface TreeTableProps extends Omit<React.DetailedHTMLProps<React.InputHTMLAttributes<HTMLDivElement>, HTMLDivElement>, 'onContextMenu' | 'onSelect' | 'ref' | 'value'> {
+export interface TreeTableProps {
     /**
      * Whether to show it even there is only one page.
      * @defaultValue true


melloware:
@blan19 check out my PR: https://github.com/primefaces/primereact/pull/7273

I made it consistent with what Datatable is doing and it had size ignored.  It also handled default props correctly.

blan19:
@melloware 

maintaining consistency with DataTable seems like a really good. I'll close my PR since it was the wrong direction xD

PR #7288, opened 2024-10-01
Tooltip Both Fix with Dynamic State

### LINKED ISSUE #7260: Tooltip event="both" does not work if element get focus state by clicking on it
### Describe the bug

If I se "tab" case, it perfectly fowks for hover or focus, but if I try click on element, tooltip will dissapier, when i leave trigger element with cursor viz picture.

### Reproducer

_No response_

### System Information

```Shell
System: Windows 10, Browsers: all npmPackages: primereact

primereact: 10.8.3
react: 18.2.0
```

tooltip

### Expected behavior

When user set event="both", it must show tooltip on hover and when user  focus input, it must be still visible, if user leave input / whatever with curssor.

ddoemonn:
fixes #7260

### before
https://github.com/user-attachments/assets/8a310ea0-3d5a-4a53-9e9c-cca973e83ab7

### after 
https://github.com/user-attachments/assets/9eb1e6dd-384b-4152-a6ab-9fac10ce4599

### DIFF
--- a/components/lib/tooltip/Tooltip.js
+++ b/components/lib/tooltip/Tooltip.js
@@ -29,6 +29,8 @@ export const Tooltip = React.memo(
             }
         };
 
+        const [isBothFocus, setIsBothFocus] = React.useState(false);
+
         const { ptm, cx, sx, isUnstyled } = TooltipBase.setMetaData(metaData);
 
         useHandleStyle(TooltipBase.css.styles, isUnstyled, { name: 'tooltip' });
@@ -110,7 +112,7 @@ export const Tooltip = React.memo(
 
                 if (event === 'both') {
                     showEvents = ['focus', 'mouseenter'];
-                    hideEvents = ['blur', 'mouseleave'];
+                    hideEvents = isBothFocus ? ['blur'] : ['mouseleave', 'blur'];
                 }
             }
 
@@ -172,6 +174,8 @@ export const Tooltip = React.memo(
         };
 
         const show = (e) => {
+            if(e.type && e.type==="focus") setIsBothFocus(true);
+          
             currentTargetRef.current = e.currentTarget;
             const disabled = isDisabled(currentTargetRef.current);
             const empty = isContentEmpty(isShowOnDisabled(currentTargetRef.current) && disabled ? currentTargetRef.current.firstChild : currentTargetRef.current);
@@ -198,6 +202,8 @@ export const Tooltip = React.memo(
         };
 
         const hide = (e) => {
+            if(e && e.type==="blur") setIsBothFocus(false);
+    
             clearTimeouts();
 
             if (visibleState) {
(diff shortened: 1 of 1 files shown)


components/lib/tooltip/Tooltip.js
```diff
@@ -29,6 +29,8 @@ export const Tooltip = React.memo(
             }
         };
 
+        const [isBothFocus, setIsBothFocus] = React.useState(false);
```
melloware, line 32: Move this line up to where the other `useState` values are so they are together. and call it `multipleFocusEvents`

melloware (changes requested):

components/lib/tooltip/Tooltip.js
ddoemonn: Okay, thanks for review!

melloware (approved):

components/lib/tooltip/Tooltip.js
```diff
@@ -14,6 +14,7 @@ export const Tooltip = React.memo(
         const [visibleState, setVisibleState] = React.useState(false);
         const [positionState, setPositionState] = React.useState(props.position || 'right');
         const [classNameState, setClassNameState] = React.useState('');
+        const [multipleFocusEvents, setMultipleFocusEvents] = React.useState(false);
```
melloware, line 17: Actually you can switch this to a `ref` it doesn't need to be `useState` as its not used in any hooks you are just tracking a boolean state.

melloware (changes requested):

components/lib/tooltip/Tooltip.js
ddoemonn: I can update it but there is something I didn't understand help me if I'm missing something. I put boolean inside of state because of trigger a value with using `ref` how can I achieve that?

melloware: you jiust do `yourRef.current = true` or `false`

ddoemonn: Yeap, I did the same but like I said we should trigger the component. You can try it in your local or I can commit.

melloware: interesting i was not expecting that based on looking at this code.  You are just using a boolean flag.  Changing the state is not used by any hook?

ddoemonn: I think the issue is related about refs in React don't cause re-renders that's why when we use `state` instead of `ref` we can always work with changed and dynamic value.

melloware: Ahhh OK so you need the re-render. Then I am OK how it is.

ddoemonn: Thanks!

ddoemonn: @melloware you merged but there is prettier error what should I do?

melloware: Oh I can fix it. Next time run `npm run format` on the codebase it fixes it

Luko248: When will be the release date?

Issue #7285, opened 2024-10-01
Dropdown: Blank Group Label


marengga:
### Describe the bug

Group label will be blank if OptionGroupLabel is set other than `label`.

After some debugging, I found that actual group label can be accessed inside `optionGroup` object.

```
<Dropdown 
  ...
  optionGroupLabel="name"
  ...
/>
```
Here's the issue, it'll be blank if we set other than "label"

```
<Dropdown 
  ...
  optionGroupLabel="optionGroup.name"
  ...
/>
```
It works, but is it intentional that it should be "optionGroup.[attributeName]" ?

### Reproducer

https://stackblitz.com/edit/stackblitz-starters-bgqusw?file=app%2Fpage.tsx

### System Information

```Shell
Binaries:
    Node: 21.2.0 - /usr/local/bin/node
    Yarn: 1.22.22 - /usr/local/bin/yarn
    npm: 10.2.4 - /usr/local/bin/npm
    Watchman: 2023.11.20.00 - /usr/local/bin/watchman
  npmPackages:
    primereact: ^10.8.3 => 10.8.3 
    react: ^18 => 18.2.0 
    tailwindcss: ^3.3.0 => 3.4.0
```

### Expected behavior

As documented, putting attribute name on `optionGroupLabel` should be enough

melloware:
Agreed you should not need `optionGroup.[key]` to access it.  PR is welcome if you want to look into how to fix it.

PR #7296, opened 2024-10-02
fix: flatOptions on Dropdown causing option group label blank

marengga:
### Defect Fixes

- fix [Dropdown: OptionGroup using optionGroupLabel is blank #7285 ](https://github.com/primefaces/primereact/issues/7285)

### Changes

- Merged optionGroup object to its parent (option)
- To determine if an option is a group, it is sufficient to use the `group` and `optionGroupLabel` properties

#### Before
image

#### After
image

### DIFF
--- a/components/lib/dropdown/Dropdown.js
+++ b/components/lib/dropdown/Dropdown.js
@@ -846,7 +846,7 @@ export const Dropdown = React.memo(
         };
 
         const isOptionGroup = (option) => {
-            return props.optionGroupLabel && option.optionGroup && option.group;
+            return props.optionGroupLabel && option.group;
         };
 
         const isOptionDisabled = (option) => {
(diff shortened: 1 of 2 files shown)


dgraham-nectivio:
@melloware When merging breaking changes, especially in a patch release, can we please add something to the release notes about it.

For those that are getting an error about `optionGroup` being `undefined` now, you need to change your `optionGroupTemplate` method.

Old code
```
    const optionGroupTemplate = (options: { optionGroup: { label: string } } ) => {
        return options.optionGroup.label
    }
````

Change to
```
    const optionGroupTemplate = ( optionGroup: { label: string } ) => {
        return optionGroup.label
    }
````

melloware:
@dgraham-nectivio i don't control the release notes and I don't work for PrimeTek. Just an open source contributor.

PR #7300, opened 2024-10-02
fix(steps): make active step controlled via props

### LINKED ISSUE #7298: Steps component: Unable to control the active step using activeIndex prop
### Describe the bug

Steps cannot be controlled programmatically using activeIndex, and this also doesn't work on the primereact documentation page: https://primereact.org/steps/#controlled

### System Information

```Shell
"primereact": "10.8.3",
"react": "18.2.0"
```

Click the buttons to set an activeIndex of the Steps component and observe the active step not changing as expected.

### Expected behavior

Clicking the buttons should change the active index.

ddoemonn:
fixes #7298

### before
https://github.com/user-attachments/assets/a58bf855-a861-4fd3-b49e-eb3e43e1aece

### after
https://github.com/user-attachments/assets/268a39a5-0574-4346-bdd7-cbb138bc555a

We shouldn't use `useState` for dynamic change here we need to use `props.activeIndex` directly.

Also, I checked It doesn't broke other step examples.

### DIFF
--- a/components/lib/steps/Steps.js
+++ b/components/lib/steps/Steps.js
@@ -12,7 +12,6 @@ export const Steps = React.memo(
         const props = StepsBase.getProps(inProps, context);
 
         const [idState, setIdState] = React.useState(props.id);
-        const [activeIndexState, setActiveIndexState] = React.useState(props.activeIndex);
         const elementRef = React.useRef(null);
         const listRef = React.useRef(null);
         const count = React.Children.count(props.children);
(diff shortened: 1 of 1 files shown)


melloware:
Did you test it in Controlled AND uncontrolled mode.  Typically PR stored that in a `useState` and has checks that check if `props.onStep` or somethig is ing Contolled vs Uncontrolled mode.

ddoemonn:
Yes @melloware I checked! Like I said because of using state here `activeIndex` doesn't work because in React when you sent props to a state it will use only in first render.

https://github.com/user-attachments/assets/ee609f02-4a7f-4246-9ed7-7aaceb3e40e4

melloware (approved):

Issue #7330, opened 2024-10-11
AutoComplete: Selected item is not getting highlighted when visited again unlike Dropdown component


akshayaqburst:
### Describe the bug

AutoComplete: Selected item is not getting highlighted when visited again unlike Dropdown component.

When we check dropdown component, the selected option gets highlighted in blue color as shown in the screenshot.
<img width="913" alt="image" src="https://github.com/user-attachments/assets/d9c01c6f-71e2-4d0d-ac5a-dc1f2fc2fe9f">

Similarly when we check Autocomplete dropdown - https://primereact.org/autocomplete/#dropdown - It does not highlight the selected value 
<img width="912" alt="image" src="https://github.com/user-attachments/assets/80a704ca-3bbc-4c58-8b9b-5f888849ea22">

### System Information

```Shell
npx: installed 1 in 0.852s

  System:
    OS: Linux 5.15 Ubuntu 20.04.6 LTS (Focal Fossa)
    CPU: (8) x64 11th Gen Intel(R) Core(TM) i7-1165G7 @ 2.80GHz
    Memory: 293.16 MB / 7.47 GB
    Container: Yes
    Shell: 5.0.17 - /bin/bash
  Binaries:
    Node: 14.21.3 - ~/.nvm/versions/node/v14.21.3/bin/node
    Yarn: 1.22.22 - ~/.nvm/versions/node/v14.21.3/bin/yarn
    npm: 6.14.18 - ~/.nvm/versions/node/v14.21.3/bin/npm
  npmPackages:
    react: 18.3.1 => 18.3.1
```

### Steps to reproduce the behavior

1. Go to https://primereact.org/autocomplete/#dropdown.
2. Select an option from dropdown.
3. Reopen the options list again from dropdown.
4. The previously selected option is not highlighted.

### Expected behavior

The previously selected option should highlight similar to dropdown component.

bloodykheeng:
hello @melloware  this issue  wasnt solved well still exists i think autocompletes need to get a datakey prop like how multi selects do
when i select an item here it doesnt get highlighted
i think the issue arrises with shallow comparison ie == vs ===

form schema
```

// ✅ Validation Schema
const formSchema = z
  .object({
    name: z.string().min(3, "Name must be at least 3 characters"),
    status: z.string().min(1, "Status is required"),
    description: z.string().optional().nullable(),
    bill_of_right: z
      .object({
        id: z.number().min(1, "Bill of Right ID is required"),
[...]

PR #7331, opened 2024-10-11
fix: #7330, AutoComplete: Selected item is not getting highlighted when visited again unlike Dropdown component

akshayaqburst:
### Defect Fixes
fix: #7330, AutoComplete: Selected item is not getting highlighted when visited again unlike Dropdown component

### DIFF
--- a/components/lib/autocomplete/AutoCompleteBase.js
+++ b/components/lib/autocomplete/AutoCompleteBase.js
@@ -34,7 +34,8 @@ const classes = {
     listWrapper: 'p-autocomplete-items-wrapper',
     list: ({ virtualScrollerOptions, options }) => (virtualScrollerOptions ? classNames('p-autocomplete-items', options.className) : 'p-autocomplete-items'),
     emptyMessage: 'p-autocomplete-item',
-    item: ({ suggestion, optionGroupLabel }) => (optionGroupLabel ? classNames('p-autocomplete-item', { 'p-disabled': suggestion.disabled }) : classNames('p-autocomplete-item', { 'p-disabled': suggestion.disabled })),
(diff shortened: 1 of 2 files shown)


components/lib/autocomplete/AutoCompleteBase.js
```diff
@@ -84,6 +85,10 @@ const styles = `
         padding: 0;
         list-style-type: none;
     }
+
+    .selected {
```
melloware, line 89: Selected is weird should it be `p-selected` or is this what the other component does?

melloware (changes requested):

components/lib/autocomplete/AutoCompleteBase.js
akshayaqburst: I can change it.

akshayaqburst: Changed to p-highlight similar to dropdown component.

components/lib/autocomplete/AutoCompleteBase.js
```diff
@@ -86,7 +86,7 @@ const styles = `
         list-style-type: none;
     }
 
-    .selected {
+    .p-highlight {
         background-color: rgba(6, 182, 212, 0.24) !important;
```
melloware, line 90: Should it be `!important` ?

melloware (changes requested):

components/lib/autocomplete/AutoCompleteBase.js
akshayaqburst: Removed the style added for p-highlight as a default style was already applied for autocomplete p-highlight class, and its working as expected.

PR #7352, opened 2024-10-17
Fix: CSS Modules documentation (:global selector)

### LINKED ISSUE #7347: Documentation for CSS Modules: Cascaded classes
### Describe the bug

Hi!

Your current documentation states:
> Unfortunately CSS modules do not support cascaded classes to be applied to external components however attribute selectors can be used as a common workaround

And then, in the example:
```css
/* paneldemo.module.css */
.mypanel > [class~="p-panel-header"] {
    background-color: #07c4e8;
    color: #ffffff;
}
```

However, you can use the :global selector. You can check the CSS Modules documentation here: [Composition -> Exceptions](https://github.com/css-modules/css-modules/blob/master/docs/composition.md#exceptions).

So, the example above should be:
```css
/* paneldemo.module.css */
.mypanel :global(.p-panel-header) {
    background-color: #07c4e8;
    color: #ffffff;
}
```

This is an official working solution for cascading classes in CSS Modules.

I think you should review this from your side and update your documentation accordingly.

Thanks for your work!

### Reproducer

_No response_

### System Information

```Shell
Any system
```

On any project with CSS Modules use :global selector instead attribute selector.

### Expected behavior

_No response_

StopNGo:
Fix #7347

### DIFF
--- a/components/doc/theming/scopedstyling/cssmodulesdoc.js
+++ b/components/doc/theming/scopedstyling/cssmodulesdoc.js
@@ -37,8 +37,9 @@ export default function PanelDemo() {
         <>
             <DocSectionText {...props}>
                 <p>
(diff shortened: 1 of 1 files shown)


melloware (approved):

melloware:
@StopNGo can you run `npm run format` and then commit it just a formatting issue.

PR #7355, opened 2024-10-17
Fix #7354: InputMask Classname p-filled not works fine

### LINKED ISSUE #7354: InputMask: Value not propagated
### Describe the bug

Value of inputmask is not propagated to inputtext component.
And if you set a value in inputmask and then disable/enable it, classname 'p-filled' is not set properly, causing float label working bad.

### Reproducer

https://stackblitz.com/edit/joufix?file=src%2FApp.tsx

### System Information

```Shell
System:
    OS: Windows 11 10.0.22631
    CPU: (20) x64 12th Gen Intel(R) Core(TM) i9-12900HK
    Memory: 34.26 GB / 63.71 GB
  Binaries:
    Node: 22.4.1 - ~\AppData\Local\nvs\default\node.EXE
    npm: 10.8.1 - ~\AppData\Local\nvs\default\npm.CMD
    pnpm: 9.6.0 - ~\AppData\Local\pnpm\pnpm.CMD
  Browsers:
  npmPackages:
    tailwindcss: 3.4.13 => 3.4.13
```

1. Go to reproducer
2. Toggle button "Click to see bug"

### Expected behavior

Classname 'p-filled' is properly set and float label should work fine.

ivanpajon:
### Defect Fixes
Fix #7354

### DIFF
--- a/components/lib/inputtext/InputText.js
+++ b/components/lib/inputtext/InputText.js
@@ -70,6 +70,14 @@ export const InputText = React.memo(
         const isFilled = React.useMemo(() => ObjectUtils.isNotEmpty(props.value) || ObjectUtils.isNotEmpty(props.defaultValue), [props.value, props.defaultValue]);
         const hasTooltip = ObjectUtils.isNotEmpty(props.tooltip);
 
+        React.useEffect(() => {
+            if (isFilled) {
+                DomHandler.addClass(elementRef.current, 'p-filled');
+            } else {
+                DomHandler.removeClass(elementRef.current, 'p-filled');
+            }
+        }, [props.disabled, isFilled]);
+
         const rootProps = mergeProps(
             {
                 className: classNames(props.className, cx('root', { context, isFilled })),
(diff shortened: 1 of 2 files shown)


components/lib/inputtext/InputText.js
```diff
@@ -70,6 +70,14 @@ export const InputText = React.memo(
         const isFilled = React.useMemo(() => ObjectUtils.isNotEmpty(props.value) || ObjectUtils.isNotEmpty(props.defaultValue), [props.value, props.defaultValue]);
         const hasTooltip = ObjectUtils.isNotEmpty(props.tooltip);
 
+        React.useEffect(() => {
+            if (isFilled) {
+                DomHandler.addClass(elementRef.current, 'p-filled');
+            } else {
+                DomHandler.removeClass(elementRef.current, 'p-filled');
+            }
+        }, [props.disabled, isFilled]);
```
melloware, line 79: You use props.disabled here to trigger the hook but I don't see how that is used in the hook?

ivanpajon: props.disabled was included in useEffect dependencies because each time the component is enabled/disabled it needs to run that logic

Issue #7358, opened 2024-10-19
InputNumber: InputNumber with horizontal and vertical buttons are not accessible through keyboard.


akshayaqburst:
### Describe the bug

InputNumber: InputNumber with horizontal and vertical buttons are not accessible through keyboard.

### System Information

```Shell
System:
    OS: Linux 5.15 Ubuntu 20.04.6 LTS (Focal Fossa)
    CPU: (8) x64 11th Gen Intel(R) Core(TM) i7-1165G7 @ 2.80GHz
    Memory: 1.61 GB / 7.47 GB
    Container: Yes
    Shell: 5.0.17 - /bin/bash
  Binaries:
    Node: 18.20.2 - ~/.nvm/versions/node/v18.20.2/bin/node
    Yarn: 1.22.22 - ~/.nvm/versions/node/v18.20.2/bin/yarn
    npm: 10.5.0 - ~/.nvm/versions/node/v18.20.2/bin/npm
    pnpm: 9.5.0 - ~/.nvm/versions/node/v18.20.2/bin/pnpm
  npmPackages:
    react: 18.3.1 => 18.3.1
```

### Steps to reproduce the behavior

1. Go to https://primereact.org/inputnumber/#buttons
2. Type something in the input box.
3. Try navigating through the buttons using tab key.
4. The focus goes to next section without going to increment and decrement buttons.

### Expected behavior

The focus should go to increment and decrement buttons, and user should be able to press enter and update values.

akshayaqburst:
> PrimeVue is the same way so I assume its intentional: https://primevue.org/inputnumber/#buttons
> 
> cc @tugcekucukoglu @mertsincan @cagataycivici

A person not using a mouse will not be able to increment or decrement the values using buttons. I find that a bad User experience.

tugcekucukoglu:
When the focus is on the input, the value will be updated according to the step prop using the arrow keys. We think that the InputNumber component should be exited using the tab key while the focus is still on the input. Otherwise, it would be necessary to press the tab key three times to exit.

PR #7359, opened 2024-10-19
fix: #7358, InputNumber: InputNumber with horizontal and vertical buttons are not accessible through keyboard.

akshayaqburst:
fix: #7358, InputNumber: InputNumber with horizontal and vertical buttons are not accessible through keyboard.


melloware:
Looks like this was intentional as PrimeVue does the same thing: https://github.com/primefaces/primevue/blob/69d0407fe6aa617bfd0f0d4cb5119d3446c1bf4d/packages/primevue/src/inputnumber/InputNumber.vue#L33

PR #7360, opened 2024-10-19
Fix for Issue 7329

### LINKED ISSUE #7329: Row component does not receive className and style property
### Describe the bug

Hello,

it seems to me that this issue from 2 years ago is present again: https://github.com/primefaces/primereact/issues/2693
I forked the CodeSandbox from the issue, updated primereact to the latest version and added classNames to both a Row and a Column component. While the column component sets the class, the row class stays unset. See Reproducer for an example. I was also able to reproduce it in another local project with a TreeTable as well.

### Reproducer

https://codesandbox.io/p/sandbox/primereact-test-forked-7mt96v

### System Information

```Shell
npmPackages:
  primereact: ^10.8.3 => 10.8.3
  react: ^18.2.0 => 18.2.0
```

Using a Row component with setting style or className results in a < tr > without style or className.

### Expected behavior

The resulting < tr > component should get the style or the class set.

Nickog13:
Fix #7329 

This pull request should fix issue #7329 by adding RowBase.getProps to the mergeProps call.
I have also added changes due to two build issues that I have encountered:
* I also have adjusted few reassignments that are made to constant variables, which failed the build.
* For a few ts interfaces I got build errors like this: "Type 'ToggleEvent<HTMLFieldSetElement> is missing the following properties from type 'FieldSetToggleEvent': originalEvent, value" - Are those currently present in the master branch code base? I got rid of those build errors only for now by adding onToggle to Omit. As I am not very familiar with the prime react code base, I was not sure how to fix this correctly. Would appreciate any hint/help how to fix this in the correct way.


melloware:
Weird I am not seeing the OnToggle errors you are seeing we do a ts-check on the build of every PR?

melloware:
For example the build runs `npm run type:check`

Nickog13:
That's weird then: When I clone the current repository and execute the command from above, I am instantly getting the error

melloware:
let me check

melloware:
```
C:\dev\primetek\primereact>npm run type:check

> primereact@10.8.4 type:check
> tsc -p ./api-scripts/tsconfig.json


C:\dev\primetek\primereact>
```

That is what happens when I run from Main right now?

Nickog13:
I think I had a version mismatch with next. I had already installed next 14.2.15 before cloning and my IDE changed the package.json versions without letting me know...
Sorry for the confusion. When reverting back to the version used in main 12.3.4, I don't have the errors.

melloware:
OK can you revert all those changes then? I think this PR gets a lot smaller 😄

melloware (changes requested):
Revert all unnecessary changes

melloware (approved):

Issue #7377, opened 2024-10-31
MultiSelect: Cannot access the close Icon and clear the input using keyboard in Filled mode


akshayaqburst:
### Describe the bug

MultiSelect: Cannot access the close Icon and clear the input using keyboard in Filled mode

### System Information

```Shell
System:
    OS: Linux 5.15 Ubuntu 20.04.6 LTS (Focal Fossa)
    CPU: (8) x64 11th Gen Intel(R) Core(TM) i7-1165G7 @ 2.80GHz
    Memory: 2.16 GB / 7.47 GB
    Container: Yes
    Shell: 5.0.17 - /bin/bash
  Binaries:
    Node: 18.20.2 - ~/.nvm/versions/node/v18.20.2/bin/node
    Yarn: 1.22.22 - ~/.nvm/versions/node/v18.20.2/bin/yarn
    npm: 10.5.0 - ~/.nvm/versions/node/v18.20.2/bin/npm
    pnpm: 9.5.0 - ~/.nvm/versions/node/v18.20.2/bin/pnpm
  npmPackages:
    react: 18.3.1 => 18.3.1
```

### Steps to reproduce the behavior

1. Go to https://primereact.org/multiselect/#filled
2. Select a few items from dropdown and notice the clear Icon appears.
3. Try to access the clear icon using keyboard.
Clear Icon not focussing.

### Expected behavior

Clear icon should be focussable and on pressing enter key, the chosen options should be cleared.

PR #7378, opened 2024-10-31
fix: #7377, MultiSelect: Cannot access the close Icon and clear the input using keyboard in Filled mode

akshayaqburst:
fix: #7377, MultiSelect: Cannot access the close Icon and clear the input using keyboard in Filled mode

### DIFF
--- a/components/lib/multiselect/MultiSelect.js
+++ b/components/lib/multiselect/MultiSelect.js
@@ -981,11 +981,28 @@ export const MultiSelect = React.memo(
             ZIndexUtils.clear(overlayRef.current);
         });
 
+        const onClearIconKeyDown = (event) => {
+            switch (event.code) {
+                case 'Space':
+                case 'NumpadEnter':
+                case 'Enter':
+                    if (props.inline) {
+                        break;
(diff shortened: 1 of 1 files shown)


components/lib/multiselect/MultiSelect.js
```diff
@@ -981,11 +981,28 @@ export const MultiSelect = React.memo(
...
+        const onClearIconKeyDown = (event) => {
+            switch (event.code) {
+                case 'Space':
+                case 'NumpadEnter':
+                case 'Enter':
+                    if (props.inline) {
+                        break;
+                    }
+
+                    updateModel(event, [], []);
```
melloware, line 993: i think you want to add `event.preventDefault()` and `event.stopPropagation()` so this does not bubble up to other components

melloware (changes requested):

components/lib/multiselect/MultiSelect.js
akshayaqburst: Done @melloware

melloware (approved):

PR #7404, opened 2024-11-10
Fix #7333: Picklist: When Items are moved to source or target, they are being still selected.

### LINKED ISSUE #7333: Picklist: When Items are moved to source or target, they are being still selected.
### Describe the bug

When Items in picklist are selected from whether source or target and moved to the other list, remain still selected causing bad UX.
For example, here I moved `Activity Code` to the target List, it still remains selected, if I choose another item from the target, i.e. classes and clicked move to source, both are moved
Note: I can't upgrade the used version.
image

### Reproducer

_No response_

### System Information

```Shell
primereact: 10.6.6
react: ^18.3.1
```

You can run the example here, try to move any of the items from the source to target, then select new item from the target list, the item that was being recently moved still selected.
https://stackblitz.com/edit/vitejs-vite-hcngdj?file=package.json

_No response_

### Expected behavior

_No response_

jayaphanindra686:
### Defect Fixes
Fix #7333

### DIFF
--- a/components/lib/picklist/PickList.js
+++ b/components/lib/picklist/PickList.js
@@ -114,6 +114,7 @@ export const PickList = React.memo(
                         });
                     }
 
+                    onSelectionChange({ originalEvent, value: selectedValue }, 'targetSelection', props.onTargetSelectionChange);
                     break;
 
                 case 'allToTarget':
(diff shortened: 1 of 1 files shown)


melloware (approved):

melloware:
I made an improved fix here:https://github.com/primefaces/primereact/pull/7571

PR #7417, opened 2024-11-14
Fixing cssTransitionfalse does not disable animations globally

### LINKED ISSUE #7395: PrimeReactProvider: cssTransition false does not disable animations globally
### Describe the bug

When setting `cssTransition` to `false` in `PrimeReactProvider`, animations are not disabled globally:

````
<PrimeReactProvider value={{ cssTransition: false }}>
    <App />
</PrimeReactProvider>
````

Setting `transitionOptions={{ disabled: true }}` on the Component itself works though.

### Reproducer

https://stackblitz.com/edit/cc98j5-u72fsi?file=src%2FApp.jsx

### System Information

```Shell
...
```

`<PrimeReactProvider value={{ cssTransition: false }}>`

### Expected behavior

Disabled animations on a global level

tneeraja95:
### Defect Fixes
When submitting a PR, please also <ins>**create an issue**</ins> documenting the error and [manually link to an issue](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#manually-linking-a-pull-request-or-branch-to-an-issue-using-the-issue-sidebar) or mention it in the description using #<issue_id>.

Fix #7395

Description - 
Problem is in file `PrimeReactContext.js`, line 15

` const [cssTransition, setCssTransition] = useState(propsValue.cssTransition || true);`

|| is OR, which basically means that even when cssTransition from props is `false`, `false || true` return `true`. 

Instead of OR what we really want is 'Nullish coalescing operator (??)', which returns its right-hand side operand when its left-hand side operand is [null](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/null) or [undefined](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/undefined), and otherwise returns its left-hand side.

Updating the same for other props as well.

### DIFF
--- a/components/lib/api/PrimeReactContext.js
+++ b/components/lib/api/PrimeReactContext.js
@@ -12,7 +12,7 @@ export const PrimeReactProvider = (props) => {
     const [locale, setLocale] = useState(propsValue.locale || 'en');
     const [appendTo, setAppendTo] = useState(propsValue.appendTo || null);
     const [styleContainer, setStyleContainer] = useState(propsValue.styleContainer || null);
-    const [cssTransition, setCssTransition] = useState(propsValue.cssTransition || true);
+    const [cssTransition, setCssTransition] = useState(propsValue.cssTransition ?? true);
     const [autoZIndex, setAutoZIndex] = useState(propsValue.autoZIndex || true);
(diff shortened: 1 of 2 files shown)


melloware:
Yeah I wonder if all those should be `??` as well?   Can you make that change?

tneeraja95:
Thanks @melloware. Doing the same for other props as well in the same PR.

PR #7420, opened 2024-11-15
Performance (datatable): avoiding unnecessary re-render of cells of rows with no state-change

### LINKED ISSUE #2979: DataTable: seems to re-render all the rows on select all, even with the pagination.
### Describe the bug

Hi, so I'm trying to implement a bulk edit on a DataTable with a large amount of entries. But the more entries the table contains, the more slower it is to display the select rows.
I did some tests with a lazy loading and it's much quicker without taking care of the entries number. But this way I lose the sorting and ordering feature of DataTable.

So is there a way to only re-render the visible rows of a DataTable with pagination?

Thanks!

### Reproducer

https://codesandbox.io/s/primereact-test-forked-vqhyjv?file=/src/index.js

### PrimeReact version

8.1.1

### React version

18.x

### Language

ALL

### Build / Runtime

Create React App (CRA)

### Browser(s)

FireFox 101, Chromium 102

1. Open the link
2. Click on the checkbox in the header
3. change the `ENTRY_MULTIPLIER`  value to see the render speed change.

### Expected behavior

Render only the visible rows of the DataTable when the checkbox in the header is clicked.

acc-cassio:
### Defect Fixes
Fix #2979: Datatable performance improved by taking out parameters and functions of the BodyCells and placing in the BodyRow or TableBody components. Also, the set of BodyCell props to be checked when deciding for a re-render in React.memo was improved. This avoids now a lot of unnecessary re-render, including the issue from the linked issue.

Note that I also created a function called "selectiveCompare" in the BodyCell component to make the selection of which props should be checked at the React.memo. Maybe it makes sense to move it to the Utils, since it potentially can be used for optimizing the datatable further and more generally other library components. This selective comparison of the props is important since React.memo will not work as expected mainly if there are objects being parsed in the props.


accioly-cassio:
There might still be a bug when the metakey is used for column sorting. I'll verify this later today or tomorrow.

components/lib/datatable/BodyCell.js
```diff
@@ -14,6 +14,67 @@ import { DomHandler, IconUtils, ObjectUtils, classNames } from '../utils/Utils';
 import { RowCheckbox } from './RowCheckbox';
 import { RowRadioButton } from './RowRadioButton';
 
+function getNestedValue(obj, path) {
```
melloware, line 17: These 3 functions probably can be moved to ObjectUtils since they are generic functions.

melloware (changes requested):
This is fantastic just one recommendation

components/lib/datatable/BodyCell.js
accioly-cassio: Done :). Also verified that there is no bug related to the metakey when sorting.

melloware (approved):

tumngust2:
Any  chance TreeTable also getting this improvements?

melloware:
@acc-cassio can you fix your merge conflicts?  Sorry i know this PR has been open a long time.

acc-cassio:
Thank you for solving the conflict, @melloware. I just ran prettier and should be ready to merge now

melloware (approved):

melloware:
@acc-cassio your fix reintroduced a bug with the AlignFrozen column scenario: #7545

The scenarios is available to test in the Showcase if you run `npm run dev` and go to the FrozenColumn demo.

melloware:
What was happening before in BodyCell.js

```js
        React.useEffect(() => {
            if (getColumnProp('frozen')) {
                updateStickyPosition();
            }
        });
```

But now after your change I can see this hook not firing.

acc-cassio:
Hi @melloware, I will take a look at it this afternoon

melloware:
Thanks @acc-cassio

acc-cassio:
Fixed. I guess I have to create a new PR?

Falcowoski:
Hello, how are you?

It seems that _unfortunately_ this improvement created some bugs. 

The `body` property of the `Column` component is frozen _(Ironic, perhaps)_ and does not respond to any kind of change.

I understand that performance is something important and cannot be ignored, but at the same time, isn't it strange that we are breaking React's rules in this case? The `body` prop has changed, so the component should re-render accordingly.

_I'm not sure_ if this is a bug (the property is not in the dependencies array) or simply a performance decision.

_I noticed this same problem in some other properties during my investigation, but unfortunately I don't remember their names at the moment 😢_ 

https://stackblitz.com/edit/data-table-bug?file=src%2FApp.tsx

acc-cassio:
Hi Falcowoski,

If you include the body in the dependency array, you'd break the performance boost gained with this fix, since the Cells of all Columns containing a function in the body would get re-rendered at any state change of the datatable, even if the state changes would be unrelated to their respective Column bodies. This will happen because in JS the functions are compared by reference, so even if two functions are equal, the comparison would return 'false' because they are not the same 'objects'.

What you can do to reach the bevaviour you desire is to include the count in the row data, which, in the above example, are the products. The row data is checked in the memoization and, when some change is detected, triggers the re-render of the corresponding row.

See this applied to the example above: https://stackblitz.com/edit/data-table-bug-s6kc3zbd?file=src%2FApp.tsx

@melloware, if further issues like this one pop up, what could also be though of is to create a boolean parameter which would control the memoization checking. With it, you could choose between the new and the old re-rendering behaviours.

melloware:
@acc-cassio that is a great idea.  Anyone who reports a broken table can just flip the flag back.  Or what if you add a param that is  `rerender="body,frozen,alignFrozen"`  something where its not re-rednered by default but allow a user to override one property or a few properties?

acc-cassio:
Yeah, we could do that too. It'd be quite an advance feature to use though since the cell props are not necessarily the same as the datatable props, so you would need to be familiar with the BodyCell component. But at least the option would be there.

Let's wait to see if this is a more general rather than isolated issue (for which there is an easy fix, as shown).

melloware:
agreed. lets gather some more feedback or hear any more bugs before we make a decision on how to fix.

Falcowoski:
Not to be rude, but it doesn't make sense to add extra values (like states or props) inside the table data...

I understand that we're going to lose performance, but that's part of how React works. There's not much to avoid.

**It's not "common" for a React component to ignore changes to a prop**. It goes completely against how React works by default, to the point that I spent more than 15 minutes trying to understand what the hell was going on.

If the issue is performance when adding it to the array dependencies, and yes, **the function will change its reference with each render, just memorize it with useCallback.** It's an expected behavior, considering how React handles this kind of thing.

Regarding a "_rerender_" prop, it works, but it doesn't make much sense, honestly.

Again, I don't want to be rude, but in my own way I've been working with design systems for over two years, and I've also personally suffered a lot when dealing with tables and performance, so just like you, I just want to improve what we have here even more 🤝 

**Thanks for the quick response and fix suggestions. Fortunately, we're not on that version of the library yet, so it's not an issue at the moment.**

---

**In short: We shouldn't be running away from React, but rather solving problems in its style, even if it's not perfect.**

**It's cliché, I know, but I believe it has a lot to add to the final consumer experience.**

Falcowoski:
Oh, I'm really looking forward to version 11! I was checking out the _branch_ and even took the opportunity to test it locally! 

**It's looking really cool!**

I'm migrating from a private design system with **Radix UI** in a project and we've been bumping into these _"non-React"_ behaviors from time to time, but I saw that the team is changing this in v11 🙏 

We already have Figma for teams and we are waiting for the new theme in **PrimeReact**, so whatever I can do to help, let me know 🤝

melloware:
@Falcowoski i totally understand where you are coming from but if you look at the number of DT performance issue tickets and th e number of upvotes and people asking for this on Discord its easier said than done!

As for V11 I agree everyone is waiting to get their hands on it!

Falcowoski:
Yes, you brought up a great point, however, wouldn't it be more interesting and simpler to invert the logic here? If the user needs extra performance, he should configure this, that is, he should activate this freezing functionality and/or memoize some props himself. 

_Of course, this is assuming that the performance issue only happens in more advanced cases with a lot of data._ 

This way, the developer experience will be simple for everyday cases and complex cases can have an extra topic in the documentation talking about performance.

I believe this will reduce future issues or discussions about "strange behavior".

What do you think?

acc-cassio:
This a misunderstanding on how useCallback works. In this case, wrapping the prop function into it won't help much, since you'd have to put the prop function itself as a dependency of the useCallback. This would result in the wrapped function getting re-created and causing the re-renders anyway, since the comparison of the function given in the dependency array would always fail, as explained, since JS compared objects and functions by reference.

useCallback needs to be used always where the function is originally created. In this case, directly at the BasicDemo() component, so that only number, string, boolean, etc variables are in the useCallback dependency. Wrapping it inside the Datatable component itself would not do much to stabilize the function.

Of course, the main principle of React is to re-render any part of the dom when state changes. But these re-renders have to be restricted to what necessary, otherwise any sufficiently complex react app would be impracticle to use due to poor performance. Re-rendering all cells of a 10x1000 datatable when the staty only one cell changes is terrible. This PR corrected exactly that.

Another point: I am not part of the primereact team and do not win anything for helping here. I use this great lib for a personal project and am trying to give my contribution.

So far, I think the best solution is indeed the re-render prop, which, if used in combination with useCallback in the parent component (in the example, the BasicDemo component), would allow the user to achieve both the performance boost and control the component's reactivity.

Falcowoski:
First of all, thank you for your reply and for your work on this pull request. I understand where you're coming from and in this case, it seems like the best option 👍

Regarding useCallback, that wasn't exactly my suggestion, but I understand what you meant 🤝

Thanks!

melloware:
Also forgot to mention I also don't work for PrimeTek @acc-cassio and I are just people who like helping people!

melloware:
@omikader if you want to read @acc-cassio added 3 new useful props for controlling the behavior: https://github.com/primefaces/primereact/pull/8035

Aaron-Spence:
I am all for performance gains, but unfortunately 10.9.5 introduced breaking changes for us, which remained in 10.9.6. Now with 10.9.6 if I set cellMemo={false} the bugs disappear. If cellMemo={true} they return.

I absolutely think that cellMemo must default to false. Else melloware's dream of less complaints about performance will simply be replaced with complaints of strange behaviour! Instead, the DataTable docs could alert users to the potential performance increase provided by cellMemo={true}.

In brief, `onCellEditComplete()` upon entering then exiting an editable cell the _second_ time, without modifying the values at all, will have a `newValue` of null, and thus clear the cell's value. I have not tracked down why yet.

https://github.com/user-attachments/assets/3621d111-4806-474a-8456-328eff780ca0

acc-cassio:
Hi @Aaron-Spence, please create a new issue with a reproducer so we can help you tracking it down and eventually fix primereact's code if necessary.

PR #7431, opened 2024-11-18
Fix: Dropdowmitem: Passing key directly without using spread

### LINKED ISSUE #7430: DropdownItem: Error: "key" prop is being spread into JSX
### Describe the bug

Inside `DropdownItem`, which is a item in the dropdown list, so key is required prop. 
 
Screenshot 2024-11-18 at 8 01 17 PM

Inside `DropdownItem.js`, `key` is in `itemProps` which should be outside and directly in `<li>`

### Reproducer

https://stackblitz.com/edit/xqgdym?file=src%2FApp.jsx

### System Information

```Shell
react : "18.3.1"
primereact: "10.8.4"
```

1. Open inspect
2. Click on the dropdown
3. Can see error in console.

### Expected behavior

No error

tneeraja95:
### Defect Fixes

Fix #7430 : "key" prop is being spread into JSX

Solution : 
Passing key directly without using spread

### DIFF
--- a/components/lib/dropdown/DropdownItem.js
+++ b/components/lib/dropdown/DropdownItem.js
@@ -33,7 +33,6 @@ export const DropdownItem = React.memo((props) => {
         {
             id: `dropdownItem_${index}`,
             role: 'option',
-            key: props.label,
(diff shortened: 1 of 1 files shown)


melloware (approved):
Thanks

Issue #7405, opened 2024-11-10
Datatable: CellEdit and general issues with portal editors


gnawjaren:
### Describe the bug

There are a few issues with portal editors like dropdown and alike - caused by the onClickOutside and closing edit mode.
There were a few changes that adressed this, but some issues remain - like not being able to filter in dropdown and some other issues in other editors. I can give a repro, but i think the issue is well known.
For us it is critical to have such options, like dropdown filter.
#7158 #2666 #5320
#3079 #2097

I have tested a different approach to solve this:
```diff
diff --git a/components/lib/datatable/BodyCell.js b/components/lib/datatable/BodyCell.js
index 043d2160a..701156235 100644
--- a/components/lib/datatable/BodyCell.js
+++ b/components/lib/datatable/BodyCell.js
@@ -63,9 +63,17 @@ export const BodyCell = React.memo((props) => {
         return getColumnProp('cellEditValidateOnClose');
     };

+    const isIgnoredElement = (element) => {
+        const isCellEditor = (el) => el.getAttribute && el.getAttribute('data-p-is-cell-editor');
+        return isCellEditor(element) || DomHandler.getParents(element).find((el) => isCellEditor(el));
+    };
+
     const [bindDocumentClickListener, unbindDocumentClickListener] = useEventListener({
         type: 'click',
         listener: (e) => {
+            if (isIgnoredElement(e.target)) {
+                return
+            }
             setTimeout(() => {
                 if (!selfClick.current && isOutsideClicked(e.target)) {
                     // #2666 for overlay components and outside is clicked
```
This does give the option to control if you wish to ignore the overlay on outside click, i tested this local and it works fine - with primereact components and even third party. It gives you the option to add this attribute to the overlay (e.g. 
```js
pt={
  panel:{
    "data-p-is-cell-editor":true
  }
}
```
or just create an element on the root page, append the cell editors there and call it a day. 
I just wrote and tested this as a proof of concept, maybe there are better ways - but i think the approach to provide an option to ignore specific editors would be great for many use cases. And it doesnt break anything.
I can provide a pr if this approach seems ok.
Kind regards and have a good day!

### System Information

```Shell
primereact: 10.8.4
react: 18.3.1
```

### Steps to reproduce the behavior

Use an editor with portal in Datatable with editmode="cell" and try to interact with it.

### Expected behavior

Interaction with the component - not closing of the editor on click.

gnawjaren:
It seems to work fine in 10.8.5

[...]

Issue #2097, opened 2021-06-08
DataTable editor with 3rd party dropdown / overlay


jc-1234:
### There is no guarantee in receiving an immediate response in GitHub Issue Tracker, If you'd like to secure our response, you may consider *PrimeReact PRO Support* where support is provided within 4 business hours

**I'm submitting a ...**  (check one with "x")
```
[ x ] bug report
[ ] feature request
[ ] support request => Please do not submit support request here, instead see https://forum.primefaces.org/viewforum.php?f=57
```

**Codesandbox Case (Bug Reports)**
https://codesandbox.io/s/quirky-hill-w9ek1?file=/src/demo/DataTableEditDemo.tsx

**Current behavior**
When editor is a 3rd party or custom dropdown / overlay, clicking inside the dropdown or overlay will close the editor and turn cell back to view only.  However, using PrimeReact dropdown or overlay works fine.

**Expected behavior**
DataTable cell editor should support 3rd party or custom dropdown / overlay.  Or allow a way for developers to prevent closing of cell editor when clicking inside a custom dropdown or overlay.

**Minimal reproduction of the problem with instructions**
https://codesandbox.io/s/quirky-hill-w9ek1?file=/src/demo/DataTableEditDemo.tsx
1. Click on a cell in Date column
2. Try to switch month in the drop down calendar via the top right > button in the calendar
3. Drop down calendar closes right away and user can not switch months

**Please tell us about your environment:**
Windows 10, latest npm and node

* **React version:**
v17.x

* **PrimeReact version:**
v6.3.2

* **Browser:** [all | Chrome XX | Firefox XX | IE XX | Safari XX | Mobile Chrome XX | Android X.X Web Browser | iOS XX Safari | iOS XX UIWebView | iOS XX WKWebView ]
Chrome
 
* **Language:** [all | TypeScript X.X | ES6/7 | ES5]
Typescript

sdt-jchau:
any update on this?  Will this be fixed ever so PrimeReact works with other React components?

melloware:
I think this is related to this: https://github.com/primefaces/primereact/issues/2666

Anything clicked on that is not the target closes the editor as it thinks its "done"

habubey:
Issue tracker is used for defects only as part of our commitment to quality and continuous improvement in all areas. Enhancements are collected as valuable community feedback and managed internally so moving this enhancement ticket to our internal project management backlog.

PR #7439, opened 2024-11-20
Datatable: Allow ignore of data-tag in CellEdit outside click

gnawjaren:
### Defect Fixes
Overlay Editors behave different when in CellEdit mode. This PR adds a workaround - adding the attribute
`data-pr-is-overlay:true` to the overlay excludes it from the onClickOutside handler.
This *could* be extended by adding it to the overlay editors by default. 
We could then even remove the setTimeout.
It also works für third party editors. 
#7405 #2097


gnawjaren:
It seems to work fine in 10.8.5

Issue #7527, opened 2024-12-26
Dropdown: Editable Dropdown search not working as expected.


akshayaqburst:
### Describe the bug

Dropdown: Editable Dropdown search not working as expected.

In the video attached, when I search for 225, it highlights 225000 instead of 225. In case of a perfect match, I expect it to highlight that value, else it can search with other criterias.

https://github.com/user-attachments/assets/c8ff8235-ccce-4d70-a9c4-fa0213ed764c

### Reproducer

https://stackblitz.com/edit/yxs5khtq?file=src%2FApp.jsx

### System Information

```Shell
System:
    OS: Linux 5.15 Ubuntu 20.04.6 LTS (Focal Fossa)
    CPU: (8) x64 11th Gen Intel(R) Core(TM) i7-1165G7 @ 2.80GHz
    Memory: 4.10 GB / 7.47 GB
    Container: Yes
    Shell: 5.0.17 - /bin/bash
  Binaries:
    Node: 14.21.3 - ~/.nvm/versions/node/v14.21.3/bin/node
    Yarn: 1.22.22 - ~/.nvm/versions/node/v14.21.3/bin/yarn
    npm: 6.14.18 - ~/.nvm/versions/node/v14.21.3/bin/npm
  npmPackages:
    react: 18.3.1 => 18.3.1
```

### Steps to reproduce the behavior

1. Go to https://stackblitz.com/edit/yxs5khtq?file=src%2FApp.jsx
2. Search for 225.
3. '225000' is highlighted instead of 225.

### Expected behavior

Search should check for an exact match and else it should check for startsWith.

PR #7528, opened 2024-12-26
fix: #7527, Dropdown: Editable Dropdown search not working as expected

akshayaqburst:
fix: #7527, Dropdown: Editable Dropdown search not working as expected

https://github.com/primefaces/primereact/issues/7527

### DIFF
--- a/components/lib/dropdown/Dropdown.js
+++ b/components/lib/dropdown/Dropdown.js
@@ -588,12 +588,24 @@ export const Dropdown = React.memo(
             return isOptionDisabled(option) ? findPrevOption(i) : option;
         };
 
+        const findInArray = (visibleOptions, searchText) => {
+            if (!searchText || !visibleOptions?.length) return -1;
+
+            const normalizedSearch = searchText.toLocaleLowerCase();
+
+            const exactMatch = visibleOptions.findIndex((item) => getOptionLabel(item).toLocaleLowerCase() === normalizedSearch);
+
+            if (exactMatch !== -1) return exactMatch;
+
+            return visibleOptions.findIndex((item) => getOptionLabel(item).toLocaleLowerCase().startsWith(normalizedSearch));
+        };
+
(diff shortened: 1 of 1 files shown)


components/lib/dropdown/Dropdown.js
```diff
@@ -588,12 +588,24 @@ export const Dropdown = React.memo(
...
+            if (!searchText || !visibleOptions?.length) return -1;
+
+            const normalizedSearch = searchText.toLocaleLowerCase();
+
+            const exactMatch = visibleOptions.findIndex((item) => getOptionLabel(item).toLocaleLowerCase() === normalizedSearch);
+
+            if (exactMatch !== -1) return exactMatch;
+
+            return visibleOptions.findIndex((item) => getOptionLabel(item).toLocaleLowerCase().startsWith(normalizedSearch));
+        };
```
melloware, line 601: Won't this do it more efficientialy by only calling FindIndex 1 instead of passing over the whole array twice?

```js
const search = (visibleOptions, searchText) => {
    if (!searchText || !visibleOptions?.length) return -1;

    const normalizedSearch = searchText.toLocaleLowerCase();

    return visibleOptions.findIndex((item) => {
        const label = getOptionLabel(item).toLocaleLowerCase();
        return label === normalizedSearch || label.startsWith(normalizedSearch);
    });
};
```

I would also call the method `search` instead of `findInArray`

melloware (changes requested):

components/lib/dropdown/Dropdown.js
melloware: Oh no i see you want to search the whole list for an exact match before trying again with `startswith`

akshayaqburst: Right, we will need to iterate the array twice in this case. Once to find the exact match and then to find matches with startsWith.

melloware (approved):

PR #7554, opened 2025-01-03
fix:#7546#Datatable: Sort with multiselect - wrong selection

### LINKED ISSUE #7093: DataTables Multiselect Broken With Column Sorting
### Describe the bug

DataTable multi-select using shift+click is broken when sorting by a column. I sort by column B then click on element 0, shift_click down to element 5 and it doesn't select elements 0-5 it selects what I believe to be elements 0-5 of the DEFAULT sorting order, which to the end user looks almost like its randomly selecting elements.

https://github.com/user-attachments/assets/968657cc-7c74-496f-ab55-fdbf893ac826

### Reproducer

https://stackblitz.com/edit/vitejs-vite-nvgvfm?file=src%2FApp.jsx

### System Information

```Shell
Node: 18.20.3
    Yarn: 1.22.19
    npm: 10.2.3
    pnpm: 8.15.6

    primereact: 10.8.2 => 10.8.2 
    react: ^18.2.0 => 18.2.0
```

Sort by a column so that it is not in the default sorting order
Click an element
Shift+click another element
Note that it is not selecting the rows expected

### Expected behavior

Select the rows in the range selected based on the displayed row order.

### LINKED ISSUE #7546: Datatable: Sort with multiselect - wrong selection
### Describe the bug

Duplicate of issue #7468

Problem persists even after PR #7492 

### Reproducer

https://stackblitz.com/edit/vitejs-vite-lvxdbt

### System Information

```Shell
primereact 10.9.1
```

_No response_

### Expected behavior

_No response_

kl-nevermore:
Fix #7546 
Fix #7093
passing processed data should be the convenient solution

- `props.value`  only contains the current data (the virtual list only contains data in the visible area)
- `props.tableProps.value`  only contains raw data
- maybe `onRowDrop` have the same problem

https://github.com/user-attachments/assets/c1b527fa-ae63-48f9-808a-c92bcc5a9ca8

@sja-cslab

### DIFF
--- a/components/lib/datatable/DataTable.js
+++ b/components/lib/datatable/DataTable.js
@@ -1642,7 +1642,7 @@ export const DataTable = React.forwardRef((inProps, ref) => {
         );
     };
 
-    const createTableBody = (options, selectionModeInColumn, empty, isVirtualScrollerDisabled) => {
+    const createTableBody = (options, selectionModeInColumn, empty, isVirtualScrollerDisabled, processedData) => {
         const first = getFirst();
         const { rows, columns, contentRef, style, className, spacerStyle, itemSize } = options;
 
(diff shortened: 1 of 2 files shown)


sja-cslab:
Looks correct to me

melloware (approved):

Issue #6583, opened 2024-05-07
Calendar: Prop typings are not working correctly


JosuerBague:
### Describe the bug

The selectionMode prop shows a TS error regardless of value passed: 'single' | 'multiple' | 'range'.

### Reproducer
https://stackblitz.com/edit/vitejs-vite-1hvojf?file=src%2FApp.tsx

### PrimeReact version

10.6.5

### React version

18.x

### Language

TypeScript

### Build / Runtime

Vite

### Steps to reproduce the behavior

I'm creating a custom wrapper around the Calendar component, like so:

```
import { Calendar as PRCalendar, CalendarProps } from 'primereact/calendar';

type Omissions = 'disabled' | 'invalid' | 'required' | 'visible';

type Properties = {
  isDisabled?: boolean;
  isInvalid?: boolean;
  isRequired?: boolean;
  isVisible?:boolean;
} & Omit<CalendarProps, Omissions>

const Calendar: React.FC<Properties> = ({
  isDisabled,
  isInvalid,
  isRequired,
  isVisible,
 ...props,
}) = (
   <PRCalendar
      disabled={isDisabled}
      invalid={isInvalid}
      required={isRequired}
      visible={isVisible}
      {...props}
   />
)

```

I get the error:
![image](https://github.com/primefaces/primereact/assets/90120572/fd97446d-178d-40dd-b27b-fc4effd830a8)

Even when I just use the Component without creating a wrapper, passing in selectionMode value shows an error:

````
const Calendar: React.FC<CalendarProps> = ({ ...props }) => (
    <PRCalendar
        selectionMode="multiple"
        {...props}
    />
);
```

### Expected behavior

Calendar should accept values for selectionMode prop.

Rekl0w:
Can you provide a stackblitz link please ?

JosuerBague:
> Can you provide a stackblitz link please ?

https://stackblitz.com/edit/vitejs-vite-1hvojf?file=src%2FApp.tsx

melloware:
OK so it works for me in TypeScript here: https://stackblitz.com/edit/kd3h15?file=src%2FApp.tsx

i think the issue is the Calendar detects the `value` and sets the SelectionMode based on whether you pass it a single Date, a Date[] array, etc.

See here: https://github.com/primefaces/primereact/blob/686262f375b41cf6cecc30e475496619427ac2fb/components/lib/calendar/calendar.d.ts#L972-L1041

JosuerBague:
[...]

PR #7556, opened 2025-01-03
Fix(Calendar): Improve type declaration using generic types with default parameters (Fixes #7555)

### LINKED ISSUE #7555: Calendar: Improve type declaration using generic types with default parameters
### Describe the bug

I've encountered a problem while working with the `Calendar` component and creating an uncontrolled wrapper around it. The issue lies in the current type overloading approach used in the type definition:

`export type CalendarProps = CalendarPropsRange | CalendarPropsMultiple | CalendarPropsSingle;`

This implementation makes it difficult to dynamically manage the `selectionMode` without introducing additional boilerplate.

### Reproducer

https://stackblitz.com/edit/vitejs-vite-7rwuaddq?file=src%2FApp.tsx

### System Information

```Shell
"primereact": "^10.9.1"
"react": "^18.3.1"
```

1. Check typescript errors in the provided codesandbox

### Expected behavior

I propose replacing the type overloading approach with generic types that include default parameters. This change will enhance the flexibility and usability of the `CalendarProps` type. 

Example:
```ts
export interface CalendarProps<
    TMode extends CalendarSelectionMode = 'single',
    TValue = TMode extends 'multiple' ? Date[] : TMode extends 'range' ? (Date | null)[] : Date
> extends CalendarBaseProps {
    selectionMode?: TMode;
    value?: Nullable<TValue>;
    onChange?(event: FormEvent<TValue>): void;
}
```

iamkyrylo:
### Related Issue
Closes #7555 

### Summary
This PR implements the proposed improvement by replacing type overloading with generic types for `CalendarProps`. It addresses the problem with dynamically managing `selectionMode` as discussed in the issue.

### Screenshots
I've checked the solution by creating a small wrapper inside the project. Here is a screenshot where you can see that TypeScript correctly defines the `onChange` event type based on the passed `selectionMode`.

Default (single):
Screenshot 2025-01-03 at 12 30 31

Multiple:
Screenshot 2025-01-03 at 12 30 36

### DIFF
--- a/components/lib/calendar/calendar.d.ts
+++ b/components/lib/calendar/calendar.d.ts
@@ -965,76 +965,30 @@ interface CalendarBaseProps {
      */
     unstyled?: boolean;
 }
-/**
- * Defines valid properties in single Calendar component.
- * @group Properties
- */
-interface CalendarPropsSingle extends CalendarBaseProps {
-    /**
-     * Specifies the selection mode either "single", "range", or "multiple";
-     * @defaultValue single
-     */
-    selectionMode?: 'single' | undefined;
-    /**
-     * Value of the component.
-     * @defaultValue null
-     */
-    value?: Nullable<Date>;
-    /**
-     * Callback to invoke when value changes.
-     * @param { FormEvent<Date>} event - Custom change event
-     */
-    onChange?(event: FormEvent<Date>): void;
-}
-/**
- * Defines valid properties in range Calendar component.
- * @group Properties
- */
-interface CalendarPropsRange extends CalendarBaseProps {
-    /**
-     * Specifies the selection mode either "single", "range", or "multiple";
-     * @defaultValue single
-     */
-    selectionMode: 'range';
-    /**
-     * Value of the component.
-     * @defaultValue null
-     */
-    value?: Nullable<(Date | null)[]>;
-    /**
-     * Callback to invoke when value changes.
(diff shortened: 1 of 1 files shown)


melloware:
Interesting. Does this break backwards compatibility?

iamkyrylo:
> Interesting. Does this break backwards compatibility?

No, it shouldn't break anything. Previously, only `CalendarProps` were exposed, so users won't need to update their code, I believe. But it depends on how they applied it 🤷‍♂️

TS can still infer the `onChange` event and `value` types automatically based on the `selectionMode` value. This means you don't need to explicitly pass `TMode` type, like `Calendar<TMode>`. The key difference is that TS now directly uses the provided value instead of analyzing multiple type overloads.

melloware (approved):

melloware:
Thanks!

melloware:
This also fixes #6583

dgraham-nectivio:
To anyone who, like me, is now getting one of these errors:

`error TS2724: '"primereact/calendar"' has no exported member named 'CalendarPropsSingle'`
`error TS2724: '"primereact/calendar"' has no exported member named 'CalendarPropsMultiple'`
`error TS2724: '"primereact/calendar"' has no exported member named 'CalendarPropsRange'`

It's due to a breaking change in this commit, you now need to reference the generic type:

You need to use `CalendarProps<'single'>`, `CalendarProps<'multiple'>`, `CalendarProps<'range'>`

PR #7558, opened 2025-01-03
fix: numeric key filter validation for dash and dot input

### LINKED ISSUE #7557: KeyFilter: Incorrect handling of dash and dot in numeric inputs
### Describe the bug

https://primereact.org/keyfilter/#presets

In "Integer" and "Number" input fields where multiple dashes (-) or dots (.)  can be entered

### Reproducer

https://primereact.org/keyfilter/#presets

### System Information

```Shell
"primereact": "^10.9.1"
"react": "^18.3.1"
```

_No response_

### Expected behavior

_No response_

navedqb:
Fixes #7557

### DIFF
--- a/components/lib/keyfilter/KeyFilter.js
+++ b/components/lib/keyfilter/KeyFilter.js
@@ -71,11 +71,10 @@ export const KeyFilter = {
             return;
         }
 
+        const value = e.target.value;
         const regex = this.getRegex(keyfilter);
 
-        if (!regex.test(key)) {
-            e.preventDefault();
-        }
+        this.validateSpecialKey(e, key, keyfilter, regex, value)
     },
 
     validate(e, keyfilter) {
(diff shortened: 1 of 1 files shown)


melloware:
@navedqb this is intentional.  KeyFilter by default just filters SINGLE keystrokes to prevent input but not validating something is a valid number.   All of the PrimeTek libraries operate this way.

Look at the Whole Word example: https://primereact.org/keyfilter/#regexword

If you want to use `int` and only allow the negative at the start of the word try

```xml
<InputText id="numkeys" keyfilter={/^-?\d+$/} validateOnly onInput={validateInput} />
```

That will allow only 1 minus sign and it must be the first character only.

I am sure you can ChatGPT to find a good numeric one also that only allows one "decimal".

But the component is working as designed.

Issue #7553, opened 2025-01-03
Calendar: onSelect is forcing date while typing


manuel-bitcraft:
### Describe the bug

Adding the onSelect callback (even an empy function) changes the typing behaviour for the date.

### Reproducer

https://stackblitz.com/edit/dppelmek?file=src%2FApp.jsx

### System Information

```Shell
System:
    OS: Windows 11 10.0.22631
    CPU: (12) x64 13th Gen Intel(R) Core(TM) i7-1355U
    Memory: 41.48 GB / 63.64 GB
  Binaries:
    Node: 18.20.4 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.22.22 - C:\Program Files\nodejs\yarn.CMD
    npm: 10.7.0 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Edge: Chromium (128.0.2739.67)
    Internet Explorer: 11.0.22621.3527
  npmPackages:
    primereact: ^10.8.5 => 10.8.5 
    react: ^18 => 18.2.0 
    tailwindcss: ^3.3.0 => 3.4.1
```

### Steps to reproduce the behavior

Notes: using format dd/mm/yy
Objective: select date 15/01/2025 and then modify date to '15/11/2025' by typing

**With onSelect commented out**
1. Select date 15/01/2025
2. Place cursor after month (01)
3. Press backspace twice so date reads '15//2025'
4. type 1 so the date  reads '15/1/2025'
5. type 1 againg so date reads '15/11/2025'
6. Success

**With onSelect enabled as an empty callback**
1. Select date 15/01/2025
2. Place cursor after month (01)
3. Press backspace twice so date reads '15//2025'
4. as soon as you type 1, the date changes to '15/01/2025' and the cursor is moved at the end 
5. if you try to delete the '0' for the '01' month (so I can type 1 to set the month to 11), the date stays to '15/01/2025' and the cursor is moved at the end 
6. Failed

melloware:
looks like its related to this fix: https://github.com/primefaces/primereact/issues/7121

KumJungMin:
- I have reviewed the issue :)
- As a result, I found that the same problem occurs in the previous version (v10.8.3) as well 😢 
- After investigating the cause of the issue, I discovered that the `onSelect` event is triggered when typing in the input field.

[...]

PR #7563, opened 2025-01-04
fix(Calendar): not trigger onViewDateSelect when input typing

KumJungMin:
## Defect Fixes
- fix: #7553

<br/>

### Cause of the Issue

_issue occurred flow (current version)_
```mermaid
graph TD
    A[User types in input field] --> B[onUserInput function is triggered]
    B --> C[updateValueOnInput function is triggered]
    C --> D[onViewDateSelect function is triggered]
    D --> E[onSelect function is triggered]
    E --> F["Issue occurred!!"] 
```

- Due to the above flow, when typing in the input field, the onSelect event is triggered, causing a bug.

<br/>

### Proposed Solution
- To address the issue, the behavior of the onSelect function has been modified so that it is triggered only when the user interacts with the date picker button.
- **However, if the onSelect function is also intended to work while typing in the input field, please leave a comment, and I will update it accordingly.**

<br/>

### Test
```js
<Calendar value={date} onChange={(e) => setDate(e.value)} onSelect={(e) => console.log(e.value)} />
```

https://github.com/user-attachments/assets/cbca041c-a44b-4d82-b130-f5af6c8e3085

### DIFF
--- a/components/lib/calendar/Calendar.js
+++ b/components/lib/calendar/Calendar.js
@@ -162,7 +162,6 @@ export const Calendar = React.memo(
                     const date = value.length ? value[0] : value;
 
                     updateViewDate(event, date);
-                    onViewDateSelect({ event, date });
                 }
             } catch (err) {
                 //invalid date


melloware (approved):

melloware:
Nice it looks like this was introduced in #7155

PR #7588, opened 2025-01-09
Fix #7593: disableKeyboardNavigation prop for Tree component

### LINKED ISSUE #7593: Tree: can't use keyboard inputs for tree sub component
### Describe the bug

For example, I'm working on using the component where you can change the name of a node thanks to an input texte that appears in place of the node name. So i want to deactivate temporarily the key navigation.

image

But you can't move from right to left in the input because it activates the tree controls.
I think it's blocking for this type of situation where you want to do other manipulations inside the tree.

### Reproducer

_No response_

### System Information

```Shell
System:
    OS: Windows 11 10.0.22621
    CPU: (24) x64 13th Gen Intel(R) Core(TM) i7-13700T
    Memory: 1.32 GB / 15.65 GB
  Binaries:
    Node: 20.17.0 - C:\Program Files\nodejs\node.EXE
    npm: 10.8.2 - C:\Program Files\nodejs\npm.CMD
    pnpm: 9.12.1 - ~\AppData\Local\pnpm\pnpm.EXE

primereact 10.8.4
react 18.3.3
tailwindcss 3.4.14
```

_No response_

### Expected behavior

_No response_

CorentinAT:
Fix #7593: disableKeyboardNavigation prop for Tree component

[Small feature]
Added a "keyboardKeyboardNavigation" to prevent expansion and collapse with arrow keys, and all key navigation (enter, tab, space, arrows)

### DIFF
--- a/components/doc/common/apidoc/index.json
+++ b/components/doc/common/apidoc/index.json
@@ -52451,6 +52451,14 @@
                             "description": "A single key to control the selection with the context menu."
                         },
                         {
+                            "name": "disableKeyboardNavigation",
+                            "optional": true,
+                            "readonly": false,
+                            "type": "boolean",
+                            "default": "false",
+                            "description": "When present, it disables keyboard navigation around the tree."
+                        },
+                        {
                             "name": "disabled",
                             "optional": true,
                             "readonly": false,
--- a/components/lib/tree/Tree.js
+++ b/components/lib/tree/Tree.js
@@ -417,6 +417,7 @@ export const Tree = React.memo(
                     collapseIcon={props.collapseIcon}
                     contextMenuSelectionKey={props.contextMenuSelectionKey}
                     cx={cx}
+                    disableKeyboardNavigation={props.disableKeyboardNavigation}
                     disabled={props.disabled}
                     dragdropScope={props.dragdropScope}
                     expandIcon={props.expandIcon}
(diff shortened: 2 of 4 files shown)


CorentinAT:
For example, I'm working on using the component where you can change the name of a node thanks to an input texte that appears in place of the node name. So i want to deactivate temporarily the key navigation.
But you can't move from right to left in the input because it activates the tree controls.
I think the feature would be interesting for this type of situation where you want to do other manipulations inside the tree.

For WCAG/ARIA, it is only a property manually assigned by the developer (and I think it's good to give the developer the freedom to decide where to apply the WCAG/ARIA principles)

melloware:
also please open a ticket first with your scenario.

CorentinAT:
Ok thank you very much I didn't know this at all, I'm trying this afternoon. I quickly open a ticket

sja-cslab:
@melloware I'm not in this feature but wouldn't a simple bubble stop just work - e.stopPropagation/preventDefault?

CorentinAT:
> @melloware I'm not in this enhancement but wouldn't a simple bubble stop just work - e.stopPropagation/preventDefault?

I was completely unfamiliar with e.stopPropagation(). I've just tested it and it works in the same way as my pull request, thank you very much!

Issue #7599, opened 2025-01-13
Tree: Incorrect props and types in documentation


diegoromerodev:
### Describe the bug

The documentation mentions that there is a leaf property for TreeNode,
https://primereact.org/tree/#api.TreeNode.props.leaf
but that is incorrect, after logging the context, I can see that the property is actually named: isLeaf
![image](https://github.com/user-attachments/assets/04c3e85b-4b7a-46b3-8be5-e22f534cad27)

As an additional note, the types for the pt prop are mostly broken as well:
![image](https://github.com/user-attachments/assets/44f41d39-697d-4a8d-98da-de9c062c9170)

I sincerely appreciate your work on this library, it has saved me a few hours!

### System Information

```Shell
npmPackages:
primereact: '10.9.1',
react: "19.0.0",
typescript: "5.7.2"
```

PR #7600, opened 2025-01-14
fix: #7599 , Tree: Incorrect props and types in documentation

akshayaqburst:
fix: #7599 , Tree: Incorrect props and types in documentation


melloware:
@akshayaqburst i think we want to leave it as `leaf` in the JS I think we just need to make 1 or two changes I think I can submit a PR.

melloware:
see: https://github.com/primefaces/primereact/pull/7603

PR #7657, opened 2025-01-29
fix: floatLabel incorrectly staying active when Dropdown value is null

### LINKED ISSUE #7637: Dropdown: FloatLabel issues
### Describe the bug

when using dropdown element in combination with floatlabel, the floatlabel is "active" even when the dropdown component's value is null. ironically, if you click on the clearIcon, the floatlabel returns to the default position as it should be when the dropdown has value null.

i first noticed this bug in 10.9.1. the issue is present even in the example page [](https://primereact.org/dropdown/#floatlabel) 

### Reproducer

https://stackblitz.com/edit/x1jh55xi?file=src%2FApp.jsx

### System Information

```Shell
primereact 10.9.1
```

_No response_

### Expected behavior

_No response_

navedqb:
Fixes #7637

### DIFF
--- a/components/lib/dropdown/DropdownBase.js
+++ b/components/lib/dropdown/DropdownBase.js
@@ -1,6 +1,6 @@
 import PrimeReact from '../api/Api';
 import { ComponentBase } from '../componentbase/ComponentBase';
-import { classNames } from '../utils/Utils';
+import { classNames, ObjectUtils } from '../utils/Utils';
 
 const classes = {
     root: ({ props, focusedState, overlayVisibleState, context }) =>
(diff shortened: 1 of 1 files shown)


melloware (approved):

components/lib/dropdown/DropdownBase.js
```diff
@@ -10,7 +10,7 @@ const classes = {
             'p-focus': focusedState,
             'p-variant-filled': props.variant ? props.variant === 'filled' : context && context.inputStyle === 'filled',
             'p-dropdown-clearable': props.showClear && !props.disabled,
-            'p-inputwrapper-filled': props.value !== undefined,
+            'p-inputwrapper-filled': props.value !== undefined && props.value !== null,
```
melloware, line 13: I wonder if we should use ObjectUtils.isNotEmpty(props.value)` ?

melloware (approved):

PR #7659, opened 2025-01-29
Fix node expanded if not filtering

### LINKED ISSUE #7658: Tree: Node cannot be programmatically expanded
### Describe the bug

Hi,

I try to update prime react on project and break a Tree. Our tree has nodes who can be init with expanded or not. Tree can know if node is opened or not.

After release 10.9.2, not work... I think that [PR](https://github.com/primefaces/primereact/pull/7282/files#diff-54e22c623d60c2f02035d9621049e7d57f7673967b080e37c0ae6179feff6266R18) breaking changes.

We need to set expanded on node props, easy to manipulate.

### Reproducer

_No response_

### System Information

```Shell
10.9.2
```

_No response_

### Expected behavior

_No response_

matthieumota:
That PR fix #7658

### DIFF
--- a/components/lib/tree/Tree.js
+++ b/components/lib/tree/Tree.js
@@ -421,6 +421,7 @@ export const Tree = React.memo(
                     dragdropScope={props.dragdropScope}
                     expandIcon={props.expandIcon}
                     expandedKeys={expandedKeys}
+                    isFiltering={isFiltering}
                     isNodeLeaf={isNodeLeaf}
                     metaKeySelection={props.metaKeySelection}
                     nodeTemplate={props.nodeTemplate}
--- a/components/lib/tree/UITreeNode.js
+++ b/components/lib/tree/UITreeNode.js
@@ -15,7 +15,8 @@ export const UITreeNode = React.memo((props) => {
     const mergeProps = useMergeProps();
     const isLeaf = props.isNodeLeaf(props.node);
     const label = props.node.label;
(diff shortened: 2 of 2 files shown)


melloware:
cc @KumJungMin

KumJungMin:
If you show the test code you wrote, it will be easier to understand the situation :)

matthieumota:
In your example, you must be allow to set `expanded` key to true or false to open or not node.

KumJungMin:
> In your example, you must be allow to set `expanded` key to true or false to open or not node.

----

- Thank you for checking the sample code and for fixing my incorrect code! I understand your intention. :)
- In your code, when a node item is set to `expanded: true`, the node expands as expected. 👍 
- *_This is just a personal curiosity: In the past, when child nodes were set to expanded: true (Question A), would the entire child group expand as well?_ @matthieumota 
```js
const nodes = [
        {
            key: '0',
            label: 'Root',
            expanded: true, /* this! */
            children: [
                { key: '0-0', label: 'Child 1' },
                { key: '0-1', label: 'Child 2' }
            ]
        },
        {
            key: '2',
            label: 'Root 3',
            children: [
                { key: '2-0', label: 'Child 3.1' },
                { key: '2-1', label: 'Child 3.2', expanded: true } /* question(A) */
            ]
        }
    ];
``

matthieumota:
In the past yes ! Reason to that PR.

melloware (approved):

PR #7722, opened 2025-02-19
FileUpload: Added Custom Button label prop for after file is selected

### LINKED ISSUE #7721: Basic FileUploader - Set a Custom Button Label After Selecting a File
### Describe the bug

With the basic fileuploader, is there a way to set the label of the button after selecting a file? It seems it always gets set to the name of the file, and trying to manually set it via any of the chooseLabel/chooseOptions or uploadLabel/uploadOptions has no impact. It seems the uploadLabel/uploadOptions only apply to the advanced file uploader. I was able to customize the choose button in the basic mode with chooseOptions, but once I select a file, it sets the button label to the name of the file and I wasn't able to change it.

I'm not really familiar with this codebase, but after a quick look, maybe this is happening here: https://github.com/primefaces/primereact/blob/master/components/lib/fileupload/FileUpload.js#L693. It seems if 'hasFiles' is true, then it sets the label to the filename, otherwise the chooseLabel

### Reproducer

https://stackblitz.com/edit/ed8hwoju?file=src%2FApp.tsx

### System Information

```Shell
System:
    OS: Linux 5.15 Ubuntu 22.04.4 LTS 22.04.4 LTS (Jammy Jellyfish)
    CPU: (8) x64 11th Gen Intel(R) Core(TM) i9-11950H @ 2.60GHz
    Memory: 6.48 GB / 11.69 GB
    Container: Yes
    Shell: 5.1.16 - /bin/bash
  Binaries:
    Node: 18.19.1 - ~/.nvm/versions/node/v18.19.1/bin/node
    npm: 10.2.4 - ~/.nvm/versions/node/v18.19.1/bin/npm
  Browsers:
    Brave Browser: 122.1.63.174
  npmPackages:
    primereact: ^10.8.4 => 10.8.4
    react: ^18.2.0 => 18.2.0
```

With basic file uploader click to upload file. Once file is selected, the label of the button will be the filename. 

### Expected behavior

It would be nice if there was a way to specify a custom label for the button after selecting a file. For example 'Upload'.

timmy471:
Fix #7721

### DIFF
--- a/components/lib/fileupload/FileUpload.js
+++ b/components/lib/fileupload/FileUpload.js
@@ -690,7 +690,7 @@ export const FileUpload = React.memo(
                 ptm('label')
             );
             const chooseLabel = chooseOptions.iconOnly ? <span {...labelProps} dangerouslySetInnerHTML={{ __html: '&nbsp;' }} /> : <span {...labelProps}>{chooseButtonLabel}</span>;
-            const label = props.auto ? chooseLabel : <span {...labelProps}>{hasFiles ? filesState[0].name : chooseLabel}</span>;
+            const label = props.auto ? chooseLabel : <span {...labelProps}>{hasFiles ? (props.selectedFileLabel || filesState[0].name) : chooseLabel}</span>;
             const chooseIconProps = mergeProps(
                 {
                     className: cx('chooseIcon', { iconOnly: chooseOptions.iconOnly })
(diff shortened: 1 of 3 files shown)


melloware (changes requested):
Don't you also need to add this to the `.d.ts` Typescript file?

melloware (approved):

melloware (approved):

Issue #2012, opened 2021-05-04
Multiselect item on 0 index is getting selected


Nirav4Stridely:
I have just recently updated primereact version from 6.1.0 to 6.3.0. I have found issue that Item which set on 0 is not getting selected.

can you please help me? here is my code snippet example...

My List: 
`var list : [{id:0, name:"All"},{id:1, name:"User 1"},{id:2, name:"User 2"},{id:3, name:"User 3"},{id:4, name:"User 4"}];`

`<MultiSelect id="mulUser" filter
                            value={this.state.userIds} options={this.state.list}
                            optionLabel="name" optionValue="id" .... />`

![image](https://user-images.githubusercontent.com/76509793/117041025-23737b80-ad28-11eb-835f-75e646e8c998.png)

mertsincan:
Hi,

I couldn't replicate it. Please try;
https://codesandbox.io/s/nifty-dew-of8e5?file=/src/demo/MultiSelectDemo.js

Also, could you please try it with 6.3.1?

For now, closed. If you persist problem, please reopen this ticket.

Best Regards,

Nirav4Stridely:
Hello,

I have regenerate in your given link. please code update code. Also tried with 6.3.1.
[https://codesandbox.io/s/silly-black-i79d6?file=/src/demo/MultiSelectDemo.js](url)

I am reopening again.

mertsincan:
Thanks a lot for your update! Fixed now.

Best Regards,

Nirav4Stridely:
Hello,

Thank you for the update!

can i know the release date of this fix? Because, if i go with current V6.3.0, i need to make so many changes in my project and we are about to release now.

PR #7726, opened 2025-02-20
Update MultiSelect.js: do not return option as object if optionValue is present

### LINKED ISSUE #7724: MultiSelect: Support options with null value
### Describe the bug

MultiSelect should support options with value null, like Dropdown does.
```
function handleChange(e){
  if(e.value && e.value[0] === 1){ //true
    console.log("First selected"); //Works, if only "First" is selected
  }
  if(e.value && e.value[0] === null){ //false
    console.log("Nothing selected"); //Works not, if only "Nothing" is selected
  }
  if(e.value && e.value[0].id === null){ //true
    console.log("Nothing selected"); //Works, if only "Nothing" is selected, but works different than value !== null
  }

<MultiSelect value={[]} onChange={handleChange} optionValue="id" opions={[
  {id: null, label: "Nothing"}, 
  {id: 1, label: "First"}
]} />
```

Bug is located here: https://github.com/primefaces/primereact/blob/master/components/lib/multiselect/MultiSelect.js#L625

### Reproducer

_No response_

### System Information

```Shell
PrimeReact 10.9.2 (latest)
```

_No response_

### Expected behavior

_No response_

nico-richter:
Fix #7724
If optionValue is present it is assumed that option is object and has the attribute.
So options with null values are supported

### DIFF
--- a/components/lib/multiselect/MultiSelect.js
+++ b/components/lib/multiselect/MultiSelect.js
@@ -620,9 +620,7 @@ export const MultiSelect = React.memo(
             }
 
             if (props.optionValue) {
-                const data = ObjectUtils.resolveFieldData(option, props.optionValue);
-
-                return data !== null ? data : option;
+                return ObjectUtils.resolveFieldData(option, props.optionValue);
             }
 
             return option && option.value !== undefined ? option.value : option;


melloware:
Have you look look at Git Blame and the history. I feel like you are breaking another scenario already fixed previously...

nico-richter:
Code was added here: https://github.com/primefaces/primereact/commit/e5bf9adaec93ff4ee8da3d2f6f34526fbc13c464

To solve #2012 there was no need to add the line `return data !== null ? data : option;`
`return data` had solved the issue. But this extra line allows 0 like in #2012 but does prevent null values.

I hope, this helps to understand. It was really hard to find this, because it was not in history because of renaming and blame also did not show me this change.

PR #7785, opened 2025-03-14
#7569: Added frozenEditingMetaState for DataTable

### LINKED ISSUE #7569: DataTable: Values of frozen row's cells are duplicated on first non-frozen row
### Describe the bug

When using the DataTable component with editMode="cell", having an editable frozen row and multiple editable non-frozen rows causes the value from the frozen row to be copied to the first non-frozen row.

### Reproducer

https://stackblitz.com/edit/as8ypzgp-y2bs8hf4?file=src%2FApp.tsx

### System Information

```Shell
primereact: ^10.8.5;
react: ^18.3.1
```

I need to have a frozen row (using the frozenValue property) and multiple non-frozen rows (using the value property).

Both the frozen row and the sub-rows need to be editable (in the provided example, all cells are editable).

Here is how to reproduce the bug:

1. Click on the Name Column of the frozen row (Do not modify the value !)
image

2. Click directly to the Name column of the first non-frozen row
3. The value of the frozen row is now in the non-frozen row :
image

Do not hesitate to tell me if this is a problem on my implementation ! 

Thanks a lot !

### Expected behavior

The value of the frozen row should not be copied to the first non-frozen row

tneeraja95:
fix #7569
**Issue** : [7569](https://github.com/primefaces/primereact/issues/7569)
Values of frozen row's cells are duplicated on non-frozen rows and vice-versa. 

**Root-cause** : 
This is happening as these are editable rows and it uses useState hook to manage the state and the same variable was getting used for frozen as well as the non-frozen rows. 

**Solution**:
Use a different state hook variable for the frozen-rows.

### DIFF
--- a/components/lib/datatable/DataTable.js
+++ b/components/lib/datatable/DataTable.js
@@ -28,6 +28,7 @@ export const DataTable = React.forwardRef((inProps, ref) => {
     const [columnOrderState, setColumnOrderState] = React.useState([]);
     const [groupRowsSortMetaState, setGroupRowsSortMetaState] = React.useState(null);
     const [editingMetaState, setEditingMetaState] = React.useState({});
+    const [frozenEditingMetaState, setFrozenEditingMetaState] = React.useState({});
     const [d_rowsState, setD_rowsState] = React.useState(props.rows);
     const [d_filtersState, setD_filtersState] = React.useState({});
     const metaData = {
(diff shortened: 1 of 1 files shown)


melloware (approved):

melloware (changes requested):
Why not just keep it all editingMetaData state and simply do a switch to check if the table is `frozen`?

melloware:
I guess can there be editingMetadata and frozenMetadata at the same time?

PR #7823, opened 2025-03-25
Fix: #7817 (Autocomplete type issue)

### LINKED ISSUE #7817: AutoComplete: Select and Unselect events has any typed value
### Describe the bug

`AutoCompleteSelectEvent` (and `AutoCompleteUnselectEvent` in extension) has `value` attribute typed as explicit `any` instead of being inferred from AutoComplete's `value`.

Image

### Reproducer

_No response_

### System Information

```Shell
System:
    OS: Windows 11 10.0.26100
    CPU: (32) x64 AMD Ryzen 9 5950X 16-Core Processor
    Memory: 8.54 GB / 31.91 GB
  Binaries:
    Node: 22.14.0 - C:\Program Files\nodejs\node.EXE
    Yarn: 4.7.0 - C:\Program Files\nodejs\yarn.CMD
    npm: 11.0.0 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Edge: Chromium (134.0.3124.83)
```

_No response_

### Expected behavior

_No response_

sarathkumarsasi:
Added inferred types for AutoCompleteSelectEvent (and AutoCompleteUnselectEvent in extension). 
Fix #7817

### DIFF
--- a/components/lib/autocomplete/autocomplete.d.ts
+++ b/components/lib/autocomplete/autocomplete.d.ts
@@ -292,7 +292,7 @@ export interface AutoCompleteProps<
     /**
      * Icon of the dropdown.
      */
-    dropdownIcon?: IconType<AutoCompleteProps> | undefined;
+    dropdownIcon?: IconType<AutoCompleteProps<T>> | undefined;
     /**
      * Specifies the behavior dropdown button. Default "blank" mode sends an empty string and "current" mode sends the input value.
      * @defaultValue blank
@@ -568,7 +568,7 @@ export interface AutoCompleteProps<
  *
  * @group Component
  */
-export declare class AutoComplete extends React.Component<AutoCompleteProps, any> {
+export declare class AutoComplete<T = any> extends React.Component<AutoCompleteProps<T>, any> {
     /**
      * Used to show the overlay.
      */
--- a/components/lib/autocomplete/autocomplete.d.ts
+++ b/components/lib/autocomplete/autocomplete.d.ts
@@ -335,7 +335,7 @@ export interface AutoCompleteProps<
     /**
      * Template of a list item.
      */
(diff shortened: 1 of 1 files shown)


sarathkumarsasi:
> I don't think this is correct? It still won't pick up what `props.value` is from what I can tell?

Yes, you were right. I need to add the generic type in the props also; I have fixed it. Can you verify now?

melloware:
I still don't think this is correct...

sarathkumarsasi:
> I still don't think this is correct...

I have added typing for the autocomplete suggestions, I guess this will fix the issue. If not, can you suggest any other approach?

melloware:
right i see `value: any` shouldn't that be `value: T` ?

sarathkumarsasi:
> value: any

Yeah, I will change it. I found a way to test the changes locally, I will push after testing it.

components/lib/autocomplete/autocomplete.d.ts
```diff
@@ -38,31 +38,33 @@ export interface AutoCompletePassThroughMethodOptions {
...
+interface AutoCompleteSelectEvent<T = any> {
     /**
      * Browser event
      */
     originalEvent: React.SyntheticEvent;
     /**
      * Selected option value
      */
-    value: any;
+    value: T extends any[] ? T[number] : T;
```
melloware, line 58: i don't understand T[number] here?

sarathkumarsasi: T[number] is a way to query the type of the elements in T when T is an array or tuple.
It’s like saying: "If T is an array, give me the type of its elements; otherwise, I don’t care about this part because the condition failed."
T[number] represents the type of elements that can be accessed by indexing into T with a number. Essentially, it "extracts" the type of the array's elements.

PR #7839, opened 2025-03-29
fix(Menu): popup mode with template hover issue

### LINKED ISSUE #7826: Menu: Menu item is not focused on hover when the template option is used
### Describe the bug

When on some of the options you set the custom template, then if you hover over the item and then move the mouse outside of the dropdown, the previous item is focused instead of the last one.

Image

However, it seems to work correctly when you use the keyboard's arrow keys instead:

Image

### Reproducer

https://stackblitz.com/edit/q6j3zrtj?file=src%2FApp.jsx

### System Information

```Shell
System:
    OS: Windows 10 10.0.19045
    Memory: 17.13 GB / 31.93 GB
  Binaries:
  Browsers:
  npmPackages:
    primereact: ^10.9.3 => 10.9.3
    react: ^18.2.0 => 18.2.0
```

1. Go to https://stackblitz.com/edit/q6j3zrtj?file=src%2FApp.jsx
2. Click on the one of the available buttons
3. Hover through the items
4. Move the cursor out of the panel area

### Expected behavior

The last Item is focused, or there is a possibility just to disable this focusing as described here: #7511

sahilrajthapa:
Fixes #7826

Expose `onMouseMove` event listener to popup menu template options

```

{
    label: 'Custom template',
    template: (item, options) => {
        return (
            <div className="p-menuitem-content" data-pc-section="content" onMouseMove={options.onMouseMove}>
                <a className="p-menuitem-link">
                    Lorem ipsum
                </a>
            </div>
        );
    }
}
```

The issue can also be fixed without exposing the `onMouseMove` event listener to the template options. We can replace the [line](https://github.com/primefaces/primereact/blob/master/components/lib/menu/Menu.js#L408) with the following:

```
 const templateContent = ObjectUtils.getJSXElement(item.template, item, defaultContentOptions);  
                
 content = React.cloneElement(templateContent, {
     onMouseMove: (event) => onItemMouseMove(event, key),
})
```
@melloware Please let me know your opinion on the second approach. If it looks good, I'll update the PR.

### DIFF
--- a/components/doc/menu/popupdoc.js
+++ b/components/doc/menu/popupdoc.js
@@ -22,6 +22,18 @@ export function PopupDoc(props) {
                 {
                     label: 'Export',
                     icon: 'pi pi-upload'
+                },
+                {
+                    label: 'Custom template',
+                    template: (item, options) => {
+                        return (
+                            <div className="p-menuitem-content" data-pc-section="content" onMouseMove={(e) => options.onMouseMove(e)}>
+                                <a href="#" className="p-menuitem-link">
+                                    Lorem ipsum
+                                </a>
+                            </div>
+                        );
+                    }
                 }
             ]
         }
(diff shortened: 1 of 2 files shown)


melloware (approved):

melloware:
Hmmm that is a good point. I feel like the way you have done it is the right way and maybe we update the template example in the showcase to have onMouseMove?

sahilrajthapa:
Yeah sure, I will update the template example.

melloware (approved):

PR #7904, opened 2025-04-15
OverlayPanel fix following scroll

### LINKED ISSUE #7903: OverlayPanel: Position is not updated on scroll from wrapper element
### Describe the bug

If a part of your interface could scroll and displays an OverlayPanel: when scrolling, the overlay position is not updated.

https://github.com/user-attachments/assets/65bdffab-9f67-47f4-8cba-71ea54f828e1

Fix is coming in a PR:

https://github.com/user-attachments/assets/8921e5e1-2d5f-418d-a70d-fce08e5e066d

### Reproducer

_No response_

### System Information

```Shell
Primereact 10.9.3+
```

_No response_

### Expected behavior

_No response_

sescandell:
### Defect Fixes
Fix #7903

### DIFF
--- a/components/lib/utils/DomHandler.js
+++ b/components/lib/utils/DomHandler.js
@@ -620,7 +620,7 @@ export default class DomHandler {
      * @param {boolean} hideOverlaysOnDocumentScrolling - Whether to include window/document level scrolling
      * @returns {Array} Array of scrollable parent elements
      */
-    static getScrollableParents(element, hideOverlaysOnDocumentScrolling = false) {
+    static getScrollableParents(element) {
         let scrollableParents = [];
 
         if (element) {
@@ -674,11 +674,6 @@ export default class DomHandler {
                     addScrollableParent(parent);
                 }
             }
-        }
-
-        // Ensure window/body is always included as fallback
-        if (!scrollableParents.some((node) => node === document.body || node === window)) {
-            scrollableParents.push(hideOverlaysOnDocumentScrolling ? window : document.body);
         }
 
         return scrollableParents;
--- a/components/lib/utils/DomHandler.js
+++ b/components/lib/utils/DomHandler.js
@@ -647,10 +647,8 @@ export default class DomHandler {
              * @param {HTMLElement} node - Element to add
              */
             const addScrollableParent = (node) => {
-                if (hideOverlaysOnDocumentScrolling) {
-                    // For document/body/html elements, add window instead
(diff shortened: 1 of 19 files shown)


components/lib/utils/DomHandler.js
```diff
@@ -647,10 +647,8 @@ export default class DomHandler {
              * @param {HTMLElement} node - Element to add
              */
             const addScrollableParent = (node) => {
-                if (hideOverlaysOnDocumentScrolling) {
-                    // For document/body/html elements, add window instead
-                    scrollableParents.push(node.nodeName === 'BODY' || node.nodeName === 'HTML' || node.nodeType === 9 ? window : node);
-                }
+                // For document/body/html elements, add window instead
+                scrollableParents.push(node.nodeName === 'BODY' || node.nodeName === 'HTML' || node.nodeType === 9 ? window : node);
```
melloware, line 651: This was intentional to only do it if hidedocumentonScrolling is set. Some users want the behavior and others don't depending on their accessibility needs for closing overlays.  So this is not the right fix

melloware (changes requested):

sescandell:
@melloware you were totally right!

The thing is, the scroll event from useOverlayListener is not really a scroll event because of how DomHandler.getScrollableParents is related to hideOnScroll. In the update of this PR, I changed this DomHandler.getScrollbaleParents behavior and adapted all components.

Before PR:

https://github.com/user-attachments/assets/2257ea51-dfae-45cb-9fc1-4383f277195f

After PR:

https://github.com/user-attachments/assets/e78270da-bfd8-4b8f-bbbc-2d6157d0e264

Let me know your thoughts. 

Thanks

components/lib/utils/DomHandler.js
```diff
@@ -647,10 +646,8 @@ export default class DomHandler {
...
```
melloware, line 650: For readability i think we should add a `DomHandler.isDocumentNode(target)` method that does this `node.nodeType === 9` for readability. Then everywhere you use `node.nodeType === 9` change to use this new method which will make the code much more readable for eveyrone!

melloware (changes requested):

components/lib/utils/DomHandler.js
sescandell: Makes sense... commit is coming

melloware (approved):

yabelozub:
After these changes, scrolling with open dropdown crashes with an error

Screenshot 2025-04-28 at 16 38 34

Issue #7928, opened 2025-04-22
Calendar: Overlay arrow navigation through days is problematic


dimoiuBogdan:
### Describe the bug

When I navigate through the days in calendar overlay, as soon as I am on the last day of the row and I press "ArrowRight", I am prompted to the next month instead of going to the next arrow.

### System Information

```Shell
"primereact": "10.9.4"
"react": "19.1.0",=
```

### Steps to reproduce the behavior

1. Focus calendar input to open overlay
2. Focus any day of the week
3. Navigate through arrow keys all the way to the left / right
4. Press the arrow key once again
5. You are now on the previous / next month instead of previous / next row

![Image](https://github.com/user-attachments/assets/52c82ca3-8411-41c2-a251-fa5e049ec22c)

![Image](https://github.com/user-attachments/assets/22e5a0ef-f305-4684-bd28-2d422c8664a7)

### Expected behavior

In this situation I expect to go to the next / previous row instead of month. This is the normal UX everyone expects, which is implemented in all the available datepickers out there for MUI, Aria, etc

melloware:
agreed this is a bug.

melloware:
Looks like Primevue has the same issue: https://primevue.org/datepicker/

SOMWHY:
I would like to contribute to this, but how could I preview the Calendar or any other component?

PR #7933, opened 2025-04-23
Calendar: Overlay arrow navigation through days is problematic #7928

bogdan-dimoiu:
### Defect Fixes
I fixed issue [#7928](https://github.com/primefaces/primereact/issues/7928)

Fixed issue: When I navigate through the days in calendar overlay, as soon as I am on the last day of the row and I press "ArrowRight", I am prompted to the next month instead of going to the next arrow.

Result: Go to the next / previous row instead of month


melloware:
Marked Pending Review from PrimeTek as this deviates from the other Prime Libraries.

Issue #7935, opened 2025-04-24
FileUpload: Uploaded file name is not trimmed when files with larger filenames are uploaded.


akshayaqburst:
### Describe the bug

FileUpload: Uploaded file name is not trimmed when files with larger filenames are uploaded.

### Reproducer

https://primereact.org/fileupload/#advanced

### System Information

```Shell
System:
OS: Linux 5.15 Ubuntu 20.04.6 LTS (Focal Fossa)
CPU: (8) x64 11th Gen Intel(R) Core(TM) i7-1165G7 @ 2.80GHz
Memory: 1.43 GB / 7.47 GB
Container: Yes
Shell: 5.0.17 - /bin/bash
Binaries:
Node: 22.14.0 - ~/.nvm/versions/node/v22.14.0/bin/node
npm: 10.9.2 - ~/.nvm/versions/node/v22.14.0/bin/npm
pnpm: 10.4.1 - ~/.nvm/versions/node/v22.14.0/bin/pnpm
npmPackages:
react: 18.3.1 => 18.3.1
```

### Steps to reproduce the behavior

1. https://primereact.org/fileupload/#advanced
2. Upload a file with a larger name.
3. The file name displayed is not trimmed and display takes a bigger area.

<img width="911" alt="Image" src="https://github.com/user-attachments/assets/67785e29-52f1-4852-82b5-72dd7a034ca2" />

### Expected behavior

The file name should be trimmed as per the prop passed or a default file name length. 
Expected screenshot attached - 

<img width="728" alt="Image" src="https://github.com/user-attachments/assets/74e8c956-521a-4f90-a673-1260c2147a58" />

akshayaqburst:
I will work on this

PR #7936, opened 2025-04-24
fix: #7935, FileUpload: Uploaded file name is not trimmed when files with larger filenames are uploaded.

akshayaqburst:
fix: #7935, FileUpload: Uploaded file name is not trimmed when files with larger filenames are uploaded.

After fix - 
On hovering, tooltip will be displayed as shown.

image


akshayaqburst:
@melloware Please review and merge

PR #7951, opened 2025-04-30
DataTable: Fix and Simplify overlay handling in cell edit mode 

### LINKED ISSUE #7158: DataTable: Cell Editing and Dropdown Filter Interaction
### Describe the bug

When using a PrimeReact DataTable with cell editing mode (editMode="cell") and a Dropdown as the cell editor, there is an issue when trying to interact with the filter feature of the Dropdown.

The issue occurs as follows:

1. Clicking on an editable cell that uses a Dropdown opens the editor as expected.
2. If the Dropdown has filtering enabled (filter), typing in the filter input causes the DataTable to detect a loss of focus on the cell, prematurely closing the editor.
3. This prevents the user from selecting a filtered option because the Dropdown closes as soon as focus is lost, forcing the selection of the currently displayed value instead of allowing the user to select a new one.

### Reproducer

https://stackblitz.com/edit/vitejs-vite-z4vrkd?file=src%2FApp.tsx

### System Information

```Shell
System:
    OS: Windows 10 10.0.22631
    CPU: (12) x64 AMD Ryzen 5 5600G with Radeon Graphics
    Memory: 12.17 GB / 31.37 GB
  Binaries:
    Node: 20.12.2 - C:\Program Files\nodejs\node.EXE
    npm: 10.5.1 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Edge: Chromium (127.0.2651.74), ChromiumDev (129.0.2766.0)
  npmPackages:
    primereact: 10.6.3 => 10.6.3
    react: ^18.2.0 => 18.3.1
    tailwindcss: ^3.4.4 => 3.4.4
```

_No response_

### Expected behavior

When a user interacts with the filter input of a Dropdown in cell editing mode, the editor should remain open until the user makes a selection or explicitly moves focus away from the cell (e.g., by clicking outside the Dropdown).

### LINKED ISSUE #7405: Datatable: CellEdit and general issues with portal editors

### LINKED ISSUE #8080: Dropdown: Item list immediately closes after opening
## Issue Description
When using the Dropdown as a cell editor in the DataTable, the options list closes immediately after being opened.

### Root Cause
This issue occurs because the `isOutsideClicked()` method incorrectly returns `true`, triggering the `switchCellToViewMode()` function which closes the editor and the dropdown list. This happens because the down arrow SVG icon inside the Dropdown is replaced by the up arrow SVG (to indicate the list is expanded), causing the down arrow SVG to be detached from the DOM, making it appear as though a click occurred outside the DataTable.

### Additional Notes

- This behavior is also inconsistent due to a race condition - if the `isOutsideClicked()` function is executed before the down arrow SVG is replaced, the Dropdown opens and behaves correctly.
- When you click on the arrow's container instead of clicking directly on the arrow image (i.e. you click on the surrounding margin), it works correctly.

### Reproducer

https://stackblitz.com/edit/8apgvuc5-zrqniw2f?file=src%2FApp.jsx

### System Information

```Shell
primereact: 10.8.5 => 10.9.6
```

### DEMO

https://github.com/user-attachments/assets/86f493f1-3d39-44db-914e-5e546263405e

### Expected behavior

When using the Dropdown as a cell editor in the DataTable:

- The options list should remain open when clicked and allow the user to interact with the options, without closing prematurely.
- The Dropdown interaction should work consistently and not be impacted by DOM changes to the arrow SVG.

### Quick fix
I do not think it is the right approach but the quick fix is to apply the following style:
```css
.pp-dropdown-trigger-icon {
  pointer-events: none;
}
```

gnawjaren:
### Defect Fixes
Simplify overlay handling in cell edit mode

This PR simplifies the current overlay handling logic in cell edit mode by introducing a declarative approach.
Instead of relying on OverlayService, selfClick, and setTimeout workarounds, it checks for a data-pr-is-overlay attribute on clicked elements (or their parents) to determine whether an outside click should close the cell editor.

Benefits:
	•	Greatly reduces complexity and improves maintainability
	•	More robust and easier to debug
	•	Enables support for custom or third-party overlay editors, including full interaction, as long as they include the data-pr-is-overlay attribute on their container
	•	Makes the logic more declarative and extensible

I tested this in various parts of our project with good results.

Let me know if this approach is acceptable – I’d be happy to adjust the implementation if needed.
Fix #7405 
Fix #7158
Fix #8080

### DIFF
--- a/components/lib/autocomplete/AutoCompletePanel.js
+++ b/components/lib/autocomplete/AutoCompletePanel.js
@@ -268,7 +268,8 @@ export const AutoCompletePanel = React.memo(
                 {
                     className: classNames(props.panelClassName, cx('panel', { context })),
                     style,
-                    onClick: (e) => props.onClick(e)
+                    onClick: (e) => props.onClick(e),
+                    'data-pr-is-overlay': true
                 },
                 _ptm('panel')
             );
--- a/components/lib/calendar/CalendarPanel.js
+++ b/components/lib/calendar/CalendarPanel.js
@@ -18,7 +18,8 @@ export const CalendarPanel = React.forwardRef((props, ref) => {
                 'aria-label': localeOption('chooseDate', props.locale),
                 'aria-modal': props.inline ? null : 'true',
                 onClick: props.onClick,
-                onMouseUp: props.onMouseUp
+                onMouseUp: props.onMouseUp,
+                'data-pr-is-overlay': true
             },
             props.ptm('panel', { hostName: props.hostName })
         );
--- a/components/lib/colorpicker/ColorPickerPanel.js
+++ b/components/lib/colorpicker/ColorPickerPanel.js
@@ -14,7 +14,8 @@ export const ColorPickerPanel = React.forwardRef((props, ref) => {
             {
                 className: cx('panel', { panelProps: props, context }),
                 style: props.panelStyle,
-                onClick: props.onClick
+                onClick: props.onClick,
+                'data-pr-is-overlay': true
             },
             ptm('panel', { hostName: props.hostName })
         );
--- a/components/lib/datatable/BodyCell.js
+++ b/components/lib/datatable/BodyCell.js
@@ -8,7 +8,6 @@ import { ChevronDownIcon } from '../icons/chevrondown';
 import { ChevronRightIcon } from '../icons/chevronright';
 import { PencilIcon } from '../icons/pencil';
 import { TimesIcon } from '../icons/times';
-import { OverlayService } from '../overlayservice/OverlayService';
 import { Ripple } from '../ripple/Ripple';
 import { DomHandler, IconUtils, ObjectUtils, classNames } from '../utils/Utils';
 import { RowCheckbox } from './RowCheckbox';
@@ -21,8 +20,6 @@ export const Cell = (props) => {
     const [styleObjectState, setStyleObjectState] = React.useState({});
     const elementRef = React.useRef(null);
     const keyHelperRef = React.useRef(null);
-    const overlayEventListener = React.useRef(null);
-    const selfClick = React.useRef(false);
     const focusTimeout = React.useRef(null);
     const initFocusTimeout = React.useRef(null);
     const editingRowDataStateRef = React.useRef(null);
@@ -58,9 +55,9 @@ export const Cell = (props) => {
     };
 
     const isIgnoredElement = (element) => {
-        const isCellEditor = (el) => el.getAttribute && el.getAttribute('data-pr-is-overlay');
+        const isOverlay = (el) => el.getAttribute && el.getAttribute('data-pr-is-overlay');
 
-        return isCellEditor(element) || DomHandler.getParents(element).find((el) => isCellEditor(el));
+        return isOverlay(element) || DomHandler.getParents(element).find((el) => isOverlay(el));
     };
 
     const [bindDocumentClickListener, unbindDocumentClickListener] = useEventListener({
@@ -69,8 +66,6 @@ export const Cell = (props) => {
             if (!isIgnoredElement(e.target) && isOutsideClicked(e.target)) {
                 switchCellToViewMode(e, true);
             }
-
-            selfClick.current = false;
         },
         options: true,
         when: isEditable()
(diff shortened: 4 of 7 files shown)


gnawjaren:
It appears that all code related to overlayService and the selfClickRef is only used in this specific context. I would update this PR to remove that logic—unless this general approach is not acceptable. Otherwise, I’ll push the remaining changes shortly.

melloware:
@gnawjaren may as well push all your changes.

melloware:
Conflicts now. Is this PR still needed?

melloware:
OK i will leave it pending review.  But you should fix the merge conflicts.

melloware:
@gnawjaren can you fix the merge conflicts?  I have also linked this PR to that new ticket.

gnawjaren:
I’ve rebased the branch and resolved the conflicts. I also re-tested the changes in our project.
The main question now is whether this is the best place to add the attributes – or if it would make sense to include them in other components as well.
Currently, only Dropdown and Calendar are affected, but extending it to more components should be straightforward.

melloware:
@gnawjaren  yes please i think so this does simplify and clean up the handling.

melloware (approved):

howkymike:
@gnawjaren could you please fix the merge conflicts once again? 

Let's merge this long awaiting change to the next release :)

howkymike:
@melloware Is there anything more we can do here or can we merge it?

melloware (approved):

mertsincan:
I don’t think this is the right solution. I believe DomHandler.getParents will affect the table’s performance. Also, OverlayService is an exported eventbus, so users can integrate a select or similar component from a different library using this OverlayService.on('overlay-click')

gnawjaren:
I understand your concerns about performance.
The primary motivation for this change is to achieve fully predictable event ordering.
With the current OverlayService approach, we need to rely on setTimeout,
which can cause race conditions and subtle bugs — especially when browser and custom events are mixed.

getParents only runs on document clicks and only when an overlay is open,
so in practice the performance impact should be negligible.

The improved third-party overlay support is more of a side benefit —
the main goal here is removing the setTimeout dependency
to make the behavior more robust and maintainable.

Ultimately, I believe this trade-off is worth it,
but of course we can revert if you decide otherwise.

PR #7973, opened 2025-05-05
DataTable: Fix current implementation of overlay handling in cell edit mode

### LINKED ISSUE #7405: Datatable: CellEdit and general issues with portal editors

gnawjaren:
### Defect Fixes
Fixes #7405 
This also resolves the overlay issue by ensuring isOutsideClicked is passed to BodyRow.onCellClick.

Note: While debugging, I noticed that in link mode, some shared modules like OverlayService get bundled multiple times due to the current Rollup config. This led to unexpected behavior until I applied a local workaround to force a shared instance.

### DIFF
--- a/components/lib/datatable/BodyCell.js
+++ b/components/lib/datatable/BodyCell.js
@@ -176,7 +176,7 @@ export const Cell = (props) => {
     };
 
     const onClick = (event) => {
-        props.onClick(event, getCellCallbackParams(event), isEditable(), editingState, setEditingState, selfClick, props.column, bindDocumentClickListener, overlayEventListener);
+        props.onClick(event, getCellCallbackParams(event), isEditable(), editingState, setEditingState, selfClick, props.column, bindDocumentClickListener, overlayEventListener, isOutsideClicked);
     };
 
     const onMouseDown = (event) => {
--- a/components/lib/datatable/BodyRow.js
+++ b/components/lib/datatable/BodyRow.js
@@ -534,7 +534,7 @@ export const BodyRow = React.memo((props) => {
         }
     }, []);
 
-    const onCellClick = (event, params, isEditable, editingState, setEditingState, selfClick, column, bindDocumentClickListener, overlayEventListener) => {
+    const onCellClick = (event, params, isEditable, editingState, setEditingState, selfClick, column, bindDocumentClickListener, overlayEventListener, isOutsideClicked) => {
         if (props.editMode !== 'row' && isEditable && !editingState && (props.selectOnEdit || (!props.selectOnEdit && props.isRowSelected))) {
             selfClick.current = true;
 


melloware:
@gnawjaren in Linked mode you mean while debugging right but this does not happen in Production Packaged mode right?

melloware (approved):

gnawjaren:
Yes, it only happens when locally linking and running `npm run dev:link`. 
In that mode, the Rollup config bundles all shared modules (like OverlayService) into each component separately, so singletons no longer work as expected. This doesn’t happen in the production build.

melloware:
Does this PR need to go with your other one or is this standalone and fixes an issue?

gnawjaren:
This is standalone. They both fix it, but this one keeps everything else as it is.

PR #7986, opened 2025-05-09
Fix: #7984 (Autocomplete type issue) when multiple and Templates are used

### LINKED ISSUE #7984: AutoComplete: Bad type when using custom template definitions
### Describe the bug

The AutoComplete component will return the wrong type in "value" and "onChange" props when some other props are used.
Instead of using type in "suggestions" props, it'll use the type of templating functions.
So when you use multiple value, instead of returning Array type, it's object type and the compiler throw Error.

I found at least 3 functions that are used to define the type in the wrong way:
selectedItemTemplate > itemTemplate > onSelect

Type of returning values must be defined by the combinaison of "suggestions" props and "multiple" props. 

### Reproducer

https://stackblitz.com/edit/14tqd2rj?file=src%2FApp.tsx

### System Information

```Shell
System:
    OS: Windows 11 10.0.26100
    CPU: (12) x64 AMD Ryzen 5 3600 6-Core Processor              
    Memory: 8.18 GB / 31.93 GB
  Binaries:
    Node: 22.12.0 - C:\Program Files\nodejs\node.EXE
    npm: 10.9.0 - C:\Program Files\nodejs\npm.CMD
    pnpm: 10.8.0 - C:\Program Files\nodejs\pnpm.CMD
  Browsers:
    Edge: Chromium (130.0.2849.80)

    "primereact": "10.9.5",
    "react": "^18.2.0",
```

1. Init AutoComplete componant.
2. Use Multiple props
3. Add selectedItemTemplate || itemTemplate || onSelect props
4. The value an onChange props will be on error, with a non Array type

### Expected behavior

Value and onChange has to be Array of suggestions type[number]

GuillaumeC:
[AutoComplete: Bad type when using custom template definitions](https://github.com/primefaces/primereact/issues/7984)
Fix #7984

### DIFF
--- a/components/lib/autocomplete/autocomplete.d.ts
+++ b/components/lib/autocomplete/autocomplete.d.ts
@@ -230,7 +230,7 @@ export interface AutoCompleteContext {
  * Defines valid properties in AutoComplete component. In addition to these, all properties of HTMLSpanElement can be used in this component.
  * @group Properties
  */
-export interface AutoCompleteProps<T = any> extends Omit<React.DetailedHTMLProps<React.InputHTMLAttributes<HTMLSpanElement>, HTMLSpanElement>, 'onChange' | 'onSelect' | 'ref' | 'value'> {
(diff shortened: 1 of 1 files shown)


melloware:
@GuillaumeC have you verified this fixes your issues?

PR #7999, opened 2025-05-14
fix: prevent Enter key from clearing selection when no option focused

### LINKED ISSUE #7850: Dropdown: Press enter key without selecting and option triggers onChange event
### Describe the bug

After type some text in filter section, but not selecting any option using down key if we press Enter key, then it triggers the onChange event with undefined values.

### Reproducer

https://stackblitz.com/run?file=src%2FApp.jsx

### System Information

```Shell
System:
    OS: Windows 10 10.0.19045
    Memory: 17.13 GB / 31.93 GB
  Binaries:
  Browsers:
  npmPackages:
    primereact: ^10.9.3 => 10.9.3
    react: ^18.2.0 => 18.2.0
```

1. select an option Brazil from dropdown.
2. now try to type some random text and you will see not results displayed in the dropdown list.
3. Now If you press Enter, it removes the already selected value and trigger onChange event with undefined value.

### Expected behavior

It should not remove the already selected values and trigger onChange event if no option highlighted from the dropdown list.

codiini:
This PR fixes #7850

### DIFF
--- a/components/lib/dropdown/Dropdown.js
+++ b/components/lib/dropdown/Dropdown.js
@@ -513,7 +513,7 @@ export const Dropdown = React.memo(
                     hide();
                     resetFilter();
                     updateEditableLabel(selectedOption);
-                    
+
                     return;
                 }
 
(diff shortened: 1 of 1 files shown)


codiini:
@melloware Nice catch. I've updated the PR to reflect this. Let me know if it's A-OK

melloware:
running the build now.

PR #8009, opened 2025-05-18
feat(MultiStateCheckbox): add id prop to improve accessibility

### LINKED ISSUE #7997: MultiStateCheckbox: Missing id prop
### Describe the bug

The MultiStateCheckbox component does not expose or allow passing an id/inputId prop, making it impossible to link a <label> to the checkbox for accessibility (using htmlFor). As a result, clicking on the label does not focus or toggle the checkbox.

### Reproducer

_No response_

### System Information

```Shell
Browsers:
    Edge: Chromium (131.0.2903.63)
  npmPackages:
    primereact: ^10.8.5 => 10.8.5
    react: ^18.2.0 => 18.2.0
    tailwindcss: ^3.4.4 => 3.4.4
```

_No response_

### Expected behavior

_No response_

timmy471:
This PR fixes #7997

### DIFF
--- a/components/lib/multistatecheckbox/MultiStateCheckbox.js
+++ b/components/lib/multistatecheckbox/MultiStateCheckbox.js
@@ -170,6 +170,7 @@ export const MultiStateCheckbox = React.memo(
 
         const checkboxProps = mergeProps(
             {
+                id: props.id,
                 className: classNames(props.className),
                 style: sx('checkbox', { selectedOption }),
                 tabIndex: props.tabIndex,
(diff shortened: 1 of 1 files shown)


components/lib/multistatecheckbox/MultiStateCheckbox.js
```diff
@@ -170,6 +170,7 @@ export const MultiStateCheckbox = React.memo(
 
         const checkboxProps = mergeProps(
             {
+                id: props.id,
```
melloware, line 173: @timmy471 but the `id` is alreayd put on the div wrapper of the checkbox.  Its bad HTML practice to have 2 id's in the same DOM.  So maybe this should be `props.id + '_checkbox'` ???

melloware (changes requested):

components/lib/multistatecheckbox/MultiStateCheckbox.js
timmy471: Got that! Thank you @melloware

melloware (approved):

PR #8050, opened 2025-05-28
Fix MultiSelect non-unique keys

### LINKED ISSUE #8049: MultiSelect has non-unique key warning
### Describe the bug

MultiSelect has non-unique key warning. It is reproducible from its documentation page. See https://primereact.org/multiselect/#chips.

Image

### Reproducer

https://primereact.org/multiselect/#chips

### System Information

```Shell
-
```

1. Go to the [MultiSelect documentation page](https://primereact.org/multiselect/)
2. Go to the Chips section
3. Click on the component to see the options
4. Check the console to see following warning:

```
Warning: Each child in a list should have a unique "key" prop. See https://reactjs.org/link/warning-keys for more information. Error Component Stack
[...]
```

### Expected behavior

No warnings.

rubenochiavone:
Fixes #8049.

### DIFF
--- a/components/lib/multiselect/MultiSelectItem.js
+++ b/components/lib/multiselect/MultiSelectItem.js
@@ -78,7 +78,7 @@ export const MultiSelectItem = React.memo((props) => {
     );
 
     return (
-        <li {...itemProps} key={props.index + '_multiselectitem'}>
+        <li {...itemProps}>
             <div {...checkboxContainerProps}>
                 <Checkbox ref={checkboxRef} checked={props.selected} icon={checkboxIcon} pt={ptm('checkbox')} unstyled={isUnstyled()} tabIndex={-1} />
             </div>


melloware:
Weird shouldn't this key make it unique since its using `index` ?

melloware (changes requested):
Please explain why this fixes it?

rubenochiavone:
Here is the call site of `MultiSelectItem` in `MultiSelectPanel`'s `createItem` function:

https://github.com/primefaces/primereact/blob/87356330dbfe70ab8d38bcb1332ad188f6b0f021/components/lib/multiselect/MultiSelectPanel.js#L170-L188

Although `MultiSelectItem` returns a key in every render, it's not being used here and is causing this warning. If we remove the `key` from `MultiSelectItem` usage (line 172 above), I'd expect `MultiSelectItem` returned key was being used, but the non-unique warning still happens - which means it wasn't used. So, currently I don't see any reasonable use for keeping it. Technically, the key being return by `MultiSelectItem` is not wrong, though, I wonder if could be an issue with the react version being used.

Anyway, feel free to close the PR if you think is not worth it.

melloware:
yeah that is why i was confused i wonder if its a false positive React warning?

rubenochiavone:
> yeah that is why i was confused i wonder if its a false positive React warning?

Yeah, probably.

Issue #8066, opened 2025-06-03
Tree keyboard navigation not working as expected


radu-iamandi-visma:
### Describe the bug

According to my understanding, the [ARIA requirements](https://www.w3.org/WAI/ARIA/apg/patterns/treeview/) say that, when the node is closed, pressing RIGHT should open the children, but not change focus. 

Currently, pressing RIGHT open the children and changes focus to the first child.

### System Information

```Shell
Irrelevant
```

### Steps to reproduce the behavior

1. Tab to focus a Tree item
2. Press Arrow Right

### Expected behavior

3. Children are exposed
4. Focus not changed

PR #8067, opened 2025-06-03
fix Tree keyboard navigation when item expands #8066

radu-iamandi-visma:
According to my understanding, the [ARIA requirements](https://www.w3.org/WAI/ARIA/apg/patterns/treeview/) say that, when the node is closed, pressing RIGHT should open the children, but not change focus. 

Currently, pressing RIGHT open the children and changes focus to the first child.

Issue:  #8066

**Generated**: 
This pull request modifies the behavior of the `onArrowRight` function in the `UITreeNode` component to improve handling of tree node expansion.

Key changes in tree node behavior:

* [`components/lib/tree/UITreeNode.js`](diffhunk://#diff-54e22c623d60c2f02035d9621049e7d57f7673967b080e37c0ae6179feff6266L433-R439): Updated the `onArrowRight` function to ensure that the `expand` method respects the current `expanded` state, and removed the condition that checked for `expanded` before returning early.

### DIFF
--- a/components/lib/tree/UITreeNode.js
+++ b/components/lib/tree/UITreeNode.js
@@ -430,13 +430,13 @@ export const UITreeNode = React.memo((props) => {
     };
 
     const onArrowRight = (event) => {
-        if (isLeaf || expanded) {
+        if (isLeaf) {
             return;
         }
 
         event.currentTarget.tabIndex = -1;
 
-        expand(event, true);
+        expand(event, expanded);
     };
 
     const onArrowLeft = (event) => {


melloware:
PrimeReact is doing the same as PrimeVue: https://primevue.org/tree/#accessibility

radu-iamandi-visma:
> PrimeReact is doing the same as PrimeVue: https://primevue.org/tree/#accessibility

Not here :(
https://primereact.org/tree/#checkboxselection

PR #8068, opened 2025-06-04
Fix: ToggleButton accessibility attributes

### LINKED ISSUE #8021: ToggleButton - Accessibility bug
### Describe the bug

found two accessibility issues with ToggleButton component

1. ARIA attribute is not allowed: aria-pressed="false"
2. Ensure every form element has a label

Image

### Reproducer

_No response_

### System Information

```Shell
primereact version - 10.9.3
```

_No response_

### Expected behavior

_No response_

tneeraja95:
Fix #8021 

Fixes -

1. Added `aria-label` and `aria-labelledby` from `props` to the `ToggleButton` input for accessible naming. 
2. Removed `role` and `aria-pressed `from the `input` element, as they are not applicable to `type="checkbox"` and can create conflicting semantics.

### DIFF
--- a/components/lib/togglebutton/ToggleButton.js
+++ b/components/lib/togglebutton/ToggleButton.js
@@ -125,7 +125,8 @@ export const ToggleButton = React.memo(
                 readOnly: props.readonly,
                 value: props.checked,
                 checked: props.checked,
-                'aria-label': label
+                'aria-label': props['aria-label'],
+                'aria-labelledby': props['aria-labelledby']
             },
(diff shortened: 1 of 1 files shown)


melloware:
Here is what PrimeVue does: https://primevue.org/togglebutton/#accessibility  Can you confirm we do the same with this PR?

tneeraja95:
Updated the PR. Using `aria-label` and `aria-labelledby` from `props`.

melloware (approved):

PR #8105, opened 2025-06-27
fix:#8099:DataTable: dissapearing values when clicking quickly

### LINKED ISSUE #8099: DataTable: dissapearing values when clicking quickly
### Describe the bug

I am using a DataTable with Dropdowns as cell editors in "cell" edit mode. When clicking very quickly across many different cells, the cell values sometimes become empty.

After debugging the exact point when null value is set, I observed the following behavior:
- The Dropdown's `onChange` event is not triggered.
- The `onCellEditComplete` event of the column returns `event.newValue` as `null`.
Side note: It happens when I click very fast mulitple times so the first `onChange` event is triggered (value is changed correctly) and then the mouse click event triggers events once more (this time without `onChange`, only onCellEditComplete and onEditorSubmit with null values).

So the execution path is like this:
1. I open a cell editor and change its value.
2. I click multiple times on another cell.
3. Both the dropdown's `onChange` and the column's `onCellEditComplete` events are triggered successfully.
4. The value is correctly updated to the newly set value.
5. Immediately after this, a new `onCellEditComplete` event is executed with `newValue` equal to `null`.
6. The cell value is incorrectly changed to `null`.

Upon investigating the library further, I noticed the root issue seems to stem from the following:
1. `newValue` and `newRowData` are `null` because `editingRowDataStateRef.current` is `null`
2. There is a `setTimeout` operation that sets `editingRowDataStateRef.current` to `null` after 1 ms [**or potentially later**](https://developer.mozilla.org/en-US/docs/Web/API/Window/setTimeout#:~:text=Note%20that%20in%20either%20case%2C%20the%20actual%20delay%20may%20be%20longer%20than%20intended%3B%20see%20Reasons%20for%20delays%20longer%20than%20specified%20below.).

### Possible explanation

Here's possible explanation of what happens step by step:
1. I click on a cell, opening its editor.
2. I quickly click on another cell, triggering the `switchCellToViewMode` function to close the previous editor.
3. The `switchCellToViewMode` function invokes `onCellEditComplete` and then `closeCell` (which sets the timeout to 1ms) functions.
4. Before the timeout is executed, the new cell editor is opened, updating `editingRowDataStateRef.current` to the current cell editor.
5. After the 1 ms timeout, `editingRowDataStateRef.current` is reset to null - incorrect behavior, as it overwrites the currently active editor state.
6. I click outside of the cell, triggering the `switchCellToViewMode` function again. This time, `editingRowDataStateRef.current` is null, and `onCellEditComplete` receives null values for `event.newValue` and `event.newRowData`.

#### Problematic file 

[BodyCell.js](https://github.com/primefaces/primereact/blob/master/components/lib/datatable/BodyCell.js)

### Possible faulty commit
This issue might have been introduced by the following commit: [ecd37b65b711331cb4f839c9fb85059e875fde9d](https://github.com/primefaces/primereact/commit/ecd37b65b711331cb4f839c9fb85059e875fde9d)

@kl-nevermore, could you please review this?

### Reproducer

This is hard to reproduce because it occurs only when the application is huge enough to force `setTimeout` to invoke a lot later.

In my app `setTimeout` function can execute 20ms before another `switchCellToViewMode` function call so it breaks the functionality.

I have found the following code example (from [#5122](https://github.com/primefaces/primereact/issues/5122)) where something similar happens (it throws an error because there is .trim methd call on null value, but if it wasn't, the value would be blank): https://codesandbox.io/p/sandbox/elastic-yalow-hdlfzx?file=%2Fsrc%2FApp.tsx

https://github.com/user-attachments/assets/8ce2bf2f-e643-449a-b2f1-ce9ffc998469

### System Information

```Shell
primereact: 10.9.6
```

1. Create a DataTable with lot's of dropdowns as cell editors in "cell" edit mode.
2. Click very quickly across different cells.
3. Observe that occasionally, the values of certain cells disappear, and the dropdown's `onChange` and column's `onCellEditComplete` events do not fire correctly.

### Expected behavior

The values should not disappear when clicking around.

kl-nevermore:
Fix #8099

### DIFF
--- a/components/lib/datatable/BodyCell.js
+++ b/components/lib/datatable/BodyCell.js
@@ -367,6 +367,9 @@ export const Cell = (props) => {
             OverlayService.off('overlay-click', overlayEventListener.current);
             overlayEventListener.current = null;
         }
+        if (editingRowDataStateRef.current) {
+            editingRowDataStateRef.current = null;
+        }
     });
 
     const createLoading = () => {
--- a/components/lib/datatable/BodyCell.js
+++ b/components/lib/datatable/BodyCell.js
@@ -122,7 +122,6 @@ export const Cell = (props) => {
             unbindDocumentClickListener();
             OverlayService.off('overlay-click', overlayEventListener.current);
             overlayEventListener.current = null;
-            editingRowDataStateRef.current = null;
             selfClick.current = false;
         }, 1);
     };
--- a/components/lib/datatable/BodyCell.js
+++ b/components/lib/datatable/BodyCell.js
@@ -367,6 +367,7 @@ export const Cell = (props) => {
             OverlayService.off('overlay-click', overlayEventListener.current);
             overlayEventListener.current = null;
         }
+
         if (editingRowDataStateRef.current) {
             editingRowDataStateRef.current = null;
         }


melloware (approved):

pratyaksh123:
Any timeline on when this will be released?

pratyaksh123:
https://github.com/user-attachments/assets/d8d633b0-1c92-4283-834a-f48886f3dc81

I patched the package locally and I am still seeing the same issue  @kl-nevermore

kl-nevermore:
could you test again on master branch？

pratyaksh123:
I am using your changes from master by patching the library using pnpm ( from changes of this PR)  , still same issue

https://github.com/primefaces/primereact/issues/8106
also not sure if this is related.

kl-nevermore:
https://github.com/user-attachments/assets/9607f5d2-7ce5-473c-bd09-74ac32cd867d

I retested it on the master branch, and it looks fine.

https://github.com/primefaces/primereact/issues/8106#issuecomment-3053822761 can try it

pratyaksh123:
I tested in the master branch now, looks like it's working! Thanks a lot 
Do you know when the next version will be released?

kl-nevermore:
I don't know, you could try asking in Discord

dmitrevnik:
@pratyaksh123 

it seems there is almost another approach on master branch. in my project i already use 10.9.6 in the most of the components. can you explain how you fix this issue plz?

pratyaksh123:
@dmitrevnik  Go to https://github.com/primefaces/primereact/blob/master/DEVELOPMENT.md and use yalc to build and install the package locally. I was able to test the master branch in my project, and this issue was resolved!

pratyaksh123:
The disord isnt active , last release was june 2 , will this be released soon?

melloware:
I pinged PrimeTek to do a release

PR #8135, opened 2025-07-12
Fix #8052 year not updated with viewDate changes

### LINKED ISSUE #8052: Calendar year not updated with viewDate changes
### Describe the bug

when viewDate changes, only dates get updated while the year in the header is not. it should update the year in the header when dates gets updated.

### Reproducer

https://stackblitz.com/edit/ht8rm1ve-hndxa4gw?file=src%2FApp.tsx

### System Information

```Shell
System:
    OS: Linux 5.0 undefined
    CPU: (8) x64 Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
    Memory: 0 Bytes / 0 Bytes
    Shell: 1.0 - /bin/jsh
  Binaries:
    Node: 20.19.1 - /usr/local/bin/node
    Yarn: 1.22.19 - /usr/local/bin/yarn
    npm: 10.8.2 - /usr/local/bin/npm
    pnpm: 8.15.6 - /usr/local/bin/pnpm
  npmPackages:
    primereact: latest => 10.9.5 
    react: 18.3.1 => 18.3.1
```

1. select January 1st , 2000 and delete the date in the input field. 
2. open calendar again
3. calendar renders the dates for May 2007. but the header is May 2000, which suppose to May 2007 
4. if you select the date in the calendar, will take in 2007. not 2000. 

### Expected behavior

1. select January 1st , 2000 and delete the date in the input field. 
2. open calendar again
3. calendar dates are for May 2007 and also the header should be May 2007
4. select date in the calendar, should reflect the value that user sees and chooses

### LINKED ISSUE #8258: Calendar: updateViewDate does not update the year in the title
### Describe the bug

When using `updateViewDate(null, date)` the dates in the overlay change but the title does not update.

### Reproducer

https://stackblitz.com/edit/pkpkq47a-ambhsfzj?file=src%2FApp.jsx

### System Information

```Shell
System:
    OS: Linux 5.0 undefined
    CPU: (8) x64 Intel(R) Core(TM) i9-9880H CPU @ 2.30GHz
    Memory: 0 Bytes / 0 Bytes
    Shell: 1.0 - /bin/jsh
  Binaries:
    Node: 20.19.1 - /usr/local/bin/node
    Yarn: 1.22.19 - /usr/local/bin/yarn
    npm: 10.8.2 - /usr/local/bin/npm
    pnpm: 8.15.6 - /usr/local/bin/pnpm
  npmPackages:
    primereact: latest => 10.9.7 
    react: 18.3.1 => 18.3.1
```

1. Open the reproducer
2. Click on the custom header buttons 
3. The dates in the overlay update correctly, however, the title remains the same

I believe this might have been introduced with https://github.com/primefaces/primereact/commit/92de1ebd87f0c3937c6f7da8a5dce8d408198aa4 

### Expected behavior

The title is updated to match the view date that is set via `updateViewDate()`

ulgyd0gg:
Fix #8052 
Fix #8258
The issue occurred when we set the view date and deleted the input field. The currentYear variable retained the old value (in this case, 2000).😶‍🌫️
https://github.com/primefaces/primereact/blob/d57bbb7e4f57371e133668b42db3dc1435191f10/components/lib/calendar/Calendar.js#L3391-L3393
I adjusted displayYear to use metaYear, which fixes the issue, and the yearNavigator works correctly. However, if we modify the value of InputRef.current.value in a way that it cannot be parsed as a Date, the rendering will fall back to using ViewDate as the parameter.😺

### DIFF
--- a/components/lib/calendar/Calendar.js
+++ b/components/lib/calendar/Calendar.js
@@ -3390,7 +3390,6 @@ export const Calendar = React.memo(
         const createTitleYearElement = (metaYear) => {
             const viewDate = getViewDate();
             const viewYear = viewDate.getFullYear();
-            const displayYear = props.numberOfMonths > 1 || props.yearNavigator ? metaYear : currentYear;
 
             if (props.yearNavigator) {
                 let yearOptions = [];
@@ -3416,7 +3415,7 @@ export const Calendar = React.memo(
                     {
                         className: cx('select'),
                         onChange: (e) => onYearDropdownChange(e, e.target.value),
-                        value: displayYear
+                        value: metaYear
                     },
                     ptm('select')
                 );
@@ -3468,7 +3467,7 @@ export const Calendar = React.memo(
                 ptm('yearTitle')
             );
 
-            return currentView !== 'year' && <button {...yearTitleProps}>{displayYear}</button>;
+            return currentView !== 'year' && <button {...yearTitleProps}>{metaYear}</button>;
         };
 
         const createTitleDecadeElement = () => {


ulgyd0gg:
Fix #8052

melloware:
@ulgyd0gg unfortunately this fix for year just keeps going back and forth. See the list of tickets in the past for "year not updating" issues: https://github.com/primefaces/primereact/pulls?q=is%3Apr+year+is%3Aclosed

I think your fix is just putting it back how it was and fixed a couple of scenarios but breaks a few more.

Look at the Git Blame on those lines of code to see last ticket and fix

ulgyd0gg:
@melloware I thought about these during the changes. What’s the desired behavior—should currentValue become null if value is empty, defaulting to currentValue || metaYear? Or should we always use the last selected value? I’d like the core team’s input on this.

ulgyd0gg:
This is behavior after change. I set `newDate()` as `viewDate`
CalendarViewDate

melloware:
@ulgyd0gg i have fixed this back and forth over the years and i am not sure what the right solution is. I have fixed it for people to fix one scenario and then another bug gets reported.  Ideally we would have a list of of all test cases and make sure they all work.

PR #8146, opened 2025-07-15
fix(button): make icon inside Button clickable in OrderList itemTemplate (#8145)

### LINKED ISSUE #8145: Button: When using a Button inside of a OrderList itemTemplate component, only outside of icon is clickable
### Describe the bug

I'm upgrading PrimeReact from 10.6.6 to 10.9.6, I have an OrderList where I am providing a itemTemplate component. I have 2 icon buttons in my itemTemplate component and for some reason, after updating PrimeReact, the actual Icon portion of the buttons is no longer clickable, the area around the icon is clickable, but attempting to click the actual icon does not fire.

### Reproducer

https://stackblitz.com/edit/vitejs-vite-5ife661w?file=src%2FApp.tsx

### System Information

```Shell
primereact: 10.6.6 => 10.9.6
```

1. Click the button at the icon
2. Check the console for the log (does not fire)
3. Click the button outside the icon but still in the clickable area
4. Check the console for the log (does fire)

### Expected behavior

It should fire for both clicks

karan74jaiswal:
Fixes #8145

**Root Cause:**  
When using a Button inside an OrderList itemTemplate, only the area outside the icon was clickable. This was because the icon element inside the Button was intercepting pointer events, so clicks on the icon did not bubble up to the button element.

**Solution:**  
I set `pointer-events: none` on the icon element (`.p-button-icon`) inside the Button component. This ensures that all clicks on the icon are passed through to the button, making the entire button (including the icon) clickable.

**Change Details:**  
- Added `style: { pointerEvents: 'none' }` to the icon props in the Button component.

This change makes the Button's `onClick` handler fire regardless of where the user clicks inside the button, including directly on the icon.

### DIFF
--- a/components/lib/button/Button.js
+++ b/components/lib/button/Button.js
@@ -45,7 +45,8 @@ export const Button = React.memo(
 
             const iconsProps = mergeProps(
                 {
-                    className: cx('icon')
+                    className: cx('icon'),
+                    style: { pointerEvents: 'none' }
                 },
                 ptm('icon')
             );
--- a/components/lib/card/__snapshots__/Card.spec.js.snap
+++ b/components/lib/card/__snapshots__/Card.spec.js.snap
@@ -56,6 +56,7 @@ exports[`Card advanced 1`] = `
             <span
               class="p-button-icon p-c p-button-icon-left pi pi-check"
               data-pc-section="icon"
+              style="pointer-events: none;"
             />
             <span
               class="p-button-label p-c"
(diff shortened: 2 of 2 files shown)


melloware:
@karan74jaiswal can you review and test my PR please?  https://github.com/primefaces/primereact/pull/8147

It fixes it without adding a custom style on the component.

karan74jaiswal:
@melloware I am also a contributer not a core team member. Both approaches do the same things. I do not have right to test your pr.

melloware:
you can just remove your change and make the same change as mine in your local branch and test it. Its a small change

karan74jaiswal:
@melloware I tested your approach. Its doing the right thing. I tried to reproduce the issue, and got that bug is fixed.

melloware:
Thanks for testing!

PR #8181, opened 2025-07-30
Expose `processedData` in `DataTable`

makeeno:
This PR exposes the `processedData()` method in the `DataTable` component.

The purpose of this change is to enable the implementation of custom export functionality that respects the currently applied filters. Currently, the built-in `exportCSV()` method uses `processedData()` internally to retrieve filtered and sorted data. By making this method accessible, custom exporters can achieve the same behavior and maintain consistency with the built-in export logic.

Old but still relevant discussion here: https://github.com/orgs/primefaces/discussions/23


melloware (approved):

melloware:
I am not sure this is the best name for the exposed method?   I am trying to think of a better one but can't!

PR #8203, opened 2025-08-07
Fix #8202: Dropdown & MultiSelect prevent option focus on reopen when autoOptionFocus is false

### LINKED ISSUE #8202: Dropdown & MultiSelect: autoOptionFocus={false} is ignored when reopening panel
### Describe the bug

Hi,
There’s an issue in both Dropdown and MultiSelect components when using:

```
focusOnHover={false}
autoOptionFocus={false}
```
On first open, no option is focused as expected.
But when the panel is closed and reopened, one option becomes focused, which violates the expected behavior of those props.

### Reproducer

https://stackblitz.com/edit/ueygcbfb?file=src%2FApp.jsx

### System Information

```Shell
primereact: latest
```

1. Open the panel → ✅ no item focused.
2. Select an item.
3. Close the panel.
4. Open the panel again, an option is focused automatically.

### Expected behavior

If autoOptionFocus={false}, no option should be focused, even after reopening the panel.
setFocusedOptionIndex should be set to -1 in this case.

sergiomendi:
### Defect Fixes
This PR fixes an issue in `Dropdown` and `MultiSelect` where an option receives focus after reopening the panel, even when `autoOptionFocus={false}` 

Fix #8202.

### Feature Requests

### DIFF
--- a/components/lib/dropdown/Dropdown.js
+++ b/components/lib/dropdown/Dropdown.js
@@ -788,7 +788,7 @@ export const Dropdown = React.memo(
         };
 
         const show = () => {
-            setFocusedOptionIndex(focusedOptionIndex !== -1 ? focusedOptionIndex : props.autoOptionFocus ? findFirstFocusedOptionIndex() : props.editable ? -1 : findSelectedOptionIndex());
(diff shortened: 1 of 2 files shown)


mertsincan:
This PR breaks the current default behavior. autoOptionFocus is meant to decide whether to focus when there’s no selected item. Otherwise, the selected item should always be automatically focused.

PR #8307, opened 2025-09-29
fix: enable single-field editing in OTP input

### LINKED ISSUE #8306: InputOtp:  Unable to edit single field input
### Describe the bug

If you move with the cursor or by clicking to another https://primereact.org/inputotp/ input field you can't overwrite it even if it is selected. You have to press backspace, this will delete the current field, but also move to the field before, so after backspace you have to move the cursor back to the right.

### Reproducer

_No response_

### System Information

```Shell
primereact: 10.9.7
```

Fill in the whole input, move left with the arrow keys, try to overwrite it.

### Expected behavior

It should overwrite it, so 1244 when cursor is positioned at the 2, should become 1999, when pressing 9 three times.

navedqb:
Fixes #8306


melloware (approved):

melloware:
Can you run `npm run format` please?

PR #8319, opened 2025-10-06
#8318 Fix the showHeader default value of Editor

### LINKED ISSUE #8318: [Editor] showHeader default value is wrong
showHeader default value of Editor is wrong in the doc

yunics-highfield:
Fix #8318 

It's true actually.

https://github.com/primefaces/primereact/blob/ea77943a41b8a9a5bfca0267bf6043334e6180c3/components/lib/editor/EditorBase.js#L1084

### DIFF
--- a/components/doc/common/apidoc/index.json
+++ b/components/doc/common/apidoc/index.json
@@ -24414,7 +24414,7 @@
                             "optional": true,
                             "readonly": false,
                             "type": "boolean",
-                            "default": "true",
+                            "default": "false",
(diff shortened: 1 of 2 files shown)


melloware:
@yunics-highfield you need to change it here: https://github.com/primefaces/primereact/blob/05c76114472bb2d0ce80ba58e35d0d45237eb08a/components/lib/editor/editor.d.ts#L140

the docs are driven from the TypeScript and the TypeScript has it incorrect!  Update your PR and I will merge.

PR #8335, opened 2025-10-14
Fix #5975

### LINKED ISSUE #5975: TieredMenu: Submenu 'popup' behavior different than PrimeVue
### Describe the bug

I have two forked demos that shows the difference in behavior. In `10.2` TieredMenu submenus were activated by hover. In the latest version they can only be activated by clicking.

 - Latest version

https://stackblitz.com/edit/zjjwcj?file=src%2FApp.jsx 

 - v10.2

 https://stackblitz.com/edit/ompses?file=src%2FApp.jsx

### Reproducer

https://stackblitz.com/edit/ompses?file=src%2FApp.jsx

### PrimeReact version

latest

### React version

18.x

### Language

TypeScript

### Build / Runtime

Vite

### Browser(s)

Tested in chrome

Use the demo to see the different behavior.

### Expected behavior

I expect the submenus to activate by hover.

kyybo:
* Since version 10.3.0, a condition has been added to prevent submenus from activating on hover when in “popup” mode (commit 5615c89a0621867f8072c78738b091afe3436e2b). I don't really understand why this constraint exists. But removing it resolves issue #5975 and aligns with PrimeVue's behaviour.

* little format fix in Slider.js

Fix #5975

### DIFF
--- a/components/lib/slider/Slider.js
+++ b/components/lib/slider/Slider.js
@@ -20,7 +20,7 @@ export const Slider = React.memo(
         const barWidth = React.useRef(0);
         const barHeight = React.useRef(0);
         const touchId = React.useRef();
-        const value = props.range ? props.value ?? [props.min, props.max] : props.value ?? props.min ?? 0;
+        const value = props.range ? (props.value ?? [props.min, props.max]) : (props.value ?? props.min ?? 0);
         const horizontal = props.orientation === 'horizontal';
         const vertical = props.orientation === 'vertical';
 


melloware:
@kyybo did you look at Git Blame and see when and why that line was changed?

kyybo:
Yes, as I said in my commit message, this line appeared since the commit https://github.com/primefaces/primereact/commit/5615c89a0621867f8072c78738b091afe3436e2b
However, there is no explanation as to why the condition `&& !props.popup` is present.

When we look at PrimeVue code, there is only the condition on the `dirty` property but not the `popup` property :
https://github.com/primefaces/primevue/blob/master/packages/primevue/src/tieredmenu/TieredMenu.vue#L306

melloware:
OK run `npm run format`

PR #8323, opened 2025-10-07
fix(Calendar): prevent auto-refocus on single selection mode (#8302)

Coderxrohan:
Fixes [#8302](https://github.com/primefaces/primereact/issues/8302)

Cause: In single selection mode, when a date is selected, the overlay closes but the input field is immediately refocused with reFocusInputField(). This prevents reopening the calendar by clicking the input again.

Implementation: Updated onDateSelect to conditionally skip the refocus when selectionMode === 'single'.

```
- reFocusInputField();
+ if (props.selectionMode !== 'single') {
+     reFocusInputField();
+ }

```

Result:

✅ Single selection mode → panel closes without auto-refocus, allowing the user to reopen the calendar by clicking the input.

✅ Range / multiple selection modes → auto-refocus remains, preserving accessibility.

@Coderxrohan

Issue #8302, opened 2025-09-25
Calendar: reFocusInputField on single selection


sergiomendi:
### Describe the bug

Using Calendar as single date selector, when I select a date and the panel closes, there is an auto focus at the input field that breaks usability since I can not open the panel again without pressing the right button of the component.

### Reproducer

https://stackblitz.com/run?file=src%2FApp.jsx

### System Information

```Shell
Primereact: 10.9.7
```

### Steps to reproduce the behavior

1. Open calendar panel
2. Select a date
3. Open calendar panel again pressing the input field

### Expected behavior

I have noticed that the single Calendar has a function when there is a selection that re focus input field. This behavior should be supressed.

Coderxrohan:
Issue have been solved >> https://github.com/primefaces/primereact/pull/8323
Pull request has raised.

Result:
>>>>
1) Single selection mode → panel closes without auto-refocus, allowing the user to reopen the calendar by clicking the input.
2) Range / multiple selection modes → auto-refocus remains, preserving accessibility.

@Coderxrohan
@melloware 
@sergiomendi

PR #8359, opened 2025-10-29
fix: show date selection onClick

### LINKED ISSUE #8352: Calendar: show date-section on click
### Describe the bug

Situation: I select a date and then want to select another date again
Bug: the date section doesn't appear like the first time
source: https://primereact.org/calendar/
you can see it in the documentation page too  
note: if I click somewhere else and then click the input again, it appears.

### Reproducer

 https://primereact.org/calendar/

### System Information

```Shell
"primereact": "10.9.6"
```

1. Click a pick a date
2. Click again to pick another date
3. the date-section doesn't appear
I must click somewhere else and then in the input to let the date-section show

### Expected behavior

1. Click in the Input to show the date section
2. select a date
**3. click again directly in the input to show the date section**
**4. the date section appears to pick another date**

navedqb:
Fixes #8352

### DIFF
--- a/components/lib/calendar/Calendar.js
+++ b/components/lib/calendar/Calendar.js
@@ -99,6 +99,12 @@ export const Calendar = React.memo(
             return props.dateFormat || localeOption('dateFormat', props.locale);
         };
 
+        const onInputClick = () => {
+            if (!visible && props.showOnFocus) {
+                show();
+            }
+        };
+
         const onInputFocus = (event) => {
             if (ignoreFocusFunctionality.current) {
                 setFocusedState(true);
@@ -4206,6 +4212,7 @@ export const Calendar = React.memo(
                         autoComplete="off"
                         placeholder={props.placeholder}
                         tabIndex={props.tabIndex}
+                        onClick={onInputClick}
                         onInput={onUserInput}
                         onFocus={onInputFocus}
                         onBlur={onInputBlur}


melloware:
@Coderxrohan can you review this one and check if your other fixes already fixed this issue?

Coderxrohan:
@melloware 

I tested this behavior, https://stackblitz.com/edit/vitejs-vite-8xz26kct?file=src%2FApp.tsx

What I observed:
- Selecting a date works as expected.
- After the calendar closes, clicking the input again does not reopen the date picker immediately.
- The date can still be edited via keyboard.
- If focus is moved elsewhere and then the input is clicked again, the calendar opens.

From my testing, this behavior is not related to my recent changes and is not fixed by them.
It also feels more like focus/UX behavior than a clear functional bug, especially since keyboard input remains available (useful for large or far-range dates).

Sharing the StackBlitz repro here so we can confirm whether this is expected behavior or something that should be adjusted.

Coderxrohan:
Just to clarify the distinction between the previous fix and this issue:

The fix I worked on earlier (#8302 / #8323) addressed an auto-refocus problem in **single selection mode**.
After selecting a date, the input was being programmatically refocused via `reFocusInputField()`, which prevented reopening the calendar with a normal click.
That was fixed by skipping the refocus logic only for `selectionMode="single"`, while keeping it for range/multiple modes.

The behavior discussed here is different.
From testing, this is about the calendar **not reopening on input click** after the panel closes, unless focus is reset.
This is related to click/focus handling and overlay open logic, not forced refocus after selection.

So:
**- The earlier fix removed unwanted auto-focus after selection.
- This issue concerns whether clicking the input should explicitly reopen the overlay.**

**There is no functional overlap between the two, and my previous changes neither caused nor fixed this behavior.**
Sharing this clarification so the two concerns don’t get conflated.

melloware:
@Coderxrohan i agree.

Coderxrohan:
> @Coderxrohan i agree.

But check once the https://stackblitz.com/edit/vitejs-vite-8xz26kct?file=src%2FApp.tsx

Coderxrohan:
@​melloware  
I tested this using the PR branch and the issue appears to be resolved.  
The calendar behavior works as expected on click after selection.  
LGTM from my side.

See Current Calender Behaviour: https://stackblitz.com/edit/vitejs-vite-8xz26kct?file=package.json
With this PR: https://codesandbox.io/p/github/qburst/primereact-1/fix--8352

Coderxrohan:
https://github.com/user-attachments/assets/b0e3b17f-010e-47f4-ab8f-016e60535cc3

Testting Locally on Localhost

Coderxrohan:
@melloware 
Help Needed !

1) When testing on localhost, it works normally, as shown in the attached video above, 
Behaviour = Normal

2) When tested in https://stackblitz.com/edit/vitejs-vite-8xz26kct?file=package.json
Behavior: The Calendar doesn't open after selection

3) When tested in https://codesandbox.io/p/github/qburst/primereact-1/fix--8352 ( with this PR repo)
Behave normally.

4) When tested in https://primereact.org/calendar/ 
Behavior: The Calendar doesn't open after selection

melloware:
It might be a react strict mode thing with double hook calls?

Coderxrohan:
> It might be a react strict mode thing with double hook calls?

Yes, I think that’s very likely 👍

The inconsistent behavior only appears in environments running React Strict Mode (StackBlitz / primereact.org), while localhost and the CodeSandbox PR branch behave normally.

This points to double-invoked effects or focus/overlay timing differences under Strict Mode rather than an issue introduced by this PR.

From my side, nothing further to add — the fix looks good and the remaining behavior can be evaluated separately if we decide to handle Strict Mode explicitly.

PR #8392, opened 2025-11-13
fix(InputNumber): fix decimal rounding on spin

### LINKED ISSUE #8382: InputNumber step rounding to one decimal place
### Describe the bug

An InputNumber with the step set to a value with two or more decimal places will round steps to 1 decimal place.

In the example linked below, when stepping down on a value like 10.5 it will round the step to 0.2 and when stepping up it rounds to 0.3. If I manually set the value to 10.25 it will round down or up to the correct value. Values can be entered manually (without stepping) to more than 2 decimal places ok.

### Reproducer

https://primereact.org/inputnumber/#buttons

### System Information

```Shell
System:
    OS: Linux 6.4 openSUSE Leap 15.6
    CPU: (28) x64 Intel(R) Core(TM) i7-14700K
    Memory: 3.09 GB / 46.82 GB
    Container: Yes
    Shell: 4.4.23 - /bin/bash
  Binaries:
    Node: 22.13.1 - /usr/bin/node
    npm: 11.2.0 - /usr/local/bin/npm
  Browsers:
    Chrome: 139.0.7258.154
    Firefox: 128.7.0esr
    Firefox Developer Edition: 128.7.0esr
  npmPackages:
    primereact: ^10.9.7 => 10.9.7 
    react: ^19.2.0 => 19.2.0
```

Under the buttons section on the InputButton page, click on the + or minus button on the **Horizontal with Step** example, or use the arrow keys to change the value.

### Expected behavior

The value should change by 0.25, however it changes by 0.2 or 0.3, depending on the step direction.

selfplusplus2:
Fixes #8382

### DIFF
--- a/components/lib/inputnumber/InputNumber.js
+++ b/components/lib/inputnumber/InputNumber.js
@@ -213,8 +213,55 @@ export const InputNumber = React.memo(
             return null;
         };
 
-        const addWithPrecision = (base, increment, precision = 10) => {
-            return Math.round((base + increment) * precision) / precision;
+        function countDecimals(value) {
+            if (!isFinite(value)) return 0;
+
+            const s = String(value);
+
+            // handle exponential notation: e.g. "1e-7"
+            if (s.toLowerCase().includes('e')) {
+                const [mantissa, expStr] = s.split('e');
+                const exp = parseInt(expStr, 10);
+                const decimalPart = (mantissa.split('.')[1] || '').length;
+
+                // If exponent is negative, decimals increase
+                if (exp < 0) return decimalPart + Math.abs(exp);
+
+                // If exponent is positive, decimals shrink
+                return Math.max(0, decimalPart - exp);
+            }
+
+            // normal decimal
+            return (s.split('.')[1] || '').length;
+        }
+
+        const addWithPrecision = (base, increment) => {
+            base = Number(base);
+            increment = Number(increment);
+
+            // invalid inputs → return NaN cleanly
+            if (!isFinite(base) || !isFinite(increment)) return NaN;
+
+            const baseDec = countDecimals(base);
+            const incDec = countDecimals(increment);
+
+            // Choose required decimal precision but cap so factor remains safe
+            const decimals = Math.min(Math.max(baseDec, incDec), 15);
+            const factor = Math.pow(10, decimals);
+
+            const maxSafe = Number.MAX_SAFE_INTEGER;
+
+            // avoid unsafe multiplication
+            if (Math.abs(base) * factor > maxSafe || Math.abs(increment) * factor > maxSafe) {
+                const sum = base + increment;
+                // fallback to a safe rounding
+                const fallbackFactor = Math.pow(10, 15);
+
+                return Math.round(sum * fallbackFactor) / fallbackFactor;
+            }
+
+            // Correct integer math → avoids floating point errors
+            return Math.round(base * factor + increment * factor) / factor;
         };
 
         const repeat = (event, interval, dir) => {
--- a/components/lib/inputnumber/InputNumber.spec.js
+++ b/components/lib/inputnumber/InputNumber.spec.js
@@ -1,0 +1,101 @@
+import '@testing-library/jest-dom';
+import { render, fireEvent } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+import React from 'react';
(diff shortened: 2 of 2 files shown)


melloware:
GitHub Copilot Chat Assistant

Quick summary
- The new implementation is simpler than the old (it auto-computes precision from the operands), but it has a few pitfalls:
  - It fails for numbers printed in exponential notation (e.g. 1e-7) because toString().split('.') yields no decimal part.
  - It doesn't guard against string inputs, NaN, Infinity.
  - Math.pow(10, decimals) can exceed safe integers for large decimal counts (and large scale factors can overflow Number.MAX_SAFE_INTEGER).
  - Rounding via Math.round((base + increment) * precision) can still pick up floating errors before scaling; it's safer to scale each operand or the sum before rounding.
  - For financial/critical precision you should use a decimal library (decimal.js / Big.js / BigNumber).

Problems I found in current diff
- Exponential notation: (0.0000001).toString() === "1e-7" -> baseDec becomes 0, but you actually need 7 decimals.
- Input types: if base or increment are strings (common in UI code) you need to coerce to Number and handle invalid values.
- Very large decimals: Math.pow(10, n) with n > ~15 can cause factor > 2^53 and integer math becomes unsafe.
- Edge cases (NaN, Infinity) are not handled.

Suggested robust implementation
- Count decimals correctly including exponential notation.
- Cap decimals to a safe limit (e.g. 15) to avoid exceeding Number.MAX_SAFE_INTEGER.
- Use scaled integer addition and rounding: Math.round(base * factor + increment * factor) / factor
- Return NaN when inputs are not finite numbers, or fallback gracefully.

Example implementation (replace your function with this)
function countDecimals(value) {
  if (!isFinite(value)) return 0;
  const s = String(value);
  // handle exponential notation
  if (s.indexOf('e') >= 0) {
    const [mantissa, expStr] = s.split('e');
    const exp = parseInt(expStr, 10);
    const dec = (mantissa.split('.')[1] || '').length;
    return Math.max(0, dec - exp);
  }
  return (s.split('.')[1] || '').length;
}

const addWithPrecision = (base, increment) => {
  base = Number(base);
  increment = Number(increment);
  if (!isFinite(base) || !isFinite(increment)) return NaN;

  const baseDec = countDecimals(base);
  const incDec = countDecimals(increment);

  // cap decimals to avoid factor > Number.MAX_SAFE_INTEGER
  const decimals = Math.min(Math.max(baseDec, incDec), 15);
  const factor = Math.pow(10, decimals);

  // If scaling would overflow, fall back to simple addition with an approximate rounding
  const maxSafe = Number.MAX_SAFE_INTEGER;
  if (Math.abs(base) * factor > maxSafe || Math.abs(increment) * factor > maxSafe) {
    const sum = base + increment;
    // round to a reasonable number of digits (15)
    const fallbackFactor = Math.pow(10, 15);
    return Math.round(sum * fallbackFactor) / fallbackFactor;
  }

  return Math.round(base * factor + increment * factor) / factor;
};

Behavior examples you can use to test
- addWithPrecision(0.1, 0.2) -> 0.3
- addWithPrecision(1e-7, 1e-7) -> 2e-7
- addWithPrecision('1.234', 0.001) -> 1.235
- addWithPrecision(NaN, 1) -> NaN
- addWithPrecision(Infinity, 1) -> NaN

Recommendations
- Add unit tests covering:
  - small decimals (0.1 + 0.2)
  - exponential inputs (1e-7)
  - string inputs
  - very large/safe boundary cases
- If this function is used for money or other critical decimal arithmetic, switch to a decimal/BigNumber library (decimal.js, big.js, bignumber.js) to avoid all floating-point pitfalls.
- Keep the decimal cap (15) documented so future maintainers understand the safety tradeoff.

If you want, I can:
- open a PR with this improved implementation and tests, or
- adapt the implementation to use a specific decimal library if you prefer. Which would you like?

selfplusplus2:
I have amended to factor in exponential numbers and added some tests

melloware:
please run `npm run format`

selfplusplus2:
Formatting updated

melloware:
ok `npm run lint` next please

selfplusplus2:
Done

melloware:
weird now its saying the format is wrong again?

selfplusplus2:
It doesn't like my editor white space settings 🤣🤦. I've re-run format and lint.

PR #8409, opened 2025-11-27
fix(Tree): only toggle 'filter' expanded keys when the filter is used

### LINKED ISSUE #8408: Tree: Bad initialization of expanded keys when filter option is present
### Describe the bug

The initial value given to `expandedKeys` gets overridden if a filter field is present, even if the filter value is empty.

In the reproducer, both trees use the same initial `expandedKeys` but the tree with the filter field is not initialized properly (all nodes are collapsed).

Image

### Reproducer

https://stackblitz.com/edit/edubbxcg

### System Information

```Shell
primereact: 10.9.7
```

_No response_

### Expected behavior

Both trees should be the same upon first render.

RolT:
Fixes #8408

### DIFF
--- a/components/lib/tree/Tree.js
+++ b/components/lib/tree/Tree.js
@@ -343,13 +343,13 @@ export const Tree = React.memo(
                         filteredNodes.current.push(copyNode);
                     }
                 }
+                
+                onToggle({
+                    originalEvent: null,
+                    value: currentFilterExpandedKeys.current,
+                    navigateFocusToChild: false
+                });
             }
-
-            onToggle({
-                originalEvent: null,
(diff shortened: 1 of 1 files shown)


Coderxrohan (approved):
#Issue

expandedKeys were being reset when a filter input was present, even if the filter was empty, causing nodes to collapse unexpectedly on init and model updates.

#Fix

Filter-related expanded keys are now toggled only when the filter is actually used, preserving the original tree state otherwise.

#Impact

Initial render and runtime updates behave correctly; expanded state is no longer lost when the filter is empty.

⭐ Rating: 5/5

PR #8452, opened 2025-12-31
Fix Menu popup hover flicker

### LINKED ISSUE #7506: Menu: Wrong hovering effect
### Describe the bug

When the Menu is used as a popup, and you hover over between elements, then the visual effect have some delay causing some visual glitches.

gif3

### Reproducer

https://stackblitz.com/edit/qhjakw4h?file=src%2FApp.jsx

### System Information

```Shell
System:
    OS: Windows 10 10.0.19045
    Memory: 17.13 GB / 31.93 GB
  Binaries:
  Browsers:
  npmPackages:
    primereact: ^10.8.5 => 10.8.5
    react: ^18.2.0 => 18.2.0
```

1. Go to https://stackblitz.com/edit/qhjakw4h?file=src%2FApp.jsx
2. Hover over between the subsequent items in the Menu

### Expected behavior

No visual glitch. After switching the hover from one menu item to another, there should be the final styling applied already without any delay

Coderxrohan:
#### Description
Popup Menu items had a visible hover delay when moving rapidly between items.
This was caused by inherited CSS transitions on `.p-menuitem-link`, which led
to overlapping hover-in / hover-out animations and resulted in visual glitches.

#### Solution
Disable transitions specifically for popup Menu items by adding an explicit
rule for `.p-menu .p-menuitem-link`.

This ensures hover styles are applied immediately, eliminating flicker and
improving UX consistency.

#### Scope of Change
- CSS-only change
- Scoped strictly to popup Menu (`.p-menu`)
- No impact on Menubar, MegaMenu, or other navigation components
- No API or behavior changes

#### Verification
- Reproduced the issue locally using the Menu → Popup demo
- Verified hover behavior by moving quickly between items
- Confirmed via DevTools that `transition: none` is applied to
  `.p-menuitem-link`

Fixes #7506 (Menu: Visual glitches in hovering effect)

### DIFF
--- a/public/themes/arya-green/theme.css
+++ b/public/themes/arya-green/theme.css
@@ -1537,6 +1537,9 @@
     border-radius: 3px;
     box-shadow: 0 2px 4px -1px rgba(0, 0, 0, 0.2), 0 4px 5px 0 rgba(0, 0, 0, 0.14), 0 1px 10px 0 rgba(0, 0, 0, 0.12);
   }
+  .p-menu .p-menuitem-link {
+    transition: none;
+  }
   .p-mention-panel .p-mention-items {
     padding: 0.5rem 0;
   }


Coderxrohan:
This change adds an explicit CSS rule for `.p-menu .p-menuitem-link`
 to disable hover transitions for popup Menu items. Previously, inherited transitions caused overlapping hover-in and hover-out animations when moving quickly between items, resulting in a visible flicker. Removing the transition ensures hover styles are applied immediately, fixing the visual glitch without affecting Menubar, MegaMenu, or any other components.

PR #8453, opened 2025-12-31
Docs: Clarify locale precedence when using PrimeReactProvider

### LINKED ISSUE #8349: InputNumber: Respect Global Locale?
Hello,
I'm having trouble to understand how to use locale correctly. According to the [documentation](https://primereact.org/locale/#setlocale) I should simply call the `locale` method, but I've prepared 3 examples of using the `locale` and the behavior is different in each one of them.
The examples are based on an example I found in https://github.com/primefaces/primelocale/issues/182#issuecomment-2217449456

### Expected behavior
* The application always works as in scenario 1.

### Good enough behavior
If for some reason the expected behavior cannot be implemented:
* The proper way of locale configuration is described in the documentation.
* All components behaves consistently in each scenario.

### Scenario 1 - calling locale globally
```
addLocale('pl', plLocale['pl']);
locale('pl');
root.render(
  <React.StrictMode>
    <App />
  </React.StrictMode>
);
```
https://stackblitz.com/edit/react-zyv3qr-eut6jwny?file=src%2FApp.js,src%2Findex.js

In this scenario everything works fine:
* ✔️ Numbers are formatted using Polish locale, e.g. `1 234,56`
* ✔️ Dates are formatted using Polish locale, e.g. `22.10.2025`
* ✔️ Calendar uses Polish names
 
### Scenario 2 - calling locale globally and using PrimeReactProvider
When the application is wrapped inside PrimeReactProvider component, the locale simply stop working
```
addLocale('pl', plLocale['pl']);
locale('pl');
root.render(
  <React.StrictMode>
    <PrimeReactProvider>
      <App />
    </PrimeReactProvider>
  </React.StrictMode>
);
```
https://stackblitz.com/edit/react-zyv3qr-tauquecy?file=src%2FApp.js,src%2Findex.js

Calling the `locale` seems to have no effect - the primereact components use the default 'en' locale.
* ❌ Numbers are formatted using English locale, e.g. `1,234.56`
* ❌ Dates are formatted using English locale, e.g. `10/22/2025`
* ❌ Calendar uses English names
 
### Scenario 3 - calling locale inside a component and using PrimeReactProvider
That's actually how I currently use locale (the locale depends on current user's context, so it cannot be set globally) and that's the most messy scenario.
https://stackblitz.com/edit/react-zyv3qr-947xus6s?file=src%2FApp.js,src%2Findex.js
The only difference compared to scenario 2 is that now the `addLocale` and `locale` methods are called inside the `App`component.

In this scenario:
* ❌ Numbers are formatted using English locale, e.g. `1,234.56`
* ✔️ Dates are formatted using Polish locale, e.g. `22.10.2025`
* ✔️ Calendar uses Polish names

I guess I should be using `PrimeReactContext.setLocale` in the 2nd and 3rd scenario, but that `setLocale` method is not documented. (And if I do that, should I call the `locale` method as well, or only the `setLocale`?)

Also (and most importantly) I think that all components should respect the same locale in the 3rd scenario - there is no reason for numbers and dates to be ever formatted differently (unless someone explicitly specifies different locale in different components).

Coderxrohan:
### Description

This PR clarifies locale resolution behavior when using `PrimeReactProvider`.

It documents the locale precedence order and explains why calling `locale()` alone
does not override the provider locale once a context is present.

### Motivation

This resolves confusion discussed in #8349, where components (notably InputNumber)
appeared to ignore the global locale when wrapped with `PrimeReactProvider`.

### Changes
- Documented locale precedence order
- Clarified PrimeReactProvider as the primary locale source
- Added an explicit provider-based locale example

Fixes #8349 (InputNumber: Respect Global Locale)

### DIFF
--- a/components/doc/configuration/locale/setuplocaledoc.js
+++ b/components/doc/configuration/locale/setuplocaledoc.js
@@ -31,25 +31,27 @@ export default function MyApp({ Component }) {
                 </p>
 
                 <p>
-                    When <b>PrimeReactProvider</b> is used, it becomes the primary source of locale
-                    configuration for all PrimeReact components.
+                    When <b>PrimeReactProvider</b> is used, it becomes the primary source of locale configuration for all PrimeReact components.
                 </p>
 
                 <h5>Locale Precedence</h5>
-                <p>
-                    Locale resolution follows the order below:
-                </p>
+                <p>Locale resolution follows the order below:</p>
                 <ul>
-                    <li><code>locale</code> prop defined on a component</li>
-                    <li><code>PrimeReactProvider</code> (<code>context.locale</code>)</li>
-                    <li>Global <code>PrimeReact.locale</code> set via <code>locale()</code></li>
+                    <li>
+                        <code>locale</code> prop defined on a component
+                    </li>
+                    <li>
+                        <code>PrimeReactProvider</code> (<code>context.locale</code>)
+                    </li>
+                    <li>
+                        Global <code>PrimeReact.locale</code> set via <code>locale()</code>
+                    </li>
                     <li>Browser locale</li>
                 </ul>
 
                 <p>
                     This means that once <b>PrimeReactProvider</b> is present, calling
-                    <code> locale()</code> alone is not sufficient. The locale must be provided
-                    explicitly through the provider.
+                    <code> locale()</code> alone is not sufficient. The locale must be provided explicitly through the provider.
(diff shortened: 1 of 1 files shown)


Coderxrohan:
Added documentation clarifying locale precedence with **PrimeReactProvider** and why **locale()** does not override provider locale. No behavior changes, docs only.

Issue #7428, opened 2024-11-18
Calendar: incorrect value is displayed


avasuro:
### Describe the bug

If the date in the onChange handler of the Calendar was not updated in external state, the calendar shows incorrect date in date input.

### Reproducer

https://codesandbox.io/p/sandbox/primereact-calendar-displays-incorrect-value-if-new-value-was-not-applied-in-onchange-sw2547

### System Information

```Shell
System:
    OS: Windows 11 10.0.22621
    CPU: (12) x64 12th Gen Intel(R) Core(TM) i5-1235U
    Memory: 3.68 GB / 15.71 GB
  Binaries:
    Node: 18.20.3 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.22.19 - ~\AppData\Roaming\npm\yarn.CMD
    npm: 10.7.0 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Edge: Chromium (127.0.2651.74)
    Internet Explorer: 11.0.22621.3527
  npmPackages:
    primereact: 10.8.4 => 10.8.4
    react: 18.3.1 => 18.3.1
```

### Steps to reproduce the behavior

1. Open demo
2. Click on calendar input and select some date with a mouse
3. Confirmation dialog will appear - click "No" there

Actual behavior: some date is displayed inside the input
Expected behavior: no date should be displayed inside the input

If you click on date input again - date will disappear, as it should be.

### Expected behavior

Calendar component should take into account the date provided in props all the time, even if this date wasn't changed

Coderxrohan:
I’ve opened a PR that fixes this by ensuring the Calendar input value is always derived from `props.value` and not updated optimistically on date selection.  
This restores correct controlled component behavior when the parent intentionally does not update the value.

Coderxrohan:
@melloware 
This is a long-standing controlled-component edge case. The Calendar input could briefly display a value even when the parent does not update **value**. While subtle, it can confuse users and breaks controlled input expectations. The fix is minimal and restores proper synchronization from **props.value**.

Coderxrohan:
The final change stabilizes the Calendar input sync by keeping the latest updateInputfield via a ref, avoiding unstable hook dependencies while preserving existing behavior.

PR #8457, opened 2026-01-01
Fix Calendar displaying value when controlled state is not updated

Coderxrohan:
#### Problem
When the Calendar component is used as a controlled component, it updates its input field optimistically on date selection.  
If the parent component intentionally does not update the `value` in `onChange` (e.g. user cancels a confirmation dialog), the input temporarily displays an incorrect date.

#### Solution
- Removed optimistic input updates on date selection
- Ensured the input value is always derived from `props.value`
- Input now updates only when the external value actually changes

This restores proper controlled component behavior and fixes issue #7428.

#### Reproducer
https://codesandbox.io/p/sandbox/primereact-calendar-displays-incorrect-value-if-new-value-was-not-applied-in-onchange-sw2547

Calendar: incorrect value is displayed #7428

### DIFF
--- a/components/lib/calendar/Calendar.js
+++ b/components/lib/calendar/Calendar.js
@@ -70,13 +70,6 @@ export const Calendar = React.memo(
         const [yearOptions, setYearOptions] = React.useState([]);
 
         const previousValue = usePrevious(props.value);
-
-        React.useEffect(() => {
-            if (props.value !== previousValue) {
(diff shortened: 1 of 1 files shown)


Coderxrohan:
The final change stabilizes the Calendar input sync by keeping the latest updateInputfield via a ref, avoiding unstable hook dependencies while preserving existing behavior.

PR #8458, opened 2026-01-01
Fix Calendar mask not firing onInput event

### LINKED ISSUE #7640: Calendar: With "mask", not work callback "onInput"
### Describe the bug

With propertie "mask", callback "onIput" not work, when remove "mask" "onInput" working

https://github.com/user-attachments/assets/4edceb50-0c1e-47fd-a836-4ca5d119a386

### Reproducer

https://stackblitz.com/edit/2jbv9vw5?file=src%2FApp.jsx

### System Information

```Shell
"primereact": "10.9.1",
"react": "18.3.1",
```

1.Use Calendar Mask
2.Callback OnInput not show event

### Expected behavior

1.Use Calendar Mask
2.Callback OnInput show event

Coderxrohan:
#### Problem
When the `mask` prop is used, Calendar routes user typing through
`updateValueOnInput()` but does not forward the native `onInput` event.
As a result, `props.onInput` never fires when mask is enabled.

#### Solution
Forward `props.onInput(event)` inside `updateValueOnInput`, ensuring
consistent behavior between masked and unmasked Calendar inputs.

#### Result
- `onInput` fires correctly with `mask`
- No behavior change without `mask`
- No breaking API changes

Fixes #7640 (Calendar: With "mask", not work callback "onInput")


Coderxrohan:
@melloware
I was able to resolve this third bug quickly (within a day) **since it was located in the same Calendar file** as another high-priority issue I was already working on. I was also personally affected by this behavior, which made it easier to identify the root cause and apply the fix efficiently.

Please let me know if you’d like any refinements or adjustments.

Coderxrohan:
Working on (and one that also affected me), I was able to quickly understand & fix it.
https://github.com/primefaces/primereact/pull/8457 Fix Calendar displaying value when controlled state is not updated

x10vaishnavi:
@Coderxrohan 
Yes, The issue is now resolved and working as expected.
The fix has been implemented and verified locally. I’ve tested the affected Calendar scenarios, and the `onInput` behavior is now consistent with controlled state updates.

Thanks for this fix, as this issue caused me to panic for some time.

melloware (changes requested):

Coderxrohan:
Thanks @melloware 
For taking care of the refactor and final adjustments.
I appreciate you stepping in to align the useEffect structure and helping get this merged so quickly. 🙏

PR #8462, opened 2026-01-05
fix(docs): hide API tab when apiDocs are not available

### LINKED ISSUE #8388: [v11] docs: API page for Tabs component in the docs returns a 404
### Describe the bug

https://v11.primereact.org/docs/components/tabs/api

### Reproducer

_No response_

### System Information

```Shell
N/A
```

Issue occurs both when you open the page by the url directly as well as clicking on the "API" tab on https://v11.primereact.org/docs/components/tabs page

### Expected behavior

To show docs info

Coderxrohan:
### Problem
The Tabs documentation shows an API tab even though Tabs does not provide API documentation.
Clicking this tab navigates to a non-existent API route, resulting in a 404 on the docs site.

### Impact
Users end up on a broken page, which is confusing and gives the impression that the documentation is incomplete or incorrect.

### Fixes
The API tab is now rendered only when API documentation is available (apiDocs is provided).
This prevents navigation to dead API routes while keeping existing behavior unchanged for components that do expose API docs.

### Approach Used
I reviewed how the docs UI decides when to render the API tab and added a small conditional check so it only appears when API data exists. This keeps the fix minimal, safe, and limited to documentation logic.

Fixes #8388 ([v11] docs: API page for Tabs component in the docs returns a 404)

### DIFF
--- a/components/doc/common/doccomponent.js
+++ b/components/doc/common/doccomponent.js
@@ -48,11 +48,11 @@ export function DocComponent(props) {
                         </button>
                     </li>
                     {props.apiDocs ? (
-                    <li className={classNames({ 'doc-tabmenu-active': tab === 1 })}>
-                        <button type="button" onClick={() => activateTab(1)}>
-                            API
-                        </button>
-                    </li>
+                        <li className={classNames({ 'doc-tabmenu-active': tab === 1 })}>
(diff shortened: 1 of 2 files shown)


melloware:
looks liek you need `npm run format`

PR #8470, opened 2026-01-09
fix(FocusTrap): resolve aria_hidden_nontabbable accessibility violation

### LINKED ISSUE #8365: Dialog : Accessibility - Element "span" should not be focusable within the subtree of an element with an 'aria-hidden' attribute with value 'true'
### Describe the bug

**Accessibility  Violation**

`Element "span" should not be focusable within the subtree of an element with an 'aria-hidden' attribute with value 'true'
A hidden element should not contain any tabbable elements`

**Why is this important?**

1. Using either role="presentation" or aria-hidden="true" on a focusable element will result in some users focusing on "nothing". A focusable element with aria-hidden="true" is ignored as part of the reading order, but still part of the focus order, making its state of visible or hidden unclear. This rule checks that elements with an aria-hidden attribute do not contain focusable elements.

**Element location**
```
<span
    data-pc-section="firstfocusableelement"
    data-p-hidden-focusable="true"
    data-p-hidden-accessible="true"
    aria-hidden="true"
    role="presentation"
    tabindex="0"
    class="p-hidden-accessible p-hidden-focusable">

```

**What to do**

When designers and developers want to hide parts of a web page that are pure decoration, such as icon fonts - that are not meant to be read by assistive technologies:

- ensure that assistive technologies will ignore the element by adding aria-hidden="true" to the element *ensure none of the target elements are part of sequential focus navigation

- ensure no descendants are part of sequential focus navigation

**Example**

- In the following example, the button element is focusable by default. Therefore, using aria-hidden="true" violates the rule:

        <div aria-hidden="true">
            <div aria-hidden="false">
                <button>Some button</button>
            </div>
        </div>

**Note:** Using aria-hidden="false" on a descendant of an element with aria-hidden="true" does not expose that element. Using aria-hidden="true" hides itself and all its content from assistive technologies.

Rule ID: aria_hidden_nontabbable
Reason ID: Fail_1

### Reproducer

_No response_

### System Information

```Shell
Can be reproduced in every environment using IMB Accessibility checker
```

1. Create a Dialog.
2. Run IMB Accessibility checker.
3. Check Violations.

Image

Image

### Expected behavior

1. ensure that assistive technologies will ignore the element by adding aria-hidden="true" to the element *ensure none of the target elements are part of sequential focus navigation
2. ensure no descendants are part of sequential focus navigation

Aryan3902:
Fix  #8365 

## The Issue
The focus guard elements (the hidden spans used to trap focus) were rendered with conflicting attributes: tabIndex="0" combined with aria-hidden="true" and role="presentation".

### Reason
A focusable element (tabIndex="0") cannot be marked as hidden (aria-hidden="true") because it creates a state where focus lands on an element that technically "doesn't exist" in the accessibility tree.

## The Fix
I have removed aria-hidden={true} and role={'presentation'} from the createFocusableElement helper.

These elements are technically focusable, so they must be exposed to the accessibility tree.

They remain visually hidden via the existing p-hidden-accessible CSS class.

The onFocus event handlers continue to function correctly, redirecting focus instantly without negatively impacting screen reader announcements.

image

## How Has This Been Tested?
Automated Scan: Verified that the aria_hidden_nontabbable (Fail_1) error no longer appears in accessibility scans.

Functional Test: Verified that the focus trap still functions correctly (Focus loops back to the start when tabbing past the last element, and vice versa).

Visual Test: Confirmed no visual regression (the spans remain invisible).


melloware (approved):

melloware:
it lokos like Primevue is doing the same thing: https://github.com/primefaces/primevue/blob/b83794d8f989321bd70196d9dd6b0e8cdeef547b/packages/primevue/src/focustrap/FocusTrap.js#L99-L108

melloware:
Assigned to @tugcekucukoglu to review since PrimeVue is doing the same thing

Aryan3902:
Hi @melloware @tugcekucukoglu , Any update?

Issue #8269, opened 2025-09-05
Dropdown: clear icon and value not shown when options are empty


riccardinhocala:
### Describe the bug

When using the Dropdown component with showClear={true}, the clear icon does not appear if options is empty, even though the dropdown has a value.

Additionally, if options is empty, the value itself is not displayed — even when a custom valueTemplate is provided to explicitly render it.

This creates issues in scenarios where options are loaded asynchronously or only available at certain times (e.g., after a button click or API call). 
In these cases, I would still like to show the currently selected value and show the clear icon if value is set and showClear={true}

### Reproducer

https://stackblitz.com/edit/zavs11yi

### System Information

```Shell
primereact: latest
react: 18.3.1
```

### Steps to reproduce the behavior

1. Render a Dropdown with a non-empty value and showClear={true}.
2. Pass an empty array to options.
3. Observe that both the value and the clear icon are missing.

tanerengiiin:
Thanks for the report. In v11 (Dropdown is now Select), the clear button is shown whenever there is a value, and an object value renders its own label even with empty options. With `optionValue`, only the key is stored, so the label cannot be resolved until the options arrive. Render it yourself in the meantime, e.g. `<Select.Value>{savedCountry?.name}</Select.Value>`.

PR #8475, opened 2026-01-13
Fix: Display value and clear icon in Dropdown when options are empty

Aryan3902:
For #8269 

# Changes
1. Added a new fallback value in rendering of content which takes the value from props.value incase the selectedOption is not present, as highlighted in issue.
2. Removed the empty options value check. 

image


melloware (approved):

Aryan3902:
@melloware any updates on this one?

PR #8494, opened 2026-02-04
fix(datatable): restore state after locale initialization

### LINKED ISSUE #8478: DataTable: locale string not applyed
### Describe the bug

When i use `stateStorage` attribute, locale string not applyed.

Image

### Reproducer

https://stackblitz.com/edit/nextjs-isb8tmj8?file=app%2Fpage.tsx

### System Information

```Shell
System:
  OS: Windows 11 10.0.22631
  CPU: (16) x64 AMD Ryzen 7 9800X3D 8-Core Processor
  Memory: 34.35 GB / 61.59 GB
Binaries:
  Node: 25.2.1 - C:\Program Files\nodejs\node.EXE
  npm: 11.6.2 - C:\Program Files\nodejs\npm.CMD
Browsers:
  Chrome: 144.0.7559.60
  Edge: Chromium (140.0.3485.54)
  Firefox: 143.0.1 - C:\Program Files\Mozilla Firefox\firefox.exe
npmPackages:
  primereact: ^10.9.7 => 10.9.7
  react: 19.2.3 => 19.2.3
  tailwindcss: ^4 => 4.1.18
```

1. Open [Reproducer](https://stackblitz.com/edit/nextjs-isb8tmj8?file=app%2Fpage.tsx)
2. Enter random text and click "apply"
Image

3. locale string not applyed

### Expected behavior

_No response_

Coderxrohan:
**Description**

- This PR fixes an issue where localized filter match mode labels were not applied when stateStorage is enabled in DataTable.
- The root cause was that table state restoration was happening before the locale was available, causing filters to be initialized with default (English) strings.

**What’s changed**

1. Delayed restoreState() until PrimeReactContext.locale is ready.
2. Ensured state restoration runs only once using a ref guard.
3. Kept mount logic clean (no timeouts or double filter mutations).
4. Column widths continue to restore as before.

**Result**

- [x] Localized filter labels now work correctly with stateStorage.
- [x] No behavioral regressions for existing stateful tables.
- [x] Fix follows existing PrimeReact lifecycle patterns.

**Related Issue**

- Fixes #8478
- Fixed #8478

### DIFF
--- a/components/lib/datatable/DataTable.js
+++ b/components/lib/datatable/DataTable.js
@@ -83,6 +83,7 @@ export const DataTable = React.forwardRef((inProps, ref) => {
     const columnSortFunction = React.useRef(null);
     const columnField = React.useRef(null);
     const filterTimeout = React.useRef(null);
+    const restoredRef = React.useRef(false);
 
     if (props.rows !== d_rowsState && !props.onPage) {
         setRowsState(props.rows);
@@ -341,16 +342,18 @@ export const DataTable = React.forwardRef((inProps, ref) => {
             }
 
             if (restoredState.filters) {
-                setD_filtersState(cloneFilters(restoredState.filters));
+                const clonedFilters = cloneFilters(restoredState.filters);
+
+                setD_filtersState(clonedFilters);
 
                 if (props.onFilter) {
                     props.onFilter(
                         createEvent({
-                            filters: restoredState.filters
+                            filters: clonedFilters
                         })
                     );
                 } else {
-                    setFiltersState(cloneFilters(restoredState.filters));
+                    setFiltersState(clonedFilters);
                 }
             }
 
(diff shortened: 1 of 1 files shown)


Coderxrohan:
Hi @melloware,
The fix delays DataTable state restoration until the locale is available, resolving the localized filter labels problem with stateStorage. Would appreciate your review when you have time. Thanks!

components/lib/datatable/DataTable.js
```diff
@@ -1529,6 +1523,18 @@ export const DataTable = React.forwardRef((inProps, ref) => {
...
+        if (!restoredRef.current && isStateful()) {
+            restoredRef.current = true;
+
+            restoreState();
+
+            if (props.resizableColumns) {
+                restoreColumnWidths();
+            }
+        }
+    }, [context?.locale]);
```
melloware: Hmmm does this really work as a useEffect trigger?

melloware (changes requested):

components/lib/datatable/DataTable.js
Coderxrohan: @melloware, Good question! 
Yes, `PrimeReactContext.locale` updates when locale() is called via the API, which triggers a context change and re-render. The intent here is to defer state restoration until after locale initialization so filter match labels are resolved correctly. I’ve also guarded it with restoredRef to ensure it only runs once.

If you’d prefer something more defensive, I’m also fine switching the dependency to [`context`] instead of [`context?.locale`].

What's your opinion on this?

Coderxrohan:
Updated to use `[context]` as the dependency to make the trigger more defensive, 
while still restoring only once via restoredRef. 
Thanks for pointing that out!

melloware (approved):

PR #8497, opened 2026-02-04
fix(orderlist): preserve original order when moving multi-selected items

### LINKED ISSUE #8495: Cannot reorder multi-selected items under specific condition
### Describe the bug

Reordering multiple selected items does not work if step n is selected first and then step n‑1. So if one selects the item n, then presses ctrl and selects the second item n-1 and then tries to reorder the items by clicking the up/down buttons, it doesn't work.
However, if step n‑1 is selected before step n, reordering works as expected.

### Reproducer

_No response_

### System Information

```Shell
primereact: 10.9.5
```

https://github.com/user-attachments/assets/6597d678-fe5e-4c52-8814-2ad00565caa9

Have a list of 4 items and multi-select item 3 and then 2 (as described above), press the up or down button => trying to move them up or down isn't working.
If you select first item 2 and then 3, moving up and down is working.
Can be reproduced here: https://primereact.org/orderlist/

### Expected behavior

_No response_

### LINKED ISSUE #8496: Wrong order after moving multiple-selected items to top or bottom under specific conditions
### Describe the bug

If one selects the item n, then presses ctrl and selects the second item n-1 and then tries to reorder the items by clicking the top/bottom buttons, the items order gets changed.

### Reproducer

_No response_

### System Information

```Shell
primereact: 10.9.5
```

https://github.com/user-attachments/assets/a0ec104c-cb8a-4c87-adff-7aad46989b79

Have a list of 4 items and multi-select item 3 and then 2 (as described above), press the top or bottom button => trying to move them to the top or to the bottom is working, but the 2 elements get mixed up. Item 3 will appear befoer item 2 in the order.
If you select first item 2 and then 3, there is no such order issue.
Can be reproduced here: https://primereact.org/orderlist/

### Expected behavior

_No response_

Coderxrohan:
**Defect Fix**

This PR addresses a multi-selection ordering issue in OrderList where items selected out of sequence (e.g., Ctrl-click) were reordered incorrectly when moved to the top or bottom.

**Summary of changes:**

Normalize multi-selection by sorting selected items based on their original list index before reorder operations.
Ensures selected items always retain their visual order regardless of selection sequence.

**Notes**

This is a focused UX bug fix with no API changes.
Only components/orderlist/OrderList.js is affected.
Behavior now matches user expectations and native list reordering patterns.

Fixes: #8496
Fixed:#8496
Fixes: #8495
Fixed:#8495

(Duplicate due to ci error)

### DIFF
--- a/components/lib/orderlist/OrderList.js
+++ b/components/lib/orderlist/OrderList.js
@@ -69,9 +69,15 @@ export const OrderList = React.memo(
             if (selected) {
                 newSelection = metaKey ? selectionState.filter((_, i) => i !== selectedIndex) : [value];
             } else {
-                newSelection = metaKey ? [...selectionState, value] : [value];
-            }
+                if (metaKey) {
+                    const merged = [...selectionState, value];
 
+                    // sort selection by original list order
+                    newSelection = merged.sort((a, b) => props.value.indexOf(a) - props.value.indexOf(b));
+                } else {
+                    newSelection = [value];
+                }
+            }
             setSelectionState(newSelection);
         };
 
--- a/components/lib/orderlist/OrderList.js
+++ b/components/lib/orderlist/OrderList.js
@@ -71,7 +71,7 @@ export const OrderList = React.memo(
             } else {
                 if (metaKey) {
                     const merged = [...selectionState, value];
-                    
+
                     newSelection = merged.sort((a, b) => props.value.indexOf(a) - props.value.indexOf(b));
                 } else {
                     newSelection = [value];
--- a/components/lib/orderlist/OrderList.js
+++ b/components/lib/orderlist/OrderList.js
@@ -71,6 +71,7 @@ export const OrderList = React.memo(
             } else {
                 if (metaKey) {
                     const merged = [...selectionState, value];
+                    
                     newSelection = merged.sort((a, b) => props.value.indexOf(a) - props.value.indexOf(b));
                 } else {
                     newSelection = [value];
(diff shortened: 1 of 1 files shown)


Coderxrohan:
Fixed:

https://github.com/user-attachments/assets/875b6063-696f-47b8-b6f1-ca41c43d5161

Coderxrohan:
Hi @melloware,
I’ve opened a PR that normalizes multi-selection order in OrderList.
This resolves both issues where selecting items out of sequence breaks 

1. Up/Down (#8495) and 
2. reverses Top/Bottom (#8496).

The fix simply sorts selected items by their original list index before reorder actions — no API changes.
Would appreciate your review when you have time. Thanks!

Coderxrohan:
@melloware, ready to merge!
Both changes were verified locally and are now fixed.

components/lib/orderlist/OrderList.js
```diff
@@ -69,7 +69,13 @@ export const OrderList = React.memo(
             if (selected) {
                 newSelection = metaKey ? selectionState.filter((_, i) => i !== selectedIndex) : [value];
             } else {
-                newSelection = metaKey ? [...selectionState, value] : [value];
+                if (metaKey) {
+                    const merged = [...selectionState, value];
+
+                    newSelection = merged.sort((a, b) => props.value.indexOf(a) - props.value.indexOf(b));
```
melloware: SHould we check if props.value is not undefined or null or else this will error?  I didn't look at the rest of the code so not sure its even possible

melloware (changes requested):

Coderxrohan:
Added a small defensive fallback for `props.value` to avoid potential null/undefined during sort. 
Pushed the update,
Thanks!

components/lib/orderlist/OrderList.js
Coderxrohan: Suggested changes applied. Please take a look. Thanks!

melloware (approved):

PR #8520, opened 2026-04-01
fix(InputNumber): decide prefix suffix using formatToParts

### LINKED ISSUE #8521: InputNumber swallows first keystroke when locale uses non-ASCII numerals (e.g. bn)
## Describe the bug

When `InputNumber` is used with a locale that formats digits as non-ASCII characters
(e.g. `bn` / Bengali, which formats `1` as `১`), typing the digit `1` as the very
first character of an empty field produces `onValueChange` with `value: null` instead
of `value: 1`. All subsequent keystrokes work normally. The same bug affects every
digit on first keystroke for any locale whose numeral system does not include ASCII
digits.

## Root cause

`getPrefixExpression()` and `getSuffixExpression()` extract the prefix/suffix by
calling `formatter.format(1)` and splitting on the ASCII string `'1'`:

```js
prefixChar.current = formatter.format(1).split('1')[0];
suffixChar.current = formatter.format(1).split('1').pop();
```

For Bengali locale, `format(1)` returns `'১'` (U+09E7). That string contains no
ASCII `'1'`, so `.split('1')[0]` returns the entire value `'১'` as the prefix.
`parseValue()` then strips this "prefix" from the typed string, leaving an empty
string that evaluates to `null`.

## Affected locales

Any locale where `new Intl.NumberFormat(locale).format(1)` does not contain ASCII
`'1'` — including but not limited to: `bn`, `ar`, `ar-EG`, `ne`, `mr`, `as`, `my`.

## Environment

- PrimeReact version: 10.x (confirmed with current `main`)
- Browsers: all (pure JS logic)

### Reproducer

_No response_

### System Information

```Shell
System:
    OS: Linux 6.8 Ubuntu 24.04.4 LTS 24.04.4 LTS (Noble Numbat)
    CPU: (8) x64 11th Gen Intel(R) Core(TM) i7-1165G7 @ 2.80GHz
    Memory: 3.99 GB / 15.33 GB
    Container: Yes
    Shell: 5.2.21 - /bin/bash
  Binaries:
    Node: 24.12.0 - /usr/local/bin/node
    Yarn: 1.22.22 - /usr/bin/yarn
    npm: 11.12.1 - /usr/local/bin/npm
    pnpm: 10.21.0 - /usr/local/bin/pnpm
  Browsers:
    Chromium: 146.0.7680.164
    Firefox: 148.0.2
    Firefox Developer Edition: 148.0.2
  npmPackages:
    primereact: ^10.9.7 => 10.9.7 
    react: ^18.3.1 => 18.3.1
```

```jsx
<InputNumber locale="bn" allowEmpty onValueChange={(e) => console.log(e.value)} />
```

1. Render the component above.
2. Focus the empty input.
3. Press the key `1`.
4. Observe: `onValueChange` fires with `value: null`.

### Expected behavior

`onValueChange` should fire with `value: 1` on the first keystroke.

8lurry:
Fix: #8521 

### Problem: 
- In locale bn, typing first digit 1 (displayed as ১) can produce event.value = null on first input.

### Root cause:
- Prefix extraction logic relies on splitting formatted 1 by ASCII 1, which fails for locales using non-ASCII digits.

### Repro:
- Render InputNumber with locale bn and allowEmpty true
- Focus empty input
- Type 1
- Observe onValueChange gives null for first keystroke

### Expected:
- First keystroke should produce numeric value 1

### Proposed fix:
- Avoid ASCII-only split logic in prefix detection
- Derive prefix/suffix using formatToParts and the integer/decimal parts, which is locale-safe for non-Latin numerals

### DIFF
--- a/components/lib/inputnumber/InputNumber.js
+++ b/components/lib/inputnumber/InputNumber.js
@@ -134,7 +134,15 @@ export const InputNumber = React.memo(
             } else {
                 const formatter = new Intl.NumberFormat(_locale, { style: props.mode, currency: props.currency, currencyDisplay: props.currencyDisplay });
 
-                prefixChar.current = formatter.format(1).split('1')[0];
+                // Use formatToParts instead of split('1') so that locales with
+                // non-ASCII numerals (e.g. Bengali ১) are handled correctly.
+                const parts = formatter.formatToParts(1);
+                const integerIndex = parts.findIndex((p) => p.type === 'integer');
+
+                prefixChar.current = parts
+                    .slice(0, integerIndex)
+                    .map((p) => p.value)
+                    .join('');
             }
 
             return new RegExp(`${escapeRegExp(prefixChar.current || '')}`, 'g');
@@ -153,7 +161,15 @@ export const InputNumber = React.memo(
                     roundingMode: props.roundingMode
                 });
 
-                suffixChar.current = formatter.format(1).split('1')[1];
+                // Use formatToParts instead of split('1') so that locales with
+                // non-ASCII numerals (e.g. Bengali ১) are handled correctly.
+                const parts = formatter.formatToParts(1);
+                const integerIndex = parts.findIndex((p) => p.type === 'integer');
+
+                suffixChar.current = parts
+                    .slice(integerIndex + 1)
+                    .map((p) => p.value)
+                    .join('');
             }
 
             return new RegExp(`${escapeRegExp(suffixChar.current || '')}`, 'g');


8lurry:
### CI failure: Security Check (Step 8)

The **NodeJS CI** run is failing in **Security Check** (Step 8):  
https://github.com/primefaces/primereact/actions/runs/23857355195/job/69566392264#step:8:86

The log shows an `npm audit` finding:

- `postcss < 8.4.31` (moderate) via `node_modules/next`
- Audit output says the only full fix is `npm audit fix --force`, which would install **`next@16.2.2`** (breaking change)
- The job then errors on `npm audit fix` and exits with code **1**

This seems like a repo/CI-policy issue rather than something a PR author can safely resolve, since the suggested remediation is a **Next major upgrade**.

**Suggestion (maintainer decision):**
- Avoid running `npm audit fix` in CI (non-deterministic + can require breaking upgrades). Prefer a read-only audit such as:
  - `npm audit --omit=dev --audit-level=critical` (or whatever threshold you want CI to enforce)
- If the goal is to fully remediate, plan an explicit dependency upgrade path (e.g., upgrading Next to a version that pulls a non-vulnerable PostCSS).

Can someone confirm whether CI is intended to run `npm audit fix` here (vs `npm audit` only), and what severity threshold should fail the build?

